PR #6, opened 2022-02-23
merge Module Linking and Interface Types proposals

lukewagner:
This PR merges together and simplifies the Module Linking and Interface Types proposals into one explainer and binary format doc that can be read top-to-bottom without needing to mentally glue together two repos.  The original reason for having two repos is that Module Linking was a normal Core WebAssembly proposal while Interface Types was its own thing.  However, after the [layering and scoping](https://docs.google.com/presentation/d/1PSC3Q5oFsJEaYyV5lNJvVgh-SNxhySWUqZ6puyojMi8) decision, these were to become two proposals making up a single new layered spec in a single MVP release.  After working on this for a while, I found that, similar to our previous experience with reference-types+bulk-memory, this split was causing more harm than good, so this PR merges the two and also simultaneously moves active development over to this `component-model` repo.  In particular, with this merge, it's natural to punt this intermediate "adapter module" concept out of the MVP which is a *significant* simplification (see also `FutureFeatures.md` in this PR).

For now, I'd like to temporarily halt work in the `module-linking` and `interface-types` repos, but in the future they may come back to life:
* If there is renewed interest in adding the declarative style of module linking to Core WebAssembly, as was [originally proposed](https://github.com/WebAssembly/module-linking/blob/63cd6c0e3ac5c0cdb798a985790f51ccdd77af00/proposals/module-linking/Explainer.md), we could easily revert the `module-linking` repo to original Core proposal and move forward with that.
* Post-component-model-MVP, `interface-types` would be the natural repo to (continue) discussing custom adapter functions.

This PR also fills in a bunch of details for how interface types and canonical adapters fit into components which has been pending since the original layering decision.  Unfortunately, the canonical ABI itself is still left as a "TODO" in the `CanonicalABI.md` in this PR; in parallel with getting this PR reviewed, I'll work on updating and transforming Alex's PR in [interface-types/#140](https://github.com/WebAssembly/interface-types/pull/140).

cc @alexcrichton @peterhuene @fitzgen @sunfishcode @rossberg @Pyrolistical @guybedford


design/mvp/Explainer.md
```diff
@@ -0,0 +1,915 @@
...
+necessary to create before being able to *run* the code. Instance definitions
+create module or component instances by selecting a module/component and
+supplying a set of named *arguments* which satisfy all the named *imports* of
+the selected module/component:
+```
+instance     ::= (instance <id>? <instanceexpr>)
+instanceexpr ::= (instantiate (module <moduleidx>) (import <name> <modulearg>)*)
+               | (instantiate (component <componentidx>) (import <name> <componentarg>)*)
+               | <export>*
+               | <core:export>+
```
sunfishcode, line 115: How should a parser disambiguate between `<export>*` and `<core:export>+` here in the case of input like `(export "foo" (func $2))`?

design/mvp/Explainer.md
```diff
@@ -0,0 +1,915 @@
...
+passed as a `componentarg` when instantiating a component, not just instances.
+Component instantiation will be revisited below after introducing the
+prerequisite type and import definitions.
+
+Lastly, the `(instance <export>*)` and `(instance <core:export>+)`
+expressions allow component and module instances to be created by directly
+tupling together preceding definitions, without the need to `instantiate`
+anything. To disambiguate the empty case, we observe that there is never
+a need to import an empty module instance and thus `(instance)` is an empty
+*component* instance. The "inline" forms of these expressions in `modulearg`
```
sunfishcode, line 160: Assuming weak symbols will be added, there could be meaningful situation where an empty core-module instance makes sense: A module API could consist entirely of weak functions, and in some contexts, one may one to create a module with all functions absent.

design/mvp/Explainer.md
```diff
@@ -0,0 +1,915 @@
...
+Components, as we'll see below, have single-level imports, i.e., each import
+has only a single `name`, and thus every different kind of definition can be
+passed as a `componentarg` when instantiating a component, not just instances.
+Component instantiation will be revisited below after introducing the
+prerequisite type and import definitions.
+
+Lastly, the `(instance <export>*)` and `(instance <core:export>+)`
+expressions allow component and module instances to be created by directly
+tupling together preceding definitions, without the need to `instantiate`
+anything. To disambiguate the empty case, we observe that there is never
```
sunfishcode, line 158: Saying "without the need to `instantiate` anything is a little confusing; as I understand it, the tupling form is effectively sugar for creating a module/component which just imports and re-exports things, and then does an `instantiate` on that.

design/mvp/Explainer.md
sunfishcode: It's possible the answer here will be that weak imports are only expected to support components. However even if so, this still feels like it's taking on the risk of assuming that no future feature will ever make empty modules make sense, just to make the s-expression grammar slightly prettier.

design/mvp/Binary.md
```diff
@@ -0,0 +1,252 @@
...
+                    | 0x70 case*:vec(<case>)                    => (variant case*)
+                    | 0x6f t:<intertypeuse>                     => (list t)
+                    | 0x6e t*:vec(<intertypeuse>)               => (tuple t*)
+                    | 0x6d n*:vec(<name>)                       => (flags n*)
+                    | 0x6c n*:vec(<name>)                       => (enum n*)
+                    | 0x6b t*:vec(<intertypeuse>)               => (union t*)
+                    | 0x6a t:<intertypeuse>                     => (optional t)
+                    | 0x69 t:<intertypeuse> u:<intertypeuse>    => (expected t u)
+field             ::= n:<name> t:<intertypeuse>                 => (field n t)
+case              ::= n:<name> t:<type> 0x0                     => (case n t)
```
peterhuene, line 163:
```suggestion
case              ::= n:<name> t:<intertypeuse> 0x0                     => (case n t)
```

design/mvp/Binary.md
```diff
@@ -0,0 +1,252 @@
...
+                    | n:<name> t:<type> 0x1 i:<u32>             => (case n t (defaults-to case-label[i]))
```
peterhuene, line 164:
```suggestion
                    | n:<name> t:<intertypeuse> 0x1 i:<u32>             => (case n t (defaults-to case-label[i]))
```

design/mvp/Explainer.md
```diff
@@ -0,0 +1,915 @@
...
+moduletype-def    ::= <core:deftype>
+                    | <core:import>
+                    | (export <name> <core:importdesc>)
+core:deftype      ::= <core:functype>
+                    | ... Post-MVP additions
+componenttype     ::= (component (componenttype-def)*)
+componenttype-def ::= <import>
+                    | <instancetype-def>
+import            ::= (import <name> <deftype>)
+instancetype      ::= (instance (instancetype-def)*)
```
sunfishcode, line 317: Here and in the other deftypes, some of the examples in the PR have an identifier, like `$logging` in this from below:
```
  (import "wasi:logging" (instance $logging
    (export "log" (func (param string)))
  ))
```

design/mvp/Explainer.md
lukewagner: Excellent question; this is (maybe too) subtle.  I think the answer is that the parser doesn't have to disambiguate: since there is only one function index space (populated by functions of both `core:functype` and `functype`), `(func $f)` always consults the one function index space and thus `<export>` intersects `<core:export>` at `(func <funcidx>)` and it is *validation* that, as part of assigning a module-instance or component-instance type, requires all the exports be uniformly of core or component-layer type.

design/mvp/Explainer.md
lukewagner: Yeah, that's a great point.  I was thinking that, in such a hypothetical future, we could add some sort of explicitly-core-instance syntax.  Note that in the binary format, there already is a core-vs-component discriminant in `instanceexpr`, so this is mostly just a text format thing.

design/mvp/Explainer.md
sunfishcode: Our current thinking in the wasm-tools wast parser is that we want to have separate data structures for core modules and components, because they're sufficiently different that it's usually easier that way. However, that would want the parser to be able to determine whether something is a core module or a component. With the above ambiguity, the parser can't tell (we don't do resolution or validation until a separate phase). Of course, It's Just Software (tm) and we can fix this by making the parser produce a proto-AST where an instance can be either a module or a component, dynamically, and then have the validator produce the real AST.

However, the main benefit of s-expressions is that they directly represent the abstract structure.

design/mvp/Explainer.md
lukewagner: You're right that the net effect is achievable via helper-component, but as I've started digging into the semantics, I don't think it will actually be defined by desugaring.  Rather, I've come to understand `instance` types as the types of *instantiation-time* *record values* (which, as values, are immutable and without identity) and `instantiate` as an instruction that consumes and produces these values.  This "instantiation-time record value" concept is subtly different than what I've traditionally meant by ["instance"](https://searchfox.org/mozilla-central/source/js/src/wasm/WasmInstance.h#47) which is this stateful persistent runtime object created by `instantiate`, but I think it's a crucial distinction when you consider the case of merging the exports of several instances into a new instance.  The distinction will also become more concrete as we dig into the consumed-exactly-once validation rules for `value` (and instances containing them...).  Getting back to this comment, though: with the understanding of "instance as a record value", this tupling `(instance <export>*)` form is just a primitive value constructor.

design/mvp/Explainer.md
```diff
@@ -0,0 +1,915 @@
...
+)
+```
+could be successfully instantiated via:
+```js
+WebAssembly.instantiateStreaming(fetch('./a.wasm'), {
+  one: () => (),
+  two: "hi",
+  three: {
+    four: {
+      five: await WebAssembly.instantiateStreaming(fetch('./b.wasm'))
```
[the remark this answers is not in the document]
lukewagner: Oops, yes; thanks!

design/mvp/Explainer.md
```diff
@@ -0,0 +1,915 @@
...
+```wasm
+;; a.wasm
+(component
+  (import "one" (func))
+  (import "two" (value string))
+  (import "three" (instance
+    (export "four" (instance
+      (export "five" (module
+        (import "six" "a" (func))
+        (export "six" "b" (func))
```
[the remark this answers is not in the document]
lukewagner: Yes, good catch!  (Not sure what even happened there...)

design/mvp/Binary.md

design/mvp/Binary.md

design/mvp/Explainer.md
lukewagner: You're right; currently I hand-wave in the [import and export section](https://github.com/WebAssembly/component-model/blob/ast-and-binary/design/mvp/Explainer.md#import-and-export-definitions), saying that, specifically for `deftype`s that appear as the type of an import, you can bind an identifier.  But thinking about this some more, there are other contexts which want an `<id>` too, so probably I should just add an `<id>?` to the type constructors and just say it's not allowed in various contexts.

lukewagner: Done

design/mvp/Explainer.md
```diff
@@ -0,0 +1,938 @@
...
+   implicitly checked at every execution step by component functions. Thus,
+   after a trap, it's no longer possible to observe the internal state of a
+   component instance.
+2. Components prevent unexpected reentrance by setting the "lockdown" state
+   (in the previous bullet) whenever calling out through an import, clearing
+   the lockdown state on return, thereby preventing reentrant export calls in
+   the interim. This establishes a clear contract between separate components
+   that both prevents obscure composition-time bugs and also enables
+   more-efficient non-reentrant runtime glue code (particularly in the middle
+   of the [Canonical ABI](CanonicalABI.md)).
```
badeend, line 677: Should I interpret this as: "overlapped/multi-threaded access to an imported component will trap" ?

design/mvp/Explainer.md
```diff
@@ -0,0 +1,938 @@
...
+encodings, are given. The `latin1+utf16` encoding is [defined](CanonicalABI.md#latin1-utf16)
+in the Canonical ABI explainer. If no string-encoding option is specified, the
+default is `string=utf8`.
+
+The `into` option specifies a target instance which supplies the memory that
+the canonical ABI should operate on as well as functions that the canonical ABI
+can call to allocate, reallocate and free linear memory. Validation requires that
+the given `instanceidx` is a module instance exporting the following fields:
+```
+(export "memory" (memory 1))
```
badeend, line 492: Just to clarify, `memory 1` refers to memory index 1, i.e. the *second* memory and therefore depends on https://github.com/WebAssembly/multi-memory ?

design/mvp/Explainer.md
lukewagner: Yes.  (The longer-term plan for allowing multi-threaded access to component instances would be to follow the existing [core wasm plan](https://dl.acm.org/doi/10.1145/3360559) and define a new `shared` attribute on functions that allowed them to be called simultaneously from multiple threads (enforced by the VM).  Thus, components would have to opt into the more complicated world of multi-threading.)

design/mvp/Explainer.md
lukewagner: Sorry, no, the `1` is the non-optional [`min`](https://webassembly.github.io/spec/core/syntax/types.html#syntax-limits) field of the `memtype`.

design/mvp/Explainer.md
lukewagner: Yeah, fair enough.  What do you think about:
```
instanceexpr ::= ...
               | <export>*
               | core <core:export>*
```
?

design/mvp/Explainer.md
badeend: Fine by me! Might be worthwhile explicitly summarizing that.

```suggestion
2. Components prevent unexpected reentrance by setting the "lockdown" state
   (in the previous bullet) whenever calling out through an import, clearing
   the lockdown state on return, thereby preventing reentrant export calls in
   the interim. This establishes a clear contract between separate components
   that both prevents obscure composition-time bugs and also enables
   more-efficient non-reentrant runtime glue code (particularly in the middle
   of the [Canonical ABI](CanonicalABI.md)). This implies that components by
   default don't allow concurrency and multi-threaded access will trap.
```

lukewagner: Great, thanks!

design/mvp/Binary.md
```diff
@@ -0,0 +1,252 @@
...
+               | 0x00 0x01 c:<componentidx> a*:vec(<componentarg>) => (instantiate (component c) (import a)*)
+               | 0x01 e*:vec(<export>)                             => e*
+               | 0x02 e*:vec(<core:export>)                        => e*
+modulearg    ::= n:<name> 0x02 i:<instanceidx>                     => n (instance i)
+componentarg ::= n:<name> 0x00 m:<moduleidx>                       => n (module m)
+               | n:<name> 0x01 c:<componentidx>                    => n (component c)
+               | n:<name> 0x02 i:<instanceidx>                     => n (instance i)
+               | n:<name> 0x03 f:<funcidx>                         => n (func f)
+               | n:<name> 0x04 v:<valueidx>                        => n (value v)
+               | n:<name> 0x05 t:<typeidx>                         => n (type t) (t must be an <intertype>)
```
peterhuene, line 63: Could we relax the type restriction here to also include `<functype>`?

The reason for this is we need some way to give names to functions signatures for generating an interface definition from a component file (e.g. `foo: function(...)`).

design/mvp/Explainer.md
```diff
@@ -0,0 +1,939 @@
...
+                    | ... Post-MVP additions
+componenttype     ::= (component <id>? <componenttype-def>*)
+componenttype-def ::= <import>
+                    | <instancetype-def>
+import            ::= (import <name> <deftype>)
+instancetype      ::= (instance <id>? <instancetype-def>*)
+instancetype-def  ::= <type>
+                    | <alias>
+                    | (export <name> <deftype>)
+functype          ::= (func <id>? (param <name> <intertype>)* (result <intertype>))
```
badeend, line 322: Looking at the examples elsewhere in this explainer, parameter names seem to be optional:
```suggestion
functype          ::= (func <id>? (param <name>? <intertype>)* (result <intertype>))
```

---

Regarding the result type:
- It is currently required. Is this because `unit` is now a true datatype?
- Core wasm allows multiple result values. Can this be added here too?
- Any reason the result values can't be named either? The WIT format already allows this.

```suggestion
functype          ::= (func <id>? (param <name>? <intertype>)* (result <name>? <intertype>)+)
```

badeend:
This PR introduces some renamings:
- `f32` -> `float32`
- `f64` -> `float64`
- `option` -> `optional`

Were these intentional?

design/mvp/Explainer.md
lukewagner: Ok, I made this change.  Thanks for raising this point.

design/mvp/Explainer.md
lukewagner: Ok, with the explicit `core` discriminant, this should be a non-issue.

design/mvp/Binary.md
lukewagner: Yeah, I suppose we could could allow exporting any `typeexpr` (allowing any `deftype` to be named).  But just to help picture this: how does the function signature get named and used?  In `foo: function(...)`, I just see `foo` naming a `func` export (not `functype`).

peterhuene: One of the things I was hoping for with the component model binary format was that it could be used as a compact way to represent an interface without any actual implementation; effectively a binary encoding of something like an interface defined in a compatible IDL (e.g. `wit`). I'd expect such an "interface-only" encoding to only have a type and export section (and possible custom sections for additional data such as API documentation, etc).

Validating an exported lifted function requires having a core function in the index space, which requires an alias of an instance export, which requires an instance, which requires a core module (with an `unreachable` implementation of said function). While doable, that's a lot to emit to just give a name to a function signature and that is effectively all a function in an IDL really is.

badeend:
> But for optional, I had actually confused myself by working with too many locally-modified drafts into thinking that optional was the original name. But I see now it's option in the current interface-types explainer and in wit-bindgen today, so I'll revert to option. Thanks for double-checking that!

👍 

> The f32/f64 → float32/float64 renaming reflects the subtle semantic change (regarding NaN) proposed in https://github.com/WebAssembly/interface-types/issues/134, so that was quite intentional.

I understand. But TBH, I think that without knowing this background information most people will assume `f64` to be an alias of `float64`. If the distinction between canonicalized and non-canonicalized floats is really that important, maybe that should be reflected in the name, like: `f64-canon` (instead of `float64`)

design/mvp/Explainer.md
badeend: I suppose the optional names also affect the subtyping rules, which are currently based on parameter name. These can be changed to use parameter position only. Making the entire argument list essentially a tuple type.

design/mvp/Subtyping.md
```diff
@@ -0,0 +1,24 @@
...
+| ------------------------- | --------- |
+| `unit`                    | every interface type is a subtype of `unit` |
+| `bool`                    | |
+| `s8`, `s16`, `s32`, `s64`, `u8`, `u16`, `u32`, `u64` | lossless coercions are allowed |
+| `float32`, `float64`      | `float32 <: float64` |
+| `char`                    | |
+| `record`                  | fields can be reordered; covariant field payload subtyping; superfluous fields can be ignored in the subtype; `option` fields can be ignored in the supertype |
+| `variant`                 | cases can be reordered; contravariant case payload subtyping; superfluous cases can be ignored in the supertype; `defaults-to` cases can be ignored in the subtype |
+| `list`                    | covariant element subtyping |
+| `tuple`                   | `(tuple T ...) <: T` |
```
badeend, line 17: This subtyping rule is ambiguous when the tuple has multiple fields of the same type.
```rust
let result: i32 = (67, 89);
result == ???
```

Similar story for `union` where you can't know which case to construct.
```rust
enum MyUnion { // (union i32 i32)
  _0(i32),
  _1(i32),
}
let result: MyUnion = 56;
result == ???
```

Also, besides the special subtyping rules of `tuple`, `union`, `option` & `expected`, I assume they also inherit all regular rules from `record` or `variant` (whichever they're a specialized form of)

design/mvp/Binary.md
lukewagner: Ah, cool idea, that makes sense.  Ok, I'll remove the "must be `<intertype>`" restriction in the binary format (which is nice just from a regularity perspective).

design/mvp/Explainer.md
lukewagner: Yep!  That's actually the intention of the current subtyping sketch wording (so that languages without named parameters can use functions with only positional arguments).

lukewagner:
> If the distinction between canonicalized and non-canonicalized floats is really that important, maybe that should be reflected in the name

Since almost all folks don't need or want to know about the super-corner-case of non-canonical nan payloads and since the semantics of `float64` matches the usual assumption that there is a single NaN value and since the intended audience of interface types is folks who (eventually, once the toolchain stabilizes) shouldn't need to know any low-level details about core wasm and `f64` (in the same way that you don't need to know x64 to use Go running on x64), I think this is a subtly that is best left out of the name.

design/mvp/Subtyping.md
badeend: Ah right, missed the disclaimer at the top. :bowtie: No further questions.

lukewagner: :+1:

design/mvp/Subtyping.md
```diff
@@ -0,0 +1,24 @@
...
+
+| Type                      | Subtyping |
+| ------------------------- | --------- |
+| `unit`                    | every interface type is a subtype of `unit` |
+| `bool`                    | |
+| `s8`, `s16`, `s32`, `s64`, `u8`, `u16`, `u32`, `u64` | lossless coercions are allowed |
+| `float32`, `float64`      | `float32 <: float64` |
+| `char`                    | |
+| `record`                  | fields can be reordered; covariant field payload subtyping; superfluous fields can be ignored in the subtype; `option` fields can be ignored in the supertype |
+| `variant`                 | cases can be reordered; contravariant case payload subtyping; superfluous cases can be ignored in the supertype; `defaults-to` cases can be ignored in the subtype |
```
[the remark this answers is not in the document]
lukewagner: Oops, right, thanks!

design/mvp/Subtyping.md
```diff
@@ -0,0 +1,24 @@
...
+| `list`                    | covariant element subtyping |
+| `tuple`                   | `(tuple T ...) <: T` |
+| `option`                  | `T <: (option T)` |
+| `expected`                | `T <: (expected T _)` |
+| `union`                   | `T <: (union ... T ...)` |
+| `func`                    | parameter names must match in order; covariant parameter subtyping; superfluous parameters can be ignored in the subtype; `option` parameters can be ignored in the supertype; contravariant result subtyping |
```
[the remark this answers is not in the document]
lukewagner: Oops, right again.

design/mvp/Explainer.md
```diff
@@ -0,0 +1,937 @@
...
+On another technical note: the optional `id` in all the `deftype` type
+constructors (e.g., `(module <id>? ...)`) is only allowed to be present in the
+context of `import` since this is the only context in which binding an
+identifier makes sense.
+
+Starting with interface types, the set of values allowed for the *fundamental*
+interface types is given by the following table:
+| Type                      | Values |
+| ------------------------- | ------ |
+| `unit`                    | just one [uninteresting value] |
```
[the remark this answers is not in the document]
lukewagner: Hah, I wondered about this too.  My worry is that this will end up introducing more implementation work and special-casing than the value it provides (symmetry and describing non-return), so I'm inclined to wait until someone specifically asks for it with a concrete benefit in mind.

design/mvp/Explainer.md
```diff
@@ -0,0 +1,937 @@
...
+
+The sets of values allowed for the remaining *specialized* interface types are
+defined by the following mapping:
+```
+                            string ↦ (list char)
+              (tuple <intertype>*) ↦ (record ("𝒊" <intertype>)*) for 𝒊=0,1,...
+                   (flags <name>*) ↦ (record (field <name> bool)*)
+                    (enum <name>*) ↦ (variant (case <name> unit)*)
+              (option <intertype>) ↦ (variant (case "none") (case "some" <intertype>))
+              (union <intertype>*) ↦ (variant (case "𝒊" <intertype>)*) for 𝒊=0,1,...
```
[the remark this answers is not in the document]
lukewagner: Yeah, I suppose there is some divergence here, but I can't think of a good alternative either.  It's nice that it lines up with Web IDL (and thus the JS API) and that misunderstandings (trying to add non-disjoint cases) will be a relatively early error.

design/mvp/CanonicalABI.md
```diff
@@ -0,0 +1,3 @@
+# Canonical ABI (sketch)
+
+TODO: import and update [interface-types/#132](https://github.com/WebAssembly/interface-types/pull/132)
```
Liamolucko, line 3: This should probably link to the reopened version of the PR.
```suggestion
TODO: import and update [interface-types/#140](https://github.com/WebAssembly/interface-types/pull/140)
```

lukewagner: Oops, right, thanks!

design/mvp/Explainer.md
```diff
@@ -0,0 +1,937 @@
...
+                    | ... Post-MVP additions
+componenttype     ::= (component <id>? <componenttype-def>*)
+componenttype-def ::= <import>
+                    | <instancetype-def>
+import            ::= (import <name> <deftype>)
+instancetype      ::= (instance <id>? <instancetype-def>*)
+instancetype-def  ::= <type>
+                    | <alias>
+                    | (export <name> <deftype>)
+functype          ::= (func <id>? (param <name>? <intertype>)* (result <intertype>))
```
sunfishcode, line 320: In some of the examples, the `(result <intertype>)` is missing, such as in `(export "log" (func (param string)))`. Should the `result` part be optional, or should the examples add a `result`?

lukewagner: Ah, good catch!  In the abstract type, there is always a single result type, which can be `unit`.  Surface text formats like wit would map `-> []` to `-> unit`, so I suppose the text format could do likewise as syntactic sugar (so a missing `result` was the same as `(result unit)`).

lukewagner: I kindof like being able to leave off `result` and the overhead as a pure text format sugar seems low, so I tentatively did that.  To avoid confusing things too much (given that the explainer merges the text format with AST), I left `(result <intertype>)` non-optional in the AST grammar and just added a note regarding the optionality sugar.

lukewagner:
It looks like new comments are winding down, so if it stays that way for a week, I think I'll merge this PR, with the usual understanding that this isn't final and we can iterate in future issues/PRs.  (In the meantime, I'm working on importing and updating the Canonical ABI as the next PR.)

PR #7, opened 2022-03-08
Change "import" to "with" in the instantiate argument syntax.

### LINKED ISSUE #2: Rename `import` in thee case where it isn't an import
In [the explainer](https://github.com/WebAssembly/component-model/blob/ast-and-binary/design/mvp/Explainer.md):

> instanceexpr ::= (instantiate (module <moduleidx>) (import <name> <modulearg>)*)
>                | (instantiate (component <componentidx>) (import <name> <componentarg>)*)
>                | (instance <export>*)

Could we rename the `import` nodes to `with`? Or `modulearg`/`componentarg`, or so? `(import ...)` has a different meaning in a different context. Parsers can disambiguate it, but it's confusing for humans that don't already know their way around.

For example:
```
instanceexpr ::= (instantiate (module <moduleidx>) (with <name> <modulearg>)*)
               | (instantiate (component <componentidx>) (with <name> <componentarg>)*)
               | (instance <export>*)
```

sunfishcode:
This changes `import` to `with`, within an `(instantiate ...)`.

This makes it easier for a human looking at code without full context to figure out what they're looking at.

Fixes #2.


lukewagner:
Thanks!

PR #18, opened 2022-04-05
Allow references to type variables in interface types.

badeend:
The explainer contains the example below. However, the current grammar doesn't allow identifiers in places where interface types are expected.

```wasm
(component $C
  (type $T (list (tuple string bool)))
  (type $U (option $T))
  (type $G (func (param (list $T)) (result $U)))
  (type $D (component
    (alias outer $C $T (type $C_T))
    (type $L (list $C_T))
    (import "f" (func (param $L) (result (list u8))))
    (import "g" $G)
    (export "g" $G)
    (export "h" (func (result $U)))
  ))
)
```

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -327,7 +327,8 @@ instancetype-def  ::= <type>
                     | (export <name> <deftype>)
 functype          ::= (func <id>? (param <name>? <intertype>)* (result <intertype>))
 valuetype         ::= (value <id>? <intertype>)
-intertype         ::= unit | bool
+intertype         ::= <id>
+                    | unit | bool
                     | s8 | u8 | s16 | u16 | s32 | u32 | s64 | u64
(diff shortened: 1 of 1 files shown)


lukewagner:
You're right this is missing but this omission is currently intentional and part of the caveat given in the "On a technical note: ..." paragraph right after the type grammar (and part of the overall "hand-waving" mentioned in the [opening paragraph](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#grammar)).  The idea is to avoid cluttering the explainer's grammar with the full details of name resolution (as you see in, e.g., the [formal core text format](https://webassembly.github.io/spec/core/text/modules.html#type-uses)).

Also, technically, including `<id>` as an option in `<intertype>` would imply that you could write `(type $T $U)` (we'd need to do a bit more grammar refactoring to extract definitions from uses).

PR #19, opened 2022-04-05
start -> invoke

badeend:
Under the [Start Definitions](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#start-definitions) section, the explainer starts off by hinting that module start functions and component start functions are each others equivalent only at a different level:

> Like modules, components can have start functions that are called during instantiation

However, if I understand the rest of the section correctly, they are more like opposites:
- Core Wasm's `(start ...)` construct _declares_ a function to be run when instantiated, while
- ComponentModels's `(start ...)` construct _invokes_ a function during instantiation.

Additionally, ComponentModels's `(start ...)` doesn't even invoke Core Wasm's `(start ...)`, but rather arbitrary exports. Which in turn may be exported with the name "start", but that's only incidental.

If this is indeed correct, I suggest renaming ComponentModels's `(start ...)` to something else like `(invoke ...)` to avoid confusion.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -207,11 +207,11 @@ Notes:
   in `i`'s `instancetype`.
 
 
-## Start Definitions
+## Invoke Definitions
 
-(See [Start Definitions](Explainer.md#start-definitions) in the explainer.)
+(See [Invoke Definitions](Explainer.md#invoke-definitions) in the explainer.)
 ```
-start ::= f:<funcidx> arg*:vec(<valueidx>) => (start f (value arg)*)
+invoke ::= f:<funcidx> arg*:vec(<valueidx>) => (invoke f (value arg)*)
 ```
 Notes:
 * Validation requires `f` have `functype` with `param` arity and types matching `arg*`.
@@ -222,8 +222,8 @@ In addition to the type-compatibility checks mentioned above, the validation
 rules for value definitions additionally require that each value is consumed
 exactly once. Thus, during validation, each value has an associated "consumed"
 boolean flag. When a value is first added to the value index space (via
-`import`, `instance`, `alias` or `start`), the flag is clear. When a value is
-used (via `export`, `instantiate` or `start`), the flag is set. After
+`import`, `instance`, `alias` or `invoke`), the flag is clear. When a value is
+used (via `export`, `instantiate` or `invoke`), the flag is set. After
 validating the last definition of a component, validation requires all values'
 flags are set.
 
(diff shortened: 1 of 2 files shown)


badeend:
It just occurred to me that Core's `start` doesn't take any arguments and doesn't return anything, so it _too_ can be thought of as an invocation construct. :thinking:

I don't think this affects the suggested changes, though.

lukewagner:
Hmm, I don't quite follow the asymmetry you're pointing out.  The `start` sections at *both* levels invoke the referenced (pre-existing) function during instantiation.  It is true that the component-level `start` doesn't directly invoke a core-level `start`, but I think this is just a natural consequence of composition/nesting.  E.g., consider a single outer component that creates *two* nested module instances: you'd expect this to first run the first nested module `start`, then the second nested module `start`, then lastly the outer component's `start`.  Thus a component `start` isn't meant to be a wrapper around a single module `start` since there are 0..N nested modules (and components).

badeend:
> The start sections at both levels invoke the referenced (pre-existing) function during instantiation.

Yes, I came to the same realization in my follow-up comment. You can ignore the asymmetry remarks. The only thing that still holds up about my initial comment is the rename suggestion.

> E.g., consider a single outer component that creates two nested module instances: you'd expect this to first run the first nested module start, then the second nested module start, then lastly the outer component's start. Thus a component start isn't meant to be a wrapper around a single module start since there are 0..N nested modules (and components).

I fully understand. I might have jumped over a few steps, so let me back up a bit and explain my train of thought:
- Modules may declare a single function to be its `start` function.
- That function is invoked during module instantiation as part of the `(instantiate (module ...))` statement in the component.
- During instantiation, components have access to module exports.
- During instantiation, components can lift & invoke any exported function from the module (or sub components for that matter).
- During instantiation, components can even invoke *multiple* exports from the same module.


In those last two points lies my issue. The name `start` makes sense for Core wasm, because there is always 0 or 1 start function.
In the component-model, Components can invoke 0 to many functions during initialization, even from the same module instance. 

Grammatically, (as in English grammar, not the AST grammar :) ) it doesn't make sense to "start" the something multiple times, because that would imply it was somehow "stopped" in the between times? Or: using the "start" construct on the same module instance multiple times makes it super-duper-started? :upside_down_face:

I find myself having trouble to fully express myself in a non-native language about grammar of said foreign language, so apologies if my argument doesn't make sense. :) 

To reiterate, my only remaining issue is renaming `(start ...)` to something else like `(invoke ...)`.

lukewagner:
Ah, thanks for explaining your thinking in more detail and dealing with the vagaries of language.  So, regarding the "0 or 1" vs. "0 to N" difference between components and core modules: one generalization that has been discussed a couple times (and I expect will eventually happen) in core wasm is loosening the section ordering and multiplicity of core module sections so that sections (including `start`) could be repeated, as they are in components.  In that future, then component and module `start` sections would be relatively symmetric.  You may be right that repeated "start" is a bit of a weird name, but I guess that ship has sailed in core wasm :)

badeend:
Alright. Bummer, but understandable.

Issue #15, opened 2022-04-02
Default values


badeend:
TLDR, these proposed changes:
- Allow function parameters, record fields and variant case payload to have default values, as long as they're constant.
- When subtyping; allow variant cases to fall back on cases which have non-unit payload.
- Strip the `option` type from its special status in various places.

### Definition

```
literal-bool ::= 'false' | 'true'
literal-char ::= '"' <core:stringelem> '"'
constant ::= unit
	| (bool <literal-bool>)
	| (s8 <core:s8>) | (u8 <core:u8>)
	| (s16 <core:s16>) | (u16 <core:u16>)
	| (s32 <core:s32>) | (u32 <core:u32>)
	| (s64 <core:s64>) | (u64 <core:u64>)
	| (float32 <core:f32>) | (float64 <core:f64>)
	| (char <literal-char>) | (string <core:string>)
	| (record (field <name> <constant>)*)
	| (variant (case <name> <constant>))
	| (list <intertype> <constant>*)
```

Regarding the `list` constant: every provided `<constant>` must a subtype of the list's `<intertype>`.

### Aliases

```
                 false ↦ (bool false)
                  true ↦ (bool true)
   (tuple <constant>*) ↦ (record (field "𝒊" <constant>)*) for 𝒊=0,1,...
       (flags <name>*) ↦ (record (field <name> (bool true))*)
                  none ↦ (variant (case "none" unit))
     (some <constant>) ↦ (variant (case "some" <constant>))
       (ok <constant>) ↦ (variant (case "ok" <constant>))
    (error <constant>) ↦ (variant (case "error" <constant>))
```

### Types

Every constant expression has a fully deterministic interface type.

- Primitive constants map 1:1 to an interface type.
- List constants declare the element intertype in the constant expression itself.
- Record and Variant constants generate an ad-hoc interface type derived from the constant itself. For example:

```
Constant:
	(record
		(field "a" (s16 42))
		(field "b" (string "Wasm"))
		(field "c" (variant (case "ok" (bool true))))
	)

Generated type:
	(record
		(field "a" s16)
		(field "b" string)
		(field "c" (variant (case "ok" bool)))
	)
```

### Updates to existing definitions

```diff
- (record (field <name> <intertype>)*)
+ (record (field <name> <intertype> (default <constant>)?)*)
	* <constant> must be a subtype of <intertype>


- (variant (case <name> <intertype> (defaults-to <name>)?)*)
[...]

PR #27, opened 2022-04-24
Avoid potential name clash with default values.

badeend:
While typing out [Default values](https://github.com/WebAssembly/component-model/issues/15) a couple of weeks ago, I noticed an unfortunate naming collision between the terms "default" values and the "defaults-to" property of variants.

Regardless of the outcome of that issue, I suggest renaming `defaults-to` to something else. Just as a precautionary measure.


lukewagner:
Eliminating the `-` might be nice; how about `refines`?

badeend:
No strong opinion. I've updated the PR.

lukewagner:
Thanks!  lgtm, I'll wait a day for any other comments then merge.

Issue #11, opened 2022-03-29
Index spaces feel inconsistent about what they might or might not contain


alexcrichton:
Personally I get confused about the various index spaces that are introduced in the component model:

* Type index space - this includes all interface types, component types, module types, etc. This, however, cannot name module instance types.
* Module index space - this includes only defined core wasm modules
* Component index space - this includes only defined components
* Function index space - this includes all of core wasm functions, lifted functions, and lowered functions
* Instance index space - this includes both module instances and component instances

Personally I keep getting lost about which index space includes what. For example just now I wasn't sure if modules and components were in the same index space because their instances are in the same index space.

Would it be possible to separate these index spaces? For example could the function and index spaces be split between module/component types?

fitzgen:
+1 this is something I've repeatedly run up against as well.

It seems like we might as well "use the type system" a little bit to ensure that e.g. a function we are lifting is actually a core function and not a component function.

lukewagner:
I think you're right that the index space division is rather ad hoc; it's attempting to balance a few competing goals and compromising as a result, but maybe it's not the right compromise.

So there's a pretty clear core wasm convention (which, if we're embedding core wasm into components, I think we should maintain) that the `sort` in `(sort <idx>)` or `(sort <dfn>)` selects the index space (where `sort` is one of `type`, `func`, etc).  Thus, if we had, e.g., four different index spaces for {core,component}x{function,instance}, we'd need four distinct `sort`s, e.g.: `(core func <idx>|<dfn>)`/`(core instance <idx>|<dfn>)`/`(func <idx>/<dfn>)`/`(instance <idx>/<dfn>)`.  Originally, this seemed like unnecessary extra verbosity to me (given that we have to validate the result anyway), but I'm interested to discuss this option more if we think this partitioning of index spaces is worth it.

The big irregularity with the current scheme that you mentioned that bugs me too is that components and modules are in different index spaces; if they followed the function/instance convention, they'd share a single index space.  My main hangup here is that if we put both components/modules into the same index space, we'd have to pick a single shared `X` for which we'd write `(X $id)` to refer to both modules and components (and then maybe `(X <def>)` to define either one?!), and I can't think of a good new `X` and reusing `module` or `component` seems like it would create more confusion.

Finally: the type index space doesn't contain module instance types only because there's no need for them at the moment.  If the need arose, there's no reason we couldn't add them (and/or core function types as well).

alexcrichton:
[...]

PR #29, opened 2022-05-02
Split core definitions out into separate index spaces

lukewagner:
This PR addresses #11 as proposed in that issue.  The split ended up touching a bunch of the AST/text/binary and even the basic terminology used so the PR ended up being fairly extensive and requiring a few tries to get something that feels right.  But ultimately, all the changes here should be superficial; the essential concepts should be the same.

As part of this change, a few simplifications and better symmetries (between core and component layers) presented themselves which are included in the PR:
* Instead of enumerating all the different definitions in various places, the PR defines `sort` and `core:sort` which enumerate them in one place so that they can be reused.  A "sort" is what you get when you consider that `(func)` and `(import (func))` are literally different kinds of definitions but both go into the same index space.  So "sort" is 1:1 with index space, and we can say that a definition has a "sort".
* "Interface types" and `intertype` increasingly seem like the wrong names for the type of "first class values used by components" when juxtaposed with `core:valtype`.  Moreover, the "interface" of a component isn't *just* defined by value types, but also the other types too (`func`, `module`, `component`, `instance`).  Thus, this PR replaces `intertype` with `valtype`, drops the word "interface" when referring to values, value types, functions, function types and instead qualifies these words with "core" and "component" when needed.
* `deftype` was changed to align with the GC proposal which had a slightly different meaning and context (previously, "deftype" meant "definition type" (i.e., the type of second-class definitions)), but now it means "defined typed" (i.e., things you can stick in a `(type ...)` definition, which is subtly different).
* Type imports/exports better align with the existing core type imports/exports proposal by wrapping the exact-type bounds in `(eq ...)`.
* The production names ending in `-def` (that appeared inside module, instance and component types) were renamed to end in `decl` ("declarator") because, in my experience, the hyphen is hard to use in communication.  "Declarator" isn't exactly a perfect name, but at least it's a single word that we can give a clear meaning in this spec).
* The `kind` field of the binary format `preamble` was renamed to `layer` because "kind" seems overloaded.

There were also a few bug fixes that I found when rereading everything.  In particular, I see before I was a bit sloppy with abusing the syntactic sugar, so a few examples got an extra `(func ...)` or `(type ...)` wrapped around the identifier.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -84,17 +84,17 @@ export              ::= n:<name> si:<sortidx>                              => (e
 Notes:
 * Reused Core binary rules: [`core:name`]
 * The `core:sort` values are chosen to match the discriminant opcodes of
-  [`core:importdesc`] so that `core:exportdesc` (below) is identical.
+  [`core:importdesc`].
 * `type` is added to `core:sort` in anticipation of the [type-imports] proposal. Until that
   proposal, core modules won't be able to actually import or export types, however, the
   `type` sort is allowed as part of outer aliases (below).
 * `module` and `instance` are added to `core:sort` in anticipation of the [module-linking]
-  proposal, which would add these types to Core WebAssembly. Again, core modules won't be
-  able to actually import or export modules/instances, but they are used for aliases.
+  proposal, which would add these types to Core WebAssembly. Until then, they are useful
+  for aliases (below).
+* Validation of `core:instantiatearg` would initially only allow the `instance`
+  sort, but would be extended to accept other sorts as core wasm is extended.
 * The indices in `sortidx` are validated according to their `sort`'s index
   spaces, which are built incrementally as each definition is validated.
-* The types of arguments supplied by `instantiate` are validated against the
-  types of the matching import according to the [subtyping](Subtyping.md) rules.
 
 ## Alias Definitions
 
@@ -269,12 +269,13 @@ flags are set.
 (See [Import and Export Definitions](Explainer.md#import-and-export-definitions)
 in the explainer.)
 ```
-import ::= n:<name> et:<externtype> => (import n et)
+import ::= n:<name> ed:<externdesc> => (import n ed)
 export ::= n:<name> si:<sortidx>    => (export n si)
 ```
 Notes:
 * Validation requires all import and export `name`s are unique.
-
+* Validation requires any exported `sortidx` to have a valid `externdesc`
+  (which disallows core sorts other than `core module`).
 
 
 [`core:section`]: https://webassembly.github.io/spec/core/binary/modules.html#binary-section
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -216,7 +216,7 @@ Notes:
 
 (See [Canonical Definitions](Explainer.md#canonical-definitions) in the explainer.)
 ```
-canon    ::= 0x00 0x00 f:<core:funcidx> ft:<typeidx> opts:<opts> => (canon lift f type-index-space[ft] opts (func))
+canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift f opts type-index-space[ft])
            | 0x01 0x00 f:<funcidx> opts:<opts>                   => (canon lower f opts (core func))
 opts     ::= opt*:vec(<canonopt>)                                => opt*
 canonopt ::= 0x00                                                => string-encoding=utf8
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -57,7 +57,7 @@ Notes:
 core:instance       ::= ie:<instance-expr>                                 => (instance ie)
 core:instanceexpr   ::= 0x00 m:<moduleidx> arg*:vec(<core:instantiatearg>) => (instantiate m arg*)
                       | 0x01 e*:vec(<core:export>)                         => e*
-core:instantiatearg ::= n:<name> 0x11 i:<instanceidx>                      => (with n (instance i))
+core:instantiatearg ::= n:<name> 0x12 i:<instanceidx>                      => (with n (instance i))
 core:sortidx        ::= sort:<core:sort> idx:<u32>                         => (sort idx)
 core:sort           ::= 0x00                                               => func
                       | 0x01                                               => table
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -203,7 +203,7 @@ Notes:
   with type opcodes starting at SLEB128(-1) (`0x7f`) and going down,
   reserving the nonnegative SLEB128s for type indices.
 * Validation of `valtype` requires the `typeidx` to refer to a `defvaltype`.
-* Validation of `moduledecl` (currently) only allows `outer` `type` `alias`
+* Validation of `instancedecl` (currently) only allows `outer` `type` `alias`
   declarators.
 * As described in the explainer, each component and instance type is validated
   with an initially-empty type index space. Outer aliases can be used to pull
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -102,7 +102,6 @@ Notes:
 ```
 core:alias       ::= sort:<core:sort> target:<core:aliastarget> => (core alias target (sort))
 core:aliastarget ::= 0x00 i:<core:instanceidx> n:<name>         => export i n
-                   | 0x01 ct:<varu32> idx:<varu32>              => outer ct idx
 
 alias            ::= sort:<sort> target:<aliastarget>           => (alias target (sort))
 aliastarget      ::= 0x00 i:<instanceidx> n:<name>              => export i n
@@ -131,7 +130,6 @@ core:deftype    ::= ft:<core:functype>                     => ft               (
 core:moduletype ::= 0x50 md*:vec(<core:moduledecl>)        => (module md*)
 core:moduledecl ::= 0x00 i:<core:import>                   => i
                   | 0x01 t:<core:type>                     => t
-                  | 0x02 a:<core:alias>                    => a
                   | 0x03 e:<core:exportdecl>               => e
 core:import     ::= m:<name> f:<name> ed:<core:externdesc> => (import m f ed)  (WebAssembly 1.0)
 core:externdesc ::= id:<core:importdesc>                   => id               (WebAssembly 1.0)
@@ -142,8 +140,6 @@ Notes:
 * `core:import` as written above is binary-compatible with [`core:import`].
 * Validation of `core:moduledecl` (currently) rejects `core:moduletype` definitions
   inside `type` declarators (i.e., nested core module types).
-* Validation of `core:moduledecl` (currently) only allows `outer` `type`
-  `alias` declarators.
 * As described in the explainer, each module type is validated with an
   initially-empty type index space. Outer aliases can be used to pull
   in type definitions from containing components.
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -73,7 +73,7 @@ instanceexpr        ::= 0x00 c:<componentidx> arg*:vec(<instantiatearg>)   => (i
                       | 0x01 e*:vec(<export>)                              => e*
 instantiatearg      ::= n:<name> si:<sortidx>                              => (with n si)
 sortidx             ::= sort:<sort> idx:<varu32>                           => (sort idx)
-sort                ::= 0x00                                               => core module
+sort                ::= 0x00 csi:<core:sortidx>                            => core csi
                       | 0x01                                               => func
                       | 0x02                                               => value
                       | 0x03                                               => type
@@ -194,7 +194,7 @@ instancedecl  ::= 0x01 t:<type>                        => t
                 | 0x03 ed:<exportdecl>                 => ed
 importdecl    ::= n:<name> et:<externtype>             => (import n et)
 exportdecl    ::= n:<name> et:<externtype>             => (export n et)
-externtype    ::= 0x00 i:<core:typeidx>                => (core module core-type-index-space[i])
+externtype    ::= 0x00 0x10 i:<core:typeidx>           => (core module core-type-index-space[i])  (must be moduletype)
                 | sort:<sort> i:<typeidx>              => (sort type-index-space[i])  (sort must match type)
 ```
 Notes:
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -73,7 +73,7 @@ instanceexpr        ::= 0x00 c:<componentidx> arg*:vec(<instantiatearg>)   => (i
                       | 0x01 e*:vec(<export>)                              => e*
 instantiatearg      ::= n:<name> si:<sortidx>                              => (with n si)
 sortidx             ::= sort:<sort> idx:<varu32>                           => (sort idx)
-sort                ::= 0x00 si:<core:sortidx>                             => si
+sort                ::= 0x00                                               => core module
                       | 0x01                                               => func
                       | 0x02                                               => value
                       | 0x03                                               => type
@@ -150,30 +150,12 @@ Notes:
 
 ```
 type          ::= dt:<deftype>                         => (type dt)
-deftype       ::= vt:<valtype>                         => vt
+deftype       ::= dvt:<defvaltype>                     => dvt
                 | ft:<functype>                        => ft
+                | tt:<typetype>                        => tt
                 | ct:<componenttype>                   => ct
                 | it:<instancetype>                    => it
-functype      ::= 0x40 param*:vec(<param>) t:<valtype> => (func param* (result t))
-param         ::= 0x00 t:<valtype>                     => (param t)
-                | 0x01 n:<name> t:<valtype>            => (param n t)
-componenttype ::= 0x41 cd*:vec(<componentdecl>)        => (component cd*)
-instancetype  ::= 0x42 id*:vec(<instancedecl>)         => (instance id*)
-componentdecl ::= 0x00 id:<importdecl>                 => id
-                | id:<instancedecl>                    => id
-instancedecl  ::= 0x01 t:<type>                        => t
-                | 0x02 a:<alias>                       => a
-                | 0x03 ed:<exportdecl>                 => ed
-importdecl    ::= n:<name> ed:<externdesc>             => (import n ed)
-exportdecl    ::= n:<name> ed:<externdesc>             => (export n ed)
-externdesc    ::= 0x00 i:<core:typeidx>                => core-type-index-space[i]  (must be moduletype)
-                | 0x01 i:<typeidx>                     => type-index-space[i]  (must be func|instance|componenttype)
-                | 0x02 t:<valtype>                     => (value t)
-                | 0x03 tb:<typebound>                  => (type tb)
-typebound     ::= 0x00 i:<typeidx>                     => (eq type-index-space[i])  (any deftype)
-                | 0x00 t:<valtype>                     => (eq t)
-valtype       ::= i:<typeidx>                          => type-index-space[i]  (must be valtype)
-                | 0x7f                                 => unit
+primvaltype   ::= 0x7f                                 => unit
                 | 0x7e                                 => bool
                 | 0x7d                                 => s8
                 | 0x7c                                 => u8
@@ -187,6 +169,7 @@ valtype       ::= i:<typeidx>                          => type-index-space[i]  (
                 | 0x74                                 => float64
                 | 0x73                                 => char
                 | 0x72                                 => string
+defvaltype    ::= pvt:<primvaltype>                    => pvt
                 | 0x71 field*:vec(<field>)             => (record field*)
                 | 0x70 case*:vec(<case>)               => (variant case*)
                 | 0x6f t:<valtype>                     => (list t)
@@ -196,9 +179,27 @@ valtype       ::= i:<typeidx>                          => type-index-space[i]  (
                 | 0x6b t*:vec(<valtype>)               => (union t*)
                 | 0x6a t:<valtype>                     => (option t)
                 | 0x69 t:<valtype> u:<valtype>         => (expected t u)
+valtype       ::= i:<typeidx>                          => type-index-space[i]  (must be defvaltype)
+                | pit:<primvaltype>                    => pit
 field         ::= n:<name> t:<valtype>                 => (field n t)
 case          ::= n:<name> t:<valtype> 0x0             => (case n t)
                 | n:<name> t:<valtype> 0x1 i:<varu32>  => (case n t (refines case-label[i]))
+typetype      ::= tb:<typebound                        => (type tb)
+typebound     ::= 0x00 i:<typeidx>                     => (eq type-index-space[i])
+functype      ::= 0x40 param*:vec(<param>) t:<valtype> => (func param* (result t))
+param         ::= 0x00 t:<valtype>                     => (param t)
+                | 0x01 n:<name> t:<valtype>            => (param n t)
+componenttype ::= 0x41 cd*:vec(<componentdecl>)        => (component cd*)
+instancetype  ::= 0x42 id*:vec(<instancedecl>)         => (instance id*)
+componentdecl ::= 0x00 id:<importdecl>                 => id
+                | id:<instancedecl>                    => id
+instancedecl  ::= 0x01 t:<type>                        => t
+                | 0x02 a:<alias>                       => a
+                | 0x03 ed:<exportdecl>                 => ed
+importdecl    ::= n:<name> et:<externtype>             => (import n et)
+exportdecl    ::= n:<name> et:<externtype>             => (export n et)
+externtype    ::= 0x00 i:<core:typeidx>                => (core module core-type-index-space[i])
+                | sort:<sort> i:<typeidx>              => (sort type-index-space[i])  (sort must match type)
 ```
 Notes:
 * The type opcodes follow the same negative-SLEB128 scheme as Core WebAssembly,
@@ -209,10 +210,6 @@ Notes:
 * As described in the explainer, each component and instance type is validated
   with an initially-empty type index space. Outer aliases can be used to pull
   in type definitions from containing components.
-* The rule for `typebound` contains both an unrestricted `<typeidx>` case and,
-  within `valtype`, a `valtype`-restricted `<typeidx>` case. Since the former
-  is a strict generalization of the latter, there is no ambiguity. The net
-  effect is that `eq` accepts all types.
 
 
 ## Canonical Definitions
@@ -273,7 +270,7 @@ flags are set.
 (See [Import and Export Definitions](Explainer.md#import-and-export-definitions)
 in the explainer.)
 ```
-import ::= n:<name> ed:<externdesc> => (import n ed)
+import ::= n:<name> et:<externtype> => (import n et)
 export ::= n:<name> si:<sortidx>    => (export n si)
 ```
 Notes:
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -73,7 +73,7 @@ instanceexpr        ::= 0x00 c:<componentidx> arg*:vec(<instantiatearg>)   => (i
                       | 0x01 e*:vec(<export>)                              => e*
 instantiatearg      ::= n:<name> si:<sortidx>                              => (with n si)
 sortidx             ::= sort:<sort> idx:<varu32>                           => (sort idx)
-sort                ::= 0x00 csi:<core:sortidx>                            => core csi
+sort                ::= 0x00 cs:<core:sort>                                => core cs
                       | 0x01                                               => func
                       | 0x02                                               => value
                       | 0x03                                               => type
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -131,13 +131,11 @@ core:moduletype ::= 0x50 md*:vec(<core:moduledecl>)        => (module md*)
 core:moduledecl ::= 0x00 i:<core:import>                   => i
                   | 0x01 t:<core:type>                     => t
                   | 0x03 e:<core:exportdecl>               => e
-core:import     ::= m:<name> f:<name> ed:<core:externdesc> => (import m f ed)  (WebAssembly 1.0)
-core:externdesc ::= id:<core:importdesc>                   => id               (WebAssembly 1.0)
-core:exportdecl ::= n:<name> ed:<core:externdesc>          => (export n ed)
+core:importdecl ::= i:<core:import>                        => i
+core:exportdecl ::= n:<name> d:<core:importdesc>           => (export n d)
 ```
 Notes:
-* Reused Core binary rules: [`core:importdesc`], [`core:functype`]
-* `core:import` as written above is binary-compatible with [`core:import`].
+* Reused Core binary rules: [`core:import`], [`core:importdesc`], [`core:functype`]
 * Validation of `core:moduledecl` (currently) rejects `core:moduletype` definitions
   inside `type` declarators (i.e., nested core module types).
 * As described in the explainer, each module type is validated with an
@@ -148,7 +146,6 @@ Notes:
 type          ::= dt:<deftype>                         => (type dt)
 deftype       ::= dvt:<defvaltype>                     => dvt
                 | ft:<functype>                        => ft
-                | tt:<typetype>                        => tt
                 | ct:<componenttype>                   => ct
                 | it:<instancetype>                    => it
 primvaltype   ::= 0x7f                                 => unit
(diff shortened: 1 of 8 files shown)


lukewagner:
(For anyone who eagerly looked at the PR yesterday, the second commit fixes the binary encoding of `valtype` to match how it worked with `intertype`, and also refactors the type grammar a bit.)

design/mvp/Explainer.md
```diff
@@ -105,125 +123,153 @@ dead code.
...
+The syntax for defining component instances is symmetric to core module
+instances, but with a distinct component-level definition of `sort`:
+```
+instance       ::= (instance <id>? <instanceexpr>)
+instanceexpr   ::= (instantiate <componentidx> <instantiatearg>*)
+                 | <export>*
+instantiatearg ::= (with <name> <sortidx>)
+                 | (with <name> (instance <export>*))
+sortidx        ::= (<sort> <varu32>)
+sort           ::= core module
```
pl-semiotics, line 193: I wonder if this should be `core <core:sort>` in the grammar, and we can just let validation deal with ruling out non-shared-nothing imports/exports. I think that makes the AST nicer (vs. having to deal with a few special cases of `core:sort` here), and having more things here seems important for the future, since component aliases also use `sort`.

lukewagner: Great question!  I've oscillated on this a few times too.  Having had it the other way for a while, I like capturing this distinction in the grammar since the main point of `sortidx` is to pass values *between* components (hence the restriction to just the component-level things with the special allow list of core things).  So it makes sense that the set of things you pass between components is distinct (not a simple superset) of the set of things you pass between core modules.  I wonder if this suggests renaming `sortidx` to `externidx` or something (since `extern` is now the regular prefix for these kinds of "at the boundaries" productions)?

design/mvp/Explainer.md
```diff
@@ -105,125 +123,150 @@ dead code.
...
-refers to the current component. To maintain the acyclicity of module
+there are two kinds of "targets" for an alias: the `export` of an instance and
+a definition in an index space of an `outer` component (containing the current
+component):
+```
+core:alias       ::= (alias <core:aliastarget> (<core:sort> <id>?))
+core:aliastarget ::= export <core:instanceidx> <name>
+                   | outer <varu32> <varu32>
+
+alias            ::= (alias <aliastarget> (<sort> <id>?))
```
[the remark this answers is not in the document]
lukewagner: Ooh, great catch!  I also felt unsure about the thing you mentioned at the end so hearing you mention it being weird too makes he more sure it's a bad idea.  I think when you put these together, it suggests simply removing the `outer` case from `core:alias` for now.  In a module-linking future, once nested `core:moduletype` are allowed, it could be re-added, but it would be restricted to only traverse through outer `core:moduletype` -- it wouldn't be able to "see" into components.  (Outside of a `core:moduletype` context, you're effectively at core-top-level so the only valid outer-count is `0`.)

Your suggestion of just having a single component-level `alias` and adding `core export` as a case to `aliastarget` isn't bad, but that also requires adding all `core:sort` to `sort` which I'm also not super in favor (as explained in my reply below).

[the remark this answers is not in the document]
lukewagner: (Tentatively) [fixed](https://github.com/WebAssembly/component-model/pull/29/commits/0e704d502ae06eb1d154679f5186b64b29c48a05)

design/mvp/Explainer.md
pl-semiotics: I could perhaps see renaming it to something like `externidx`; I am increasingly feeling that it's confusing to have something named `sort` that includes some-but-not-all of the core sorts---the `sort` grammar I think should include either all or none of the core things, because it doesn't (by name or really general usage, since it shows up in at least one other place) seem like it should be specialized to (im|ex)ports.  Perhaps we could do something like layer on an `externarg ::= sortidx | (core module <varu32>)` (with `sortidx` and `sort` referring only to component sorts)? However, we might then have to do something like add productions to `alias` that use either `<sort>` or `core <sort>`... (if we do this, the "extend component aliastarget and remove core alias" option comes back on the table, though.)

Stepping back a little, I think I'm uncomfortable with making these kinds of specializations on behalf of what can be imported/exported due to the tension between my impression that export and alias have different requirements for what they can/can't reference (they're currently basically the same, modulo possibly core types for convenience in the alias case, but it feels to me like this is by accident, and fragile to both stateful aliases arriving in the future and new kinds of core objects appearing), and my impression that export and alias should should share this `sort` mechanic because they both reach into arbitrary index spaces.

Also, I feel like the specialization makes working with the AST more annoying, because you can't easily reuse things designed to deal with core sorts when working on these component-or-limited-core sorts (since the limited core doesn't use the same AST nodes as the real `core:sort`).  So as I think about/work with the actual AST more, I find myself increasingly on the side of just putting `core:sort` inside `sort` in the AST.  The text and binary formats could do the necessary validation and then use that more uniform AST, perhaps? Although to me it really feels like something that belongs in validation---typechecking a module's (im|ex)ports seems like the right place to put a requirement around what can be (im|ex)ported.

design/mvp/Explainer.md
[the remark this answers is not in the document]
pl-semiotics: I also thought about constraining the core outer alias production to only reference surrounding core module (types), rather than also including components, but I think that in a private-types-in-core feature, it might be important to have a mechanism for the type of an exported module to alias in a private type that's part of the parent component's core type index space?  Although perhaps you could convert anything you wanted to do that to into a type import on the module, and then pass around a component containing both the type and the module until instantiation time, or something.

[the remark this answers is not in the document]
lukewagner: Good point and yep!  With type imports and the forthcoming currying support (a 'bind' operator), I think we'd have something equivalent.

design/mvp/Explainer.md
lukewagner: Yeah, those are really good points; I think you're right.  Thanks for digging into this!  [Fixed](https://github.com/WebAssembly/component-model/pull/29/commits/6d2e44a407928eba7dab1e2325f1d29bb1938b6b)

design/mvp/Explainer.md
```diff
@@ -234,141 +285,174 @@ definition, resolved using normal lexical scoping rules. For example, the
...
 ```
 is desugared into:
 ```wasm
 (component $C
-  (module $M ...)
+  (core module $M ...)
   (component
-    (alias outer $C $M (module $C_M))
-    (instance (instantiate (module $C_M)))
+    (core alias outer $C $M (module $C_M))
```
pl-semiotics, line 299: Shouldn't this be `(alias outer $C $M (core module $C_M))`? Since we got rid of (/restricted to core instances/modules) `core:aliastarget`'s `outer production.

design/mvp/Explainer.md
```diff
@@ -234,141 +285,174 @@ definition, resolved using normal lexical scoping rules. For example, the
...
 
 Lastly, for symmetry with [imports][func-import-abbrev], aliases can be written
-in an inverted form that puts the definition kind first:
+in an inverted form that puts the sort first:
 ```wasm
-(func $f (import "i" "f")) ≡ (import "i" "f" (func $f))  ;; (existing)
-(func $g (alias $i "g1"))  ≡ (alias $i "g1" (func $g))   ;; (new)
+(func $f (import "i" "f"))            ≡ (import "i" "f" (func $f))             (WebAssembly 1.0)
+(func $g (alias export $i "g1"))      ≡ (alias export $i "g1" (func $g))
+(core func $g (alias export $i "g1")) ≡ (core alias export $i "g1" (func $g))
```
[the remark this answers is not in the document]
lukewagner: I think the former is also what you get if you think of `(core ...)` as saying to parse the `...` as future-core syntax and future-core syntax defines this inverted abbreviation for (core) aliases -- we're essentially just inheriting it.

design/mvp/Explainer.md
lukewagner: Good spot!  I think if we really wanted to, we could come up with a way of rationalizing some sort of "first pass" in which the component layer resolved inline-alias-sugar by expansion into aliases before "handing off" to the proper core level... but it seems best to start conservative here and see if we really want this extra complexity.  So I [fixed](https://github.com/WebAssembly/component-model/pull/29/commits/eb68de9bf48f40b617be6a825f2957b80c288f46) the example to alias a component instead of a module.

design/mvp/Explainer.md
```diff
@@ -105,125 +123,158 @@ dead code.
...
+The syntax for defining component instances is symmetric to core module
+instances, but with an expanded component-level definition of `sort`:
+```
+instance       ::= (instance <id>? <instanceexpr>)
+instanceexpr   ::= (instantiate <componentidx> <instantiatearg>*)
+                 | <export>*
+instantiatearg ::= (with <name> <sortidx>)
+                 | (with <name> (instance <export>*))
+sortidx        ::= (<sort> <varu32>)
+sort           ::= core-prefix(<core:sortidx>)
```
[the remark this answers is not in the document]
pl-semiotics: Sorry for the unclarity---I meant, shouldn't there be `core:sort` rather than `core:sortidx` inside the `core-prefix()`?

design/mvp/Explainer.md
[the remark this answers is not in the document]
pl-semiotics: Ah, yes, that makes sense!

design/mvp/Explainer.md
pl-semiotics: That sounds like a good plan, thanks!

design/mvp/Explainer.md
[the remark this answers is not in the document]
lukewagner: Oh haha, sorry, I see it now; my eyes were glazing right over the `idx` in both the code and your comment :P  [fixed](https://github.com/WebAssembly/component-model/pull/29/commits/bb78e7017722459258107b8d63c35987ccf8b80e)

lukewagner:
This most recent update removes the hand-waving in type grammar since @pl-semiotics showed it was ambiguous and simply wrong in a few cases.  The type grammar now uses explicit type indices (calling out when they need `(type <typeidx>)`/`(type $id)` vs when they're just `<typeidx>`/`$id`) and inline type expressions that should more-faithfully capture the actual proposed text format here.  It also lines up more clearly with the binary format, which is nice.  @pl-semiotics LMKWYT

design/mvp/Binary.md
```diff
@@ -17,197 +17,227 @@ and validation will be present in the [formal specification](../../spec/).
...
-
+* Reused Core binary rules: [`core:name`]
+* The `core:sort` values are chosen to match the discriminant opcodes of
+  [`core:importdesc`] so that `core:exportdesc` (below) is identical.
+* `type` is added to `core:sort` in anticipation of the [type-imports] proposal. Until that
+  proposal, core modules won't be able to actually import or export types, however, the
+  `type` sort is allowed as part of outer aliases (below).
+* `module` and `instance` are added to `core:sort` in anticipation of the [module-linking]
+  proposal, which would add these types to Core WebAssembly. Again, core modules won't be
+  able to actually import or export modules/instances, but they are used for aliases.
```
peterhuene, line 93: The previous AST only allowed core instances for module instantiation arguments, with the expectation that the implementation flatten the instance to a module's two-level imports.

With this update are we expecting to only support instances arguments for module instantiation when/if adopted into the core spec?

That is to say, should implementations continue to support instance arguments and flatten or to not support them until the core spec does?

lukewagner: Yes, core instances would continue to be importable by core modules via `instantiate` just like before.  The general theme here is that we're preemptively including core definitions that would (in the future) be added by module-linking so that we can have a simpler two-level component|core division instead of the previous awkward 3-level component|core-added-by-component-layer|core-1.0 division.

design/mvp/Explainer.md
```diff
@@ -234,141 +284,170 @@ definition, resolved using normal lexical scoping rules. For example, the
...
+                  | <core:structtype>                         (GC proposal)
+                  | <core:arraytype>                          (GC proposal)
+                  | <core:moduletype>
+core:moduletype ::= (module <id>? <core:moduledecl>*)
+core:moduledecl ::= <core:importdecl>
+                  | <core:type>
+                  | <core:exportdecl>
+core:importdecl ::= (import <name> <name> <core:importdesc>)
+core:exportdecl ::= (export <name> <core:exportdesc>)
+core:exportdesc ::= strip-id(<importdesc>)
```
rossberg, line 360: Should this be `core:importdesc`?

design/mvp/Explainer.md
```diff
@@ -234,141 +284,170 @@ definition, resolved using normal lexical scoping rules. For example, the
...
+
+where strip-id(X) parses '(' sort Y ')' when X parses '(' sort <id>? Y ')'
+```
+
+Here, `core:deftype` (short for "defined type") is inherited from the [gc]
+proposal and extended with a `module` type constructor. If module-linking is
+added to Core WebAssembly, an `instance` type constructor would be added as
+well but, for now, it's left out since it's unnecessary. Also, in the MVP,
+validation will reject nested `core:moduletype`, since, before module-linking,
+core modules cannot themselves import or export other core modules.
```
rossberg, line 370: Since the syntax is reusing core:importdesc, I don't think you could even write that?

design/mvp/Explainer.md
```diff
@@ -234,141 +284,170 @@ definition, resolved using normal lexical scoping rules. For example, the
...
+
+The body of a module type contains an ordered list of "module declarators"
+which describe, at a type level, the imports and exports of the module. In a
+module-type context, import and export declarators can both reuse the existing
+[`core:importdesc`] production defined in WebAssembly 1.0, with the only
+difference being that, in the text format, `core:importdesc` can bind an
+identifier for later reuse while `core:exportdesc` cannot.
+
+In preparation for the forthcoming addition of [type-imports] to Core
```
rossberg, line 379: The explanation that follows seems to be independent of type imports.

design/mvp/Explainer.md
```diff
@@ -383,73 +462,92 @@ subtyping. In particular, a `variant` subtype can contain a `case` not present
...
+                          unit ↦ (record)
+                (enum <name>+) ↦ (variant (case <name> unit)+)
+            (option <valtype>) ↦ (variant (case "none") (case "some" <valtype>))
+            (union <valtype>+) ↦ (variant (case "𝒊" <valtype>)+) for 𝒊=0,1,...
+(expected <valtype> <valtype>) ↦ (variant (case "ok" <valtype>) (case "error" <valtype>))
+                        string ↦ (list char)
 ```
 Note that, at least initially, variants are required to have a non-empty list of
 cases. This could be relaxed in the future to allow an empty list of cases, with
 the empty `(variant)` effectively serving as a [bottom type] and indicating
```
rossberg, line 479: It would be an empty type, but not a bottom type -- that's an empty type that also is a subtype of every other, which wouldn't be the case here.

design/mvp/Explainer.md
```diff
@@ -383,73 +462,92 @@ subtyping. In particular, a `variant` subtype can contain a `case` not present
...
-WebAssembly validation rules allow duplicate imports, this means that some
-valid modules will not be typeable and will fail validation if used with the
-Component Model.
+The remaining 3 type constructors in `deftype` use `valtype` to describe
+shared-nothing functions, components and component instances:
+
+The `func` type constructor describes a component-level function definition
+that takes and returns `valtype`. In contrast to [`core:functype`] which, as a
+low-level compiler target for a stack machine, returns zero or more results,
+`functype` always returns a single type, with `unit` being used for functions
```
rossberg, line 488: I'm still a bit saddened by this aymmetry. If languages can bind to tuple types, shouldn't they likewise be able to bind to functions with multiple results?

design/mvp/Explainer.md
```diff
@@ -383,73 +462,92 @@ subtyping. In particular, a `variant` subtype can contain a `case` not present
...
+
+The `type` case of `importdesc`/`exportdesc` describes an imported or exported
+type along with its bounds. The bounds currently only have an `eq` option that
+says that the imported/exported type must be exactly equal to the referenced
+type. There are two main use cases for this in the short-term:
+* Type exports allow a component or interface to associate a name with a
+  structural type (e.g., `(export "nanos" (type (eq u64)))`) which bindings
+  generators can use to generate type aliases (e.g., `typedef uint64_t nanos;`).
+* Type imports and exports allow a component to explicitly specify the
+  type parameters used to monomorphize a generic interface being imported
```
rossberg, line 523: That is rather vague. IIUC, that's is a toolchain thing only expressible outside of the language. So it's not clear how to read it here.

rossberg (commented):
Great, I think this is cleaner and clearer than before!

design/mvp/Explainer.md
```diff
@@ -105,125 +122,158 @@ dead code.
...
+                      | global
+                      | type
+                      | module
+                      | instance
+core:export         ::= (export <name> <core:sortidx>)
+```
+When instantiating a module via `instantiate`, the two-level imports of the
+core modules are resolved as follows:
+1. The first `name` of the import is looked up in the named list of
+   `core:instantiatearg` to select a core module instance.
```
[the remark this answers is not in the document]
lukewagner: Oh right, I guess no reason not to just specialize this in the grammar.  [Fixed](https://github.com/WebAssembly/component-model/pull/29/commits/9ce5ffc7b7737f75d10bf7b6f43732bdf184b8c7)

design/mvp/Explainer.md
```diff
@@ -105,125 +122,158 @@ dead code.
...
+next section.
+
+The syntax for defining component instances is symmetric to core module
+instances, but with an expanded component-level definition of `sort`:
+```
+instance       ::= (instance <id>? <instanceexpr>)
+instanceexpr   ::= (instantiate <componentidx> <instantiatearg>*)
+                 | <export>*
+instantiatearg ::= (with <name> <sortidx>)
+                 | (with <name> (instance <export>*))
```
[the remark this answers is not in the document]
lukewagner: It is indeed explained as syntactic sugar in a paragraph below (and this choice isn't present in Binary.md), but I think it's still useful to put it in this pseudo-AST/text-format grammar.

design/mvp/Explainer.md
lukewagner: Yep! [Fixed](https://github.com/WebAssembly/component-model/pull/29/commits/4f7a0ae7da78399ebc234d23cd70736840e99643)

design/mvp/Explainer.md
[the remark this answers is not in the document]
pl-semiotics: Alternatively, we could leave this in in case core eventually gets single-level imports, and just rule it out in the typechecker?

design/mvp/Explainer.md
```diff
@@ -234,141 +286,170 @@ definition, resolved using normal lexical scoping rules. For example, the
...
+componenttype ::= (component <componentdecl>*)
+instancetype  ::= (instance <instancedecl>*)
+componentdecl ::= <importdecl>
+                | <instancedecl>
+instancedecl  ::= <type>
+                | <alias>
+                | <exportdecl>
+importdecl    ::= (import <name> <importdesc>)
+exportdecl    ::= (export <name> <externdesc>)
+importdesc    ::= bind-id(<externdesc>)
```
pl-semiotics, line 424: There's an asymmetry here with the core types grammar, in that the core declarations define the import versions of these things first, and then the export versions are derived by stripping the ids off, whereas the opposite happens here. I know the core side needs to do it that way because `core:importdesc` is what's available in the grammar; is there a reason to not do the same thing here?

design/mvp/Explainer.md
```diff
@@ -383,73 +464,92 @@ subtyping. In particular, a `variant` subtype can contain a `case` not present
...
+projecting the exports out of imported instances and sharing types with outer
+components. As an example, the following component defines two equivalent
+component types, where the former defines the function type via `type`
+declarator and the latter via `alias` declarator. In both cases, the type is
+given index `0` since component types start with an empty type index space.
+```wasm
+(component $C
+  (type $C1 (component
+    (type (func (param string) (result string)))
+    (import "a" "b" (func (type 0)))
```
pl-semiotics, line 542: We switched this from being a core thing to being a component thing for clarity, so the two-level imports need to go away.

design/mvp/Explainer.md
lukewagner: The nesting would be via the `core:type` case in `core:moduledecl` inside `core:moduletype`.

design/mvp/Explainer.md
```diff
@@ -684,10 +787,10 @@ exports other components:
...
     (export "g" (func (result string)))
   ))
-  (instance $d1 (instantiate (component $D)
+  (instance $d1 (instantiate $D
     (with "c" (instance $c))
   ))
-  (instance $d2 (instantiate (component $D)
+  (instance $d2 (instantiate $D
     (with "c" (instance
       (export "f" (func $d1 "g"))
```
pl-semiotics, line 795: This needs to be `(export "f" (func (func $d1 "g")))`, I think, for the same reason as that other export-alias-inline-in-sortidx example that we mentioned.

pl-semiotics (commented):
I think this makes a lot more sense! I'll make the changes to the interpreter parser and make sure that the examples all still parse, but this version of the type grammar looks much cleaner to me :) (and the fact that it lines up much more nicely with the binary format/AST is very nice!)

design/mvp/Explainer.md
lukewagner: Hah, good point, [fixed](https://github.com/WebAssembly/component-model/pull/29/commits/0bca89702aacf697d769644bdf3e091ca9319390) (although beware, Wikipedia redirects to Bottom type ;-)

design/mvp/Explainer.md
lukewagner: I think the asymmetry is unsurprising when you consider that one is meant to be "source-level" and one is "assembly-level".  Also, from a source-level perspective, multi-return opens up a number of questions that just aren't an issue with the current design: can the multi-return values be named (for symmetry with params, where we do really want the names), but then are the language bindings special-casing the unnamed-single-value case to just return the value, which is a bit irregular, and are some toolchains adding names just for the documentative value but now the language bindings are forced to wrap the returned scalar at runtime and adding overhead (which isn't an issue for named params).

design/mvp/Explainer.md
lukewagner: Yeah, fair point.  Is [this](https://github.com/WebAssembly/component-model/pull/29/commits/85048775a27be713422bdd09ca1217cd880fce89) better?

design/mvp/Explainer.md
```diff
@@ -71,10 +73,7 @@ components can arbitrarily interleave different kinds of definitions.
 
 The `core-prefix` meta-function transforms a grammatical rule for parsing a
 Core WebAssembly definition into a grammatical rule for parsing the same
-definition, but with a `core` token added right after the leftmost paren:
-```
-core-prefix(X) ::= '(' 'core' Y ')' where X = '(' Y ')'
-```
+definition, but with a `core` token added right after the leftmost paren.
```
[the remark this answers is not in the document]
lukewagner: Oh right, good catch.  Since that's the only exception, I [changed](https://github.com/WebAssembly/component-model/pull/29/commits/d7b372676562f322b6b5d7b994db1a92f1bdc30d) `sort` to just be `core <core:sort>` (which is simpler, anyways).

design/mvp/Explainer.md
```diff
@@ -407,35 +407,30 @@ valtype       ::= unit
                 | (union <valtype>+)
                 | (option <valtype>)
                 | (expected <valtype> <valtype>)
-functype      ::= (func <id>? (param <name>? <valtype>)* (result <valtype>))
-typetype      ::= (type <id>? <typebound>)
-typebound     ::= (eq <deftype>)
-componenttype ::= (component <id>? <componentdecl>*)
-instancetype  ::= (instance <id>? <instancedecl>*)
+valtype       ::= <typeidx>
```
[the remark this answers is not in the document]
lukewagner: Yes, that's right; that's the one big hand-wave remaining (mentioned at the top).  (And where you can put a variable, you can also put an inline alias.)

I would've much rather preferred to have a `typeidx`, but in the case of `core module`, it's a `core:typeidx`.  (Binary.md separates out the two cases and so does use `typeidx`/`core:typeidx`.)

design/mvp/Explainer.md
lukewagner: Hah, yeah, I went back and forth on this.  The problem is that the `<id>?` isn't actually present inside the `core:moduletype`/`functype`/`componenttype`/`instancetype` productions to strip so we'd have to re-add them just to strip them which felt net-worse than the asymmetry we have now.

design/mvp/Explainer.md
lukewagner: Ooh, nice catch, thanks!  [fixed](https://github.com/WebAssembly/component-model/pull/29/commits/a56788b48b37d510e79b16fea1d074a762255c81)

design/mvp/Explainer.md
lukewagner: Wow, good catch!  [fixed](https://github.com/WebAssembly/component-model/pull/29/commits/47143e09a92e9368abcbca351d62482b03dbfc5d)

But now I'm starting to wonder: we're going to have, practically speaking, double-sorts-in-a-row 100% of the time with `sortidx` + inline alias sugar, which feels kindof silly and avoidable.  What if instead we say that this inline alias sugar has separate rules for `sortidx` (where you can just write `(sort $instance "name")`) and `Xidx` (where you write `(X $inst "name")`)?

lukewagner: Actually, scanning through the examples, there are a ton of places that already implicitly assume you can use an inline alias in place of a `sortidx`, so this seems like clearly the right choice, so [here's](https://github.com/WebAssembly/component-model/pull/29/commits/f1263c04b5eff2f465703402eb5bc9f5b3ebeeaa) the update to the inline alias description.

design/mvp/Binary.md
```diff
@@ -17,197 +17,231 @@ and validation will be present in the [formal specification](../../spec/).
...
-        | ... other Post-MVP Core definition kinds
-        | 0x02 0x00 ct:<varu32> i:<moduleidx>    => (alias outer ct i (module))
-        | 0x02 0x01 ct:<varu32> i:<componentidx> => (alias outer ct i (component))
-        | 0x02 0x05 ct:<varu32> i:<typeidx>      => (alias outer ct i (type))
+core:alias       ::= sort:<core:sort> target:<core:aliastarget> => (core alias target (sort))
+core:aliastarget ::= 0x00 i:<core:instanceidx> n:<name>         => export i n
+
+alias            ::= sort:<sort> target:<aliastarget>           => (alias target (sort))
+aliastarget      ::= 0x00 i:<instanceidx> n:<name>              => export i n
+                   | 0x01 ct:<u32> idx:<u32>                    => outer ct idx
```
peterhuene, line 108:
```suggestion
                   | 0x01 ct:<u32> si:<sortidx>                 => outer ct si
```

peterhuene: Erm disregard, misread the grammar.

design/mvp/Binary.md
```diff
@@ -17,197 +17,231 @@ and validation will be present in the [formal specification](../../spec/).
...
+                      | 0x01 e*:vec(<core:export>)                         => e*
+core:instantiatearg ::= n:<name> 0x11 i:<instanceidx>                      => (with n (instance i))
+core:sortidx        ::= sort:<core:sort> idx:<u32>                         => (sort idx)
+core:sort           ::= 0x00                                               => func
+                      | 0x01                                               => table
+                      | 0x02                                               => memory
+                      | 0x03                                               => global
+                      | 0x04                                               => type
+                      | 0x10                                               => module
+                      | 0x11                                               => instance
```
peterhuene, line 68: There are currently some implementations for "tag" types from the exception handling proposal that are using this `core:sort` (0x04).

Perhaps the following?

```suggestion
                      | 0x10                                               => type
                      | 0x11                                               => module
                      | 0x12                                               => instance
```

design/mvp/Binary.md
```diff
@@ -17,197 +17,231 @@ and validation will be present in the [formal specification](../../spec/).
...
-* Module, component and instance types create fresh type index spaces that are
-  populated and referenced by their contained definitions. E.g., for a module
-  type that imports a function, the `import` `moduletype-def` must be preceded
-  by either a `type` or `alias` `moduletype-def` that adds the function type to
-  the type index space.
-* Currently, the only allowed form of `alias` in instance and module types
-  is `(alias outer ct li (type))`. In the future, other kinds of aliases
-  will be needed and this restriction will be relaxed.
+* Validation of `valtype` requires the `typeidx` to refer to a `defvaltype`.
+* Validation of `moduledecl` (currently) only allows `outer` `type` `alias`
```
peterhuene, line 206:
```suggestion
* Validation of `instancedecl` (currently) only allows `outer` `type` `alias`
```

design/mvp/Binary.md

design/mvp/Binary.md
```diff
@@ -17,197 +17,231 @@ and validation will be present in the [formal specification](../../spec/).
...
+componenttype ::= 0x41 cd*:vec(<componentdecl>)        => (component cd*)
+instancetype  ::= 0x42 id*:vec(<instancedecl>)         => (instance id*)
+componentdecl ::= 0x00 id:<importdecl>                 => id
+                | id:<instancedecl>                    => id
+instancedecl  ::= 0x01 t:<type>                        => t
+                | 0x02 a:<alias>                       => a
+                | 0x03 ed:<exportdecl>                 => ed
+importdecl    ::= n:<name> ed:<externdesc>             => (import n ed)
+exportdecl    ::= n:<name> ed:<externdesc>             => (export n ed)
+externdesc    ::= 0x00 0x10 i:<core:typeidx>           => (core module (type i))
```
peterhuene, line 193:
```suggestion
externdesc    ::= 0x00 0x11 i:<core:typeidx>           => (core module (type i))
```

design/mvp/Explainer.md
rossberg: > I think the asymmetry is unsurprising when you consider that one is meant to be "source-level" and one is "assembly-level".

That's insinuating that all source languages have single return values, which of course isn't the case. One could argue that component types use a smallest common denominator approach, but that's not true either, since they feature things like variants etc.

I understand what you mean regarding bindings and names, but is that really an asymmetry? If you simply allowed the names, languages who have no use for named return values could simply ignore the names and handle them positionally. The same is actually true for parameters: there are languages where parameter names are relevant at call sites (e.g., named parameters can be used to reorder the argument list), but most languages just do calls positionally and parameter names from the types are irrelevant beyond documentation value.

design/mvp/Binary.md
lukewagner: Oh right, thanks!

design/mvp/Explainer.md
lukewagner: > One could argue that component types use a smallest common denominator approach, but that's not true either, since they feature things like variants etc.

I know, it's a fuzzy least common denominator, but in the particular case of return values, single-return (with good tuple integration) is overwhelmingly more common.

> f you simply allowed the names, languages who have no use for named return values could simply ignore the names and handle them positionally. The same is actually true for parameters:

I think the asymmetry arises due to the producing nature of return types vs. the consuming nature of param types.  The language binding for params can easily consume *either* named or positional parameters, but the language binding for results must pick one, which has consequences either way.  In particular, all languages will have **some** way to represent named return values (e.g., as a dictionary) which will have some fixed overhead vs. the unnamed scalar binding, so the language bindings will have to just pick something, biasing one way or the other.  With single-return, there would be a clear directive for language bindings to produce a name-bearing value for `record` results (and of course unnamed for scalar or `tuple` results), allowing interface authors to more-reliably say what they want.

lukewagner:
Excellent feedback and suggestions so far; thanks all!  Things seem to have quieted down, so I'll hold until the end of the week and merge if it's still quiet.  (We can keep talking about the pre-existing (relative to this PR) single-vs-multi-return issue separately.)

design/mvp/Explainer.md
rossberg: A similar choice is also necessary for parameters sometimes: there are a number of languages (especially functional ones) that only have single parameter functions. Multiple parameters are then represented either with tuples or by currying. These are two mutually exclusive choices, and a mapping for such a language will have to pick one over the other.

Likewise, there are languages where named vs positional parameter definitions is a mutually exclusive choice.

So I wouldn't count either of this as a qualitative difference.

design/mvp/Explainer.md
pl-semiotics: That makes sense to me! I thought we might want to do that when I first brought it up, and as it keeps showing up I also have been leaning further in that direction.

design/mvp/Binary.md
```diff
@@ -17,197 +17,231 @@ and validation will be present in the [formal specification](../../spec/).
...
-               | n:<name> 0x02 i:<instanceidx>                     => n (instance i)
-               | n:<name> 0x03 f:<funcidx>                         => n (func f)
-               | n:<name> 0x04 v:<valueidx>                        => n (value v)
-               | n:<name> 0x05 t:<typeidx>                         => n (type t)
-export       ::= a:<componentarg>                                  => (export a)
-name         ::= n:<core:name>                                     => n
+core:instance       ::= ie:<instance-expr>                                 => (instance ie)
+core:instanceexpr   ::= 0x00 m:<moduleidx> arg*:vec(<core:instantiatearg>) => (instantiate m arg*)
+                      | 0x01 e*:vec(<core:export>)                         => e*
+core:instantiatearg ::= n:<name> 0x11 i:<instanceidx>                      => (with n (instance i))
```
peterhuene, line 60: More fallout from reindexing the core sorts:

```suggestion
core:instantiatearg ::= n:<name> 0x12 i:<instanceidx>                      => (with n (instance i))
```

design/mvp/Explainer.md
```diff
@@ -462,52 +575,50 @@ and out-of-line type definitions:
...
+* `lift` wraps a core function (of type `core:functype`) to produce a component
+  function (of type `functype`) that can be passed to other components.
+* `lower` wraps a component function (of type `functype`) to produce a core
+  function (of type `core:functype`) that can be imported and called from Core
+  WebAssembly code inside the current component.
+
+Canonical definitions specify one of these two wrapping directions, the function
+to wrap and a list of configuration options:
+```
+canon    ::= (canon lift core-prefix(<core:funcidx>) <functype> <canonopt>* (func <id>?))
```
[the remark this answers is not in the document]
lukewagner: Oh wow, good catch!  So, in updating `<functype>` to follow the same explicit "inline type or (type <typeidx>)" scheme as elsewhere, I realized that we could make `canon lift` a lot more symmetric with imports by having the function type go with the sort and id.  So [this change](https://github.com/WebAssembly/component-model/pull/29/commits/e5b93c4985d675c058663c9b077092642b8f7f24) moves around the functype (and allows type indices).  If you look at the corresponding changes in the examples, it seems clear this is what it should've been all along.

lukewagner:
Ok, merging this, and we can continue to iterate by filing new issues.  Next up, `future` and `stream` :)

PR #33, opened 2022-05-28
Fix inverted func/alias syntax examples.

peterhuene:
This PR fixes the syntax examples for the inverted forms of canon
definitions and function aliases.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -307,7 +307,7 @@ is desugared into:
 Lastly, for symmetry with [imports][func-import-abbrev], aliases can be written
 in an inverted form that puts the sort first:
 ```wasm
-(func $f (import "i" "f"))            ≡ (import "i" "f" (func $f))             (WebAssembly 1.0)
+(func $f (import "i" "f") ...type...) ≡ (import "i" "f" (func $f ...type...))       (WebAssembly 1.0)
 (func $g (alias export $i "g1"))      ≡ (alias export $i "g1" (func $g))
 (core func $g (alias export $i "g1")) ≡ (core alias export $i "g1" (func $g))
 ```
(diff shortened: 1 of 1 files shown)


design/mvp/Explainer.md
```diff
@@ -675,9 +675,9 @@ stack-switching in component function signatures.
 Similar to the `import` and `alias` abbreviations shown above, `canon`
 definitions can also be written in an inverted form that puts the sort first:
 ```wasm
-      (func $f ...type... (import "i" "f")) ≡ (import "i" "f" (func $f ...type...))       (WebAssembly 1.0)
-      (func $h ...type... (canon lift ...)) ≡ (canon lift ... (func $h ...type...))
-(core func $h ...type... (canon lower ...)) ≡ (canon lower ... (core func $h ...type...))
+(func $f (import "i" "f") ...type...) ≡ (import "i" "f" (func $f ...type...))       (WebAssembly 1.0)
+(func $g ...type... (canon lift ...)) ≡ (canon lift ... (func $g ...type...))
```
peterhuene, line 679: Question: do we really want the types to come before the `(canon lift ...)` for this inverted form as it seems out-of-line with the other forms (inline imports come before the types and `canon lower` shouldn't have types because the lowered canonical signature is inferred from the component function type)?

I'm implementing parsing of the inverted forms in the `wast` crate right now, so I need to know what tokens to expected in a lookahead to disambiguate the forms.

peterhuene:
While I somewhat disagree that it's symmetrical with inlined imports as the latter has the types appear after the "inlined" part (e.g. `(func (import "" "") type...)`), I can understand that if we eventually want to replace the `(canon lift ...)` part in `(func type... (canon lift ...))` with instructions post MVP, it only makes sense to have it appear _after_ the types since its semantic meaning is "the body is whatever instructions make the canonical ABI happen", so I'mon board with that :+1:.

While that resolves the random question I had, this PR didn't assume we were going to make that change.

Should I leave the types in the `canon lower` inverted form (although I suspect we should clarify it'd be `core:valtype...`)?. I assume that we won't currently accept types in `canon lower`'s inverted form since we always derive the lowered signature to match the canonical ABI, but in a post-MVP world that's where they might go?

Personally I'd rather have the forms specified in what's acceptable for the MVP, which means no types in `canon lower`'s inverted form.

lukewagner:
Ah hah, you're right about the inverted form of `import` being different than what I wrote!  So I guess to be *actually* symmetric with `import`, we'd have to write  `(func (canon lift ...) ...types...)` which maybe looks goofy... but seems workable.  Then, in the same way as you can, in core wasm 1.0, write both `(func (import ...) ...types...)` *and* `(func ...types... ...instr*...)`, we could, still, in the future, add a component-level `(func ...types... ...instr*...)` that co-exists with `(func (canon lift ...) ...types...)`.  Does that make sense and/or look better from a parser impl POV?

peterhuene:
I think I've fully convinced myself to keep it the way it is currently defined by the logic that the inverted forms of `canon` are really just body placeholders (i.e. "do what the canonical ABI requires _here_") for eventual support of function bodies for custom adapters; that is to say I'm fine with how we have it now in body position and currently have very weak opposition to instead putting `(canon lift)` in the position of the other inline forms 😄.

If we do change it so that `(canon lift)` appears in the position of the other inline forms, I don't see any problems with doing so even if we expect a signature at that position for custom adapters in the future.

lukewagner (approved):
Ok, makes sense!  If it's not a parser problem, then it looks nice the way it is now, so this PR looks great.  Thanks!

PR #36, opened 2022-06-03
Add <core:type> to <componentdecl>

lukewagner:
@peterhuene spotted this was missing and necessary for a component type to be able to define a core module type so it can be imported/exported.


peterhuene (approved):
Looks great, thanks!

PR #37, opened 2022-06-05
Update post-return validation in binary spec.

peterhuene:
This PR updates the post-return validation to align it with what is
specified in the explainer.

The post-return option should accept the return values from the core function
being lifted as parameters and itself have no return values.


lukewagner:
Oops, thanks!

PR #38, opened 2022-06-06
Clarify type validation rules in the binary format.

peterhuene:
This PR clarifies that names within certain type definitions must be
unique.

It also adds validation of the refines clause in a variant case.


lukewagner (approved):
Looks great, thanks!

PR #42, opened 2022-06-08
Adjust canonical ABI option syntax to be less verbose

alexcrichton:
Currently the syntax for specifying a memory with a canonical option is:

    (memory (core memory $memory))

and optionally instance alias sugar can also be used:

    (memory (core memory $libc "memory"))

This PR proposes changing these two syntaxes to:

    (memory $memory)
    (memory $libc "memory")

with the theory that the "core" part is already implied by the canonical
option itself and otherwise saying "memory" twice is redundant.


lukewagner:
Totally agreed the current syntax feels overly verbose in these scenarios so thanks for bringing this up!

So, first, I was already on the edge about whether we needed the `core-prefix` here (given the static context of "this is definitely a core index"), so agreed we should drop that part.

If we wanted to drop the `(sort ...)` that wraps the `<core:instanceidx> <name>`, then I think we'd need to define some custom syntactic sugar just for the `instantiate` instruction because, as is, if the AST has, e.g., `(realloc <core:funcidx>)`, then the inline alias sugar only lets you substitute in a `(func $inst "exp")` for the `<core:funcidx>`, producing `(realloc (func $inst "exp"))`.  `memory` is a more ambiguous case since `memory` is a case of `sort`, but in this context, the `memory` token is serving as a label of an immediate, which could have just as well been named not-`memory`, e.g. `into`.  Because the sugar-substitution is purely lexical (it doesn't care about `sort` vs. immediate), you could say that the inline alias sugar *already* allows you to write what you're proposing in this PR.  (FWIW, I tried to think of a better immediate label than `memory` to this wasn't a special case, but I couldn't find one: `into` only makes sense for lifting; `mem` feels contrived.)

So: drop the `core-prefix` and but keep the `(func ...)` for `realloc`/`post-return`?

alexcrichton:
👍 sounds reasonable to me!

design/mvp/Explainer.md
```diff
@@ -616,9 +616,9 @@ canon    ::= (canon lift core-prefix(<core:funcidx>) <canonopt>* bind-id(<extern
 canonopt ::= string-encoding=utf8
            | string-encoding=utf16
            | string-encoding=latin1+utf16
-           | (memory core-prefix(<core:memidx>))
-           | (realloc core-prefix(<core:funcidx>))
-           | (post-return core-prefix(<core:funcidx>))
+           | (memory <core:memidx>)
+           | (realloc (func <core:funcidx>))
+           | (post-return (func <core:funcidx>))
```
lukewagner, line 621:
```suggestion
           | (realloc <core:funcidx>)
           | (post-return <core:funcidx>)
```
The idea being that you can either write `(canon lift $f (realloc $r) (func $l))` or, via inline-alias-sugar, `(canon lift $f (realloc (func $i "r")) (func $l))`.

lukewagner (approved):
Thanks! Could you also update `examples/SharedEverythingDynamicLinking.md` (where, it appears, I wasn't even already fully following the explainer...).

alexcrichton:
Sure thing, should be updated now I believe.

lukewagner:
Thanks!

PR #50, opened 2022-06-17
clarify inline sugar syntax in Explainer.md

dicej:
I had trouble understanding the examples in the `Aliases` section, partly due to not seeing how the abstract syntax in the prose matched up with the concrete syntax in the examples.  This change will hopefully save others from the same confusion.

Also, I corrected a couple of misspellings.


lukewagner:
Ah, that is better; thanks!

Issue #55, opened 2022-06-22
Clarifying trap semantics


alexcrichton:
In implementing post-return I realized that I ended up having a few questions about the semantics of traps with components which may be worth calling out explicitly in the canonical ABI explainer or otherwise just resolving my questions here:

* If a core wasm instance as part of a component instance generates a trap, is the component intended to still be usable? Or should the component be completely blocked from all future calls?
* What are the semantics of validation errors in lifting/lowering? As specified if the Python spec code is read as a trap basically throwing an exception, then I think it has these consequences:
  * In `canon_lift` if the `lower` operation traps/errors then `may_leave` is set to `False` which would trigger the prior assert the next time the function is called (if it could even be called)
  * In `canon_lift` if the `lift` operation traps/errors then `may_enter` is left as `False` and all future calls to `canon_lift` will trap
  * In `canon_lower` there are similar consequences as `canon_lift` but in reverse.

I vaguely recall the idea that when an instance traps it poisons the entire component instance as being impossible to enter, but I'm not sure if that's what's expected (since otherwise that would transitively poison everything right now I think and only an embedder could catch the error). Otherwise though I didn't see many other references to traps and recovery in the current canonical ABI explainer.

lukewagner:
Great questions:
* The intention is that, after a trap, the whole component instance (and even all transitively-linked component instances) are locked down.  And you're right that there's no way to stop propagation in wasm before it hits the host.  (The ability to contain traps is this Post-MVP "blast zone" feature for expressing and dealing with partial failure.)
* The Canonical ABI Python code doesn't implement this "lockdown on trap" behavior (the `may_enter`/`may_leave` stuff is meant to *cause* traps, not implement the lockdown *after* a trap).  I was already thinking that I need to add a new top-level Python spec-function to define *whole components* (not just individual imports/exports), and I suppose this is where the trap semantics would go.

alexcrichton:
Ok that sounds reasonable. A question along those lines though: in some cases a trap is theoretically recoverable for example with an OOM trap from a `realloc` function. In such a situation is it worth having this be recoverable or would this also simply poison the whole instance and everything it's linked to?

lukewagner:
Resolved by #57

PR #57, opened 2022-06-28
Tweak may_enter rules to ensure lockdown-on-trap

lukewagner:
As discussed in #55.  I realized that just broadening the scope of the `may_enter`-clearing achieved the desired effect.


alexcrichton (approved):
Looks good to me! Thanks!

PR #62, opened 2022-07-01
Loosen the reentrance rules a bit for parent/child scenarios

lukewagner:
This PR loosens the reentrance rules to allow `canon lift` to be called from inside a component instance such as when a parent component is called by a child component.  For example, in this example:
```wasm
(component
  (core module $Core1 ...)
  (core instance $core1 (instantiate $Core1))
  (func $func1 (canon lift (core func $core1 "func")))
  (component $Child ...)
  (instance $child (instantiate $Child (with "func1" (func $func1))))
  (core func $child_func (canon lower (func $child "func")))
  (core module $Core2 ...)
  (core instance $core2 (instantiate $Core2 (with "child" (instance (export "func" (func $child_func))))))
  (func $func2 (canon lift (core func $core2 "func")))
  (export "func2" (func $func2))
)
```
as previously written, `$func2` would clear `may_enter` so that `$func1` would trap when called by `$child`.  With this PR, the idea is that `$func1` wouldn't check `may_enter` because `called_as_export` would be `False`.  A more realistic example of when you'd want to do this is that the parent component uses `$Core1` and `$Core2` to virtualize a filesystem for `$Child`, including passing preopens into `$Child`'s exports.

One consequence of this relaxation is that same-component-instance lift+lower would be possible and such cases would need to be compiled more pessimistically (although in a way that only impacted this (rare) case).  The PR explains more.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1132,9 +1132,12 @@ the outside world through an export.
 
 Given the above closure arguments, `canon_lift` is defined:
 ```python
-def canon_lift(callee_opts, callee_instance, callee, functype, args):
-  trap_if(not callee_instance.may_enter)
-  callee_instance.may_enter = False
+def canon_lift(callee_opts, callee_instance, callee, functype, args, called_as_export):
+  if called_as_export:
+    trap_if(not callee_instance.may_enter)
+    callee_instance.may_enter = False
+  else:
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):
Seems reasonable to me, and to confirm, it's possible to optimize a same-component lift/lower if the linear memories they reference are provably disjoint, right?

lukewagner:
Yep!

PR #68, opened 2022-07-12
Say lists of numbers produce typed arrays in the JS API

### LINKED ISSUE #66: Mapping to `Uint8Array`
It seems like the component model's recommended way of representing bytes is `list<u8>`, which is reasonable. However, in the JavaScript binding, lists always map to JS arrays, which seems pretty inefficient for passing bytes around. Is there any plan for defining APIs which take/return `Uint8Array`s in the JS binding, whether that be with `list<u8>` or some other type?

(Side-note - I was originally going to ask about representing TypedArrays in general, but it doesn't really seem doable for anything other than `Uint8Array` and `Int8Array`, since WebAssembly is little endian and TypedArrays are native endian.)

lukewagner:
Resolves #66

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -955,7 +955,7 @@ At a high level, the additional coercions would be:
 | `char` | same as [`USVString`] | same as [`USVString`], throw if the USV length is not 1 |
 | `record` | TBD: maybe a [JS Record]? | same as [`dictionary`] |
 | `variant` | TBD | TBD |
-| `list` | same as [`sequence`] | same as [`sequence`] |
+| `list` | create a typed array copy for number types; otherwise produce a JS array (like [`sequence`]) | same as [`sequence`] |
 | `string` | same as [`USVString`]  | same as [`USVString`] |
 | `tuple` | TBD: maybe a [JS Tuple]? | TBD |
 | `flags` | TBD: maybe a [JS Record]? | same as [`dictionary`] of optional `boolean` fields with default values of `false` |


design/mvp/Explainer.md
```diff
@@ -955,7 +955,7 @@ At a high level, the additional coercions would be:
 | `char` | same as [`USVString`] | same as [`USVString`], throw if the USV length is not 1 |
 | `record` | TBD: maybe a [JS Record]? | same as [`dictionary`] |
 | `variant` | TBD | TBD |
-| `list` | same as [`sequence`] | same as [`sequence`] |
+| `list` | create a typed array copy for number types; otherwise produce a JS array (like [`sequence`]) | same as [`sequence`] |
```
Liamolucko, line 958: 👍

Good idea leaving the lowering as-is; that basically gets rid of any concerns about typed arrays being more annoying to construct.

Liamolucko (approved):

design/mvp/Explainer.md
lukewagner: Cool!  And, just to check my understanding: the Web IDL rules inherited by lowering would accept *both* typed and JS arrays for a `sequence`, yes?

Liamolucko: Yes, any iterable can be converted to a `sequence`. (https://webidl.spec.whatwg.org/#es-sequence)

lukewagner: Sweet, thanks!

Issue #41, opened 2022-06-08
(Re-)Consider multi-return vs unit


lukewagner:
In #29, we had an unresolved [comment thread](https://github.com/WebAssembly/component-model/pull/29#discussion_r870174315) about whether we should make component-level function results symmetric with parameters.  The question was independent (pre-existing) of #29, so that PR merged and this issue represents the continued discussion.

One interesting new technical argument just came up via @peterhuene's work: Component-level `start` sections calling functions that return "nothing" technically always return `unit`.  If we treated `unit` uniformly (pushed a new `unit` value into the value index space), then the "linearity" requirement (each value must be consumed exactly once) would mean we have to consume that `unit`... which would then produce a new `unit` (ad infinitum).  Thus, we have to special-case `unit` in the validation rules (so it doesn't push a value index) which is irregular and adds validation complexity.  It's not a massive problem, but it does add a point to the multi-return column (which can return an empty list of values).

lukewagner:
#49 raises an interesting case where having an empty return (returning `undefined` to a JS caller) would produce a more-regular JS interface than a `unit`-return (returning `null` to a JS caller).

(I haven't forgotten about this issue; I just haven't had time to think and discuss more about it yet.)

lukewagner:
Here's a third argument in favor of multi-return that I recently realized:

To support runtime instantiation of components (going beyond the declarative all-up-front instantiation we currently have), I've been imagining that we'd define a new canonical built-in `(canon instantiate $Component (func $new))` that defines a function `$new` that, when called, creates a new instance of `$Component`.  If `$Component` defines any function exports, then the return type of `$new` would be a `(handle $Component)` (with `$Component` serving as a resource type).  But if `$Component` exclusively exports values, then `$new` can simply be a normal function returning values (thereby realizing the [WASI "command" pattern](https://github.com/WebAssembly/WASI/issues/13)<sup>[1](https://docs.google.com/presentation/d/1MNVOZ8hdofO3tI0szg_i-Yoy0N2QPU2C--LzVuoGSlE/edit#slide=id.g13630a5e721_0_68)</sup>).  But now the asymmetry of parameters/results bites again: value import names can turn into parameter names but value export names must either be dropped or wrapped into a `record`.

peterhuene:
I think the above makes a lot of sense to me. I don't have any objections with the other proposed changes.

Regarding option 2 from above:

> a param list is either all-named or contains a single unnamed type

[...]

Issue #49, opened 2022-06-17
Should `unit` map to `null` or `undefined`?


Liamolucko:
The explainer currently defines `unit` to map to `null` in JS. While attempting to implement that in `wit-bindgen` (which currently uses `undefined`), I realised that that would mean that functions with no return value (desugared to `unit`) would return `null` instead of `undefined`. That seems quite unnatural for JS, where a function with no return value gives a result of `undefined`. So, it might be a better idea for it to map to `undefined` instead.

#41 would also be a solution to that.

Aside from that one case, I don't have any preference for one over the other.

Liamolucko:
A reason for `null`, though, is that lowering dictionary will throw a type error if a required field is `undefined`, without referencing the conversion for the particular type. That means that lowering a record with `undefined` for a `unit` field would fail (under the current definition of record lowering), which would be quite strange if `undefined` was what `unit`s mapped to in the other direction.

lukewagner:
Right, your second comment is in the spirit of why `null` was chosen (as an intentional singular value vs. "not present").  But your first comment is a great reason why that is awkward and a nice supporting argument in favor of multi-return; I'll mention this in that issue.

Liamolucko:
In lieu of multi-return, would it make sense to special case `unit` to map to `undefined` when used as a return value?

PR #69, opened 2022-07-14
Switch (back) to multi-return, remove unit, s/expected/result/

lukewagner:
This PR switches back to multi-return and, for the reasons outlined in #41, makes several associated changes:
* `unit` is removed (empty return can be used instead)
* `variant` and `expected` case payloads are changed to ~~lists of values~~contain either 0 or 1 `valtype`
* `expected` is renamed to `result`
* parameter/result lists can either be all-named or have a single unnamed type

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -180,7 +180,7 @@ defvaltype    ::= pvt:<primvaltype>                        => pvt
                 | 0x6b t:<valtype>                         => (option t)
                 | 0x6a t*:vec(<valtype>) u*:vec(<valtype>) => (result t* (error u*))
 namedtype     ::= n:<name> t:<valtype>                     => (field n t)
-case          ::= nt*:vec(<namedtype>) 0x0                 => (case nt*)
+case          ::= n:<name> t*:vec(<valtype>) 0x0           => (case n t*)
                 | n:<name> t*:vec(<valtype>) 0x1 i:<u32>   => (case n t* (refines case-label[i]))
 valtype       ::= i:<typeidx>                              => i
                 | pvt:<primvaltype>                        => pvt
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -181,7 +181,7 @@ defvaltype    ::= pvt:<primvaltype>                        => pvt
                 | 0x6a t*:vec(<valtype>) u*:vec(<valtype>) => (result t* (error u*))
 namedtype     ::= n:<name> t:<valtype>                     => (field n t)
 case          ::= nt*:vec(<namedtype>) 0x0                 => (case nt*)
-                | nt*:vec(<namedtype>) 0x1 i:<u32>         => (case nt* (refines case-label[i]))
+                | n:<name> t*:vec(<valtype>) 0x1 i:<u32>   => (case n t* (refines case-label[i]))
 valtype       ::= i:<typeidx>                              => i
                 | pvt:<primvaltype>                        => pvt
 functype      ::= 0x40 p*:<prlist> r*:<prlist>             => (func (param p)* (result r)*)
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -151,59 +151,61 @@ Notes:
   means that the maximum `ct` in an MVP `alias` declarator is `1`.
 
 ```
-type          ::= dt:<deftype>                             => (type dt)
-deftype       ::= dvt:<defvaltype>                         => dvt
-                | ft:<functype>                            => ft
-                | ct:<componenttype>                       => ct
-                | it:<instancetype>                        => it
-primvaltype   ::= 0x7f                                     => bool
-                | 0x7e                                     => s8
-                | 0x7d                                     => u8
-                | 0x7c                                     => s16
-                | 0x7b                                     => u16
-                | 0x7a                                     => s32
-                | 0x79                                     => u32
-                | 0x78                                     => s64
-                | 0x77                                     => u64
-                | 0x76                                     => float32
-                | 0x75                                     => float64
-                | 0x74                                     => char
-                | 0x73                                     => string
-defvaltype    ::= pvt:<primvaltype>                        => pvt
-                | 0x72 nt*:vec(<namedtype>)                => (record (field nt)*)
-                | 0x71 case*:vec(<case>)                   => (variant case*)
-                | 0x70 t:<valtype>                         => (list t)
-                | 0x6f t*:vec(<valtype>)                   => (tuple t*)
-                | 0x6e n*:vec(<name>)                      => (flags n*)
-                | 0x6d n*:vec(<name>)                      => (enum n*)
-                | 0x6c t*:vec(<valtype>)                   => (union t*)
-                | 0x6b t:<valtype>                         => (option t)
-                | 0x6a t*:vec(<valtype>) u*:vec(<valtype>) => (result t* (error u*))
-namedtype     ::= n:<name> t:<valtype>                     => (field n t)
-case          ::= n:<name> t*:vec(<valtype>) 0x0           => (case n t*)
-                | n:<name> t*:vec(<valtype>) 0x1 i:<u32>   => (case n t* (refines case-label[i]))
-valtype       ::= i:<typeidx>                              => i
-                | pvt:<primvaltype>                        => pvt
-functype      ::= 0x40 p*:<prlist> r*:<prlist>             => (func (param p)* (result r)*)
-prlist        ::= 0x00 t:<valtype>                         => [t]
-                | 0x01 nt*:vec(<namedtype>)                => nt*
-componenttype ::= 0x41 cd*:vec(<componentdecl>)            => (component cd*)
-instancetype  ::= 0x42 id*:vec(<instancedecl>)             => (instance id*)
-componentdecl ::= 0x03 id:<importdecl>                     => id
-                | id:<instancedecl>                        => id
-instancedecl  ::= 0x00 t:<core:type>                       => t
-                | 0x01 t:<type>                            => t
-                | 0x02 a:<alias>                           => a
-                | 0x04 ed:<exportdecl>                     => ed
-importdecl    ::= n:<name> ed:<externdesc>                 => (import n ed)
-exportdecl    ::= n:<name> ed:<externdesc>                 => (export n ed)
-externdesc    ::= 0x00 0x11 i:<core:typeidx>               => (core module (type i))
-                | 0x01 i:<typeidx>                         => (func (type i))
-                | 0x02 t:<valtype>                         => (value t)
-                | 0x03 b:<typebound>                       => (type b)
-                | 0x04 i:<typeidx>                         => (instance (type i))
-                | 0x05 i:<typeidx>                         => (component (type i))
-typebound     ::= 0x00 i:<typeidx>                         => (eq i)
+type          ::= dt:<deftype>                         => (type dt)
+deftype       ::= dvt:<defvaltype>                     => dvt
+                | ft:<functype>                        => ft
+                | ct:<componenttype>                   => ct
+                | it:<instancetype>                    => it
+primvaltype   ::= 0x7f                                 => bool
+                | 0x7e                                 => s8
+                | 0x7d                                 => u8
+                | 0x7c                                 => s16
+                | 0x7b                                 => u16
+                | 0x7a                                 => s32
+                | 0x79                                 => u32
+                | 0x78                                 => s64
+                | 0x77                                 => u64
+                | 0x76                                 => float32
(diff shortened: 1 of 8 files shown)


design/mvp/Explainer.md
```diff
@@ -440,23 +440,26 @@ deftype       ::= <defvaltype>
...
-                | (variant (case <id>? <name> <valtype> (refines <id>)?)+)
+                | (variant (case <id>? <name> <valtype>* (refines <id>)?)+)
                 | (list <valtype>)
                 | (tuple <valtype>*)
                 | (flags <name>*)
                 | (enum <name>+)
                 | (union <valtype>+)
                 | (option <valtype>)
-                | (expected <valtype> <valtype>)
+                | (result <valtype> (error <valtype>*)?)
```
sunfishcode, line 455: Is this missing a `*` on the first `<valtype>`?

lukewagner: Yep!  [Fixed](https://github.com/WebAssembly/component-model/pull/69/commits/88c31c28858c0bbcde5def806be91c75c8eaa612)

design/mvp/Binary.md
```diff
@@ -151,60 +151,59 @@ Notes:
...
+                | 0x71 case*:vec(<case>)                   => (variant case*)
+                | 0x70 t:<valtype>                         => (list t)
+                | 0x6f t*:vec(<valtype>)                   => (tuple t*)
+                | 0x6e n*:vec(<name>)                      => (flags n*)
+                | 0x6d n*:vec(<name>)                      => (enum n*)
+                | 0x6c t*:vec(<valtype>)                   => (union t*)
+                | 0x6b t:<valtype>                         => (option t)
+                | 0x6a t*:vec(<valtype>) u*:vec(<valtype>) => (result t* (error u*))
+namedtype     ::= n:<name> t:<valtype>                     => (field n t)
+case          ::= nt*:vec(<namedtype>) 0x0                 => (case nt*)
```
peterhuene, line 183: To match the text format:

```suggestion
case          ::= n:<name> t*:vec(<valtype>) 0x0           => (case n t*)
```

design/mvp/Binary.md
```diff
@@ -151,60 +151,59 @@ Notes:
...
+                | nt*:vec(<namedtype>) 0x1 i:<u32>         => (case nt* (refines case-label[i]))
```
peterhuene, line 184:
```suggestion
                | n:<name> t*:vec(<valtype>) 0x1 i:<u32>   => (case n t* (refines case-label[i]))
```

design/mvp/Explainer.md
```diff
@@ -440,23 +440,26 @@ deftype       ::= <defvaltype>
...
                 | <instancetype>
-defvaltype    ::= unit
-                | bool
+defvaltype    ::= bool
                 | s8 | u8 | s16 | u16 | s32 | u32 | s64 | u64
                 | float32 | float64
                 | char | string
                 | (record (field <name> <valtype>)*)
-                | (variant (case <id>? <name> <valtype> (refines <id>)?)+)
+                | (variant (case <id>? <name> <valtype>* (refines <id>)?)+)
```
Liamolucko, line 448: Is this still supposed to be limited to 0 or 1 payload types, or can you now have multiple payloads?

peterhuene: I think this is perhaps meant to be `+`?

peterhuene: I'm also trying to determine the utility of representing variant/result case payloads with a list of types.

Is this intended to be semantically different than having a single payload type that's a tuple?

design/mvp/Binary.md

design/mvp/Explainer.md
lukewagner: Hah, in explaining why I went with 0..N, I talked myself into 0..1.  I originally thought that the argument in favor of `0..N` was that, when you consider the automatic language binding of `variant`, a number of languages with algebraic data types allow their case constructors to be n-ary in a way that is syntactically distinct from a unary constructor taking a tuple (just from a quick scan of the [wikipedia page](https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(algebraic_data_type)): Coq, Haskell, Haxe, Idris, Kotlin, Rust, Scala, Swift, TypeScript)).  In particular, the [way TypeScript does ADTs](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions) makes it *quite* visible.  But to generate these idiomatic binding, we'd need the individual fields of the case payloads to have *names* (just like function results).  Sure, we could *further* generalize case payloads to have names, but now we're really duplicating `record` and so probably it's better to just say to use `tuple`/`record` and then the automatic language bindings can choose, if they want, to "flatten" `(variant (case (tuple ...))))` and `(variant (case (record ...))))` into the language's n-ary ADT constructors.  Does that make sense?

peterhuene: That makes sense to me. Thanks for cluing me in on the reasoning behind the change.

design/mvp/Binary.md
peterhuene: I see the commit for the other variant to `case`, but not this one?

lukewagner: Oh whoops, I thought they were dups, but I see they were just on adjacent lines.

peterhuene: Yeah, sorry about that, should have did a multiline suggestion, but I forgot about the other variant after I had already commented on this one 🙃

design/mvp/Explainer.md
```diff
@@ -440,23 +440,26 @@ deftype       ::= <defvaltype>
...
-                | (expected <valtype> <valtype>)
+                | (result <valtype>* (error <valtype>*)?)
 valtype       ::= <typeidx>
                 | <defvaltype>
-functype      ::= (func (param <name>? <valtype>)* (result <valtype>))
+functype      ::= (func <paramlist> <resultlist>)
+paramlist     ::= (param <name> <valtype>)*
+                | (param <valtype>)
+resultlist    ::= (result <name> <valtype>)*
+                | (result <valtype)
```
sunfishcode, line 462: This `<valtype` is missing closing angle bracket.

design/mvp/Explainer.md
peterhuene: @Liamolucko do you think we can resolve this given Luke's explanation?

peterhuene (approved):

design/mvp/Explainer.md
Liamolucko: It should probably be written down first (so, replace `<valtype>*` with `<valtype>?`).

design/mvp/Explainer.md
lukewagner: Thanks!

design/mvp/Explainer.md
lukewagner: Ok, [updated](https://github.com/WebAssembly/component-model/pull/69/commits/3f74cc6526a4d6fc9c92bae3853562ce7400ce5f).  It does nicely reduce size of the change.

lukewagner:
I just rebased this PR on top of #70 and made the analogous [changes](https://github.com/WebAssembly/component-model/pull/69/commits/6e35c8fa3adb8617ba070fbf94e6586f670f812b) to WIT.md (s/unit// s/expected/result/).  Three subjective choices that might be wrong and are thus worth calling out and discussing are:
* For `result`, instead of requiring `_` to be manually written for empty payloads (e.g., `result<_,_>`), I added a few syntactic cases: `result<T,E>`, `result<T>`, `result<_,E>` and `result`. IMO, this is much nicer-looking and I hope has an obvious-enough meaning.
* For functions, I made the param/result tuples symmetric, matching the AST changes, so that they can both be a single type or a parenthesized named tuple, e.g.: `func u32 -> u32`, `func (x:u32) -> u32`, `func u32 -> (y:u32)`, `func (x:u32) -> (y:u32)`.
* The previous bullet interacts poorly with the previous syntactic sugar of allowing `-> ()` to be elided, since it lets you have `x : func u32` which, without the parens around `u32`, may look confusing.  The two options seemed to be: (1) make params asymmetric and require parens around params always, (2) remove the optional-`-> ()` sugar, so that you'll have `x : func u32 -> ()`.  IMO, removing the sugar is more clear (this is a function; what's its return type?) and language-neutral (many languages always require an explicit return type) and most functions should be returning some sort of `result` anyways, so it's not like a case we need to syntactically optimize for, so I currently went with that, but I'd be interested to hear other opinions.

theduke:
I'm curios: what is the motivation for removing `unit` and switching back to "implicit" unit types via omission?

Is it related to subtyping, simplifying the type model or language mapping constraints?

I quite appreciated having an explicit unit type.

How would one represent a union with a potential `unit` value?

For example, in Assemblyscript: `i64 | f64 | null`, or Scala: `Unit | String`, ...

Without a dedicated `unit` type there is quite a bit of ambiguity there.

For example, one possible mapping is using `(option T)`  on the WIT side.
So `i64 | f64 |null` could be `(option (union i64 float64))`.
But sticking with the Scala example: Scala has an `Option[T]` type, so distinguishing between `Unit|String` and `Option[String]` is a matter of interpretation / preference.

There are probably other examples of where this ambiguity could potentially be problematic.

lukewagner:
`unit` was added when multi-return was originally removed (as the alternative way to express "no meaningful return value"), so since we're basically reverting that change here so that we don't end up with two ways to express the same idea (forcing API designers to ask which is better and make varying choices).  As you said, `unit` introduces per-language binding questions that are nice to avoid altogether ([e.g.](https://github.com/WebAssembly/component-model/issues/49)).  Also, for cases where one might want to use a `unit` to implicitly convey "a case without a payload", probably `union` is best avoided and a `variant` (with a case with an empty payload) will be more clear for everyone.

lukewagner:
I'm in the 40-60% range on this question, so no strong opinions here, but just to dig into the reasoning for the asymmetry a bit:

> In particular, result is almost exclusively used asymmetrically, as a result and not as a param, and will be used pervasively.

I agree that the vast majority of functions should be returning a `result` (or a `future<result>` or *something*), but I thought this implied the opposite conclusion: why have the extra syntactic sugar for leaving off the `-> ()` if noone is almost ever writing it?  It seems like you'd almost want to highlight it because maybe someone just *forgot* a `result`.

skreborn:
For what it's worth, I'm all for explicitness. I've come to appreciate having to type a little more to satisfy a compiler versus accidentally leaving off important details. I would much rather unambiguously specify an empty result set than have the compiler figure it out. In fact, I'd prefer if the compiler straight-up told me that I'm missing a return type instead of quietly rolling over it.

sunfishcode:
> I'm in the 40-60% range on this question, so no strong opinions here, but just to dig into the reasoning for the asymmetry a bit:
> 
> > In particular, result is almost exclusively used asymmetrically, as a result and not as a param, and will be used pervasively.
> 
> I agree that the vast majority of functions should be returning a `result` (or a `future<result>` or _something_), but I thought this implied the opposite conclusion: why have the extra syntactic sugar for leaving off the `-> ()` if noone is almost ever writing it? It seems like you'd almost want to highlight it because maybe someone just _forgot_ a `result`.

WASI itself is expecting to have functions with no return values, in the upcoming logging API. It's much less common, but does happen. That said, `-> ()` wouldn't be the worst, but it does seem less familiar (Typescript, Rust).

My main suggestion here is that we don't need the special cases for 0 or 1 parameters. Those would risk confusion and uninteresting variety, which is a more important concern than syntactic symmetry here. And consequently, we don't need whitespace between `func` and `(`.

lukewagner:
If we remove the "implicit `-> ()`" syntactic sugar, then there is no special-case for for 0 params/results, only for 1 param/result.  I also don't like special cases, and so it doesn't seem to always require parens; the thing that made me consider otherwise is that, independently of *parenthesization*, when it comes to *naming* we already have this design choice from #41 wherein you can leave off the name of a single param/result, making "one unnamed parameter" a special case *anyhow*.  Thus, it seemed kindof naturally aligned to say that `(` marks the beginning of 0..N `<name>: <valtype>` pairs whereas non-`(` means just `<valtype>` -- that's basically what the underlying component-model AST is saying anyway.

lukewagner:
> The circumstances that make a single unnamed return value a useful special case don't seem to apply to parameters in the same way, so my suggestion is to deemphasize this special case in the wit syntax for parameters.

I suppose it's a matter of perspective, but in the presence of multi-return, it does feel like the same circumstances apply to both: if we say that we *always* want to have parens around params, then it seems surprising to have a different rationale applied to results.  (It made sense for single-return, of course; but the context here is updating WIT to stay in lock-step with multi-return in the component model.)

sunfishcode (approved):
Looks good to me!

lukewagner:
Thanks for the review!

PR #77, opened 2022-07-29
Don't force spurious memory length check if list element size is 0

lukewagner:
@alexcrichton pointed out that the current trapping semantics require emitting a spurious length check when it's otherwise statically evident that there are no loads.


alexcrichton (approved):

alexcrichton:
I was [thinking a bit more about this](https://github.com/WebAssembly/component-model/issues/78) and had a few more questions which might be good to resolve before landing this.

lukewagner:
Canceling as discussed in #78.

PR #80, opened 2022-08-02
CLI Embedding Proposal

esoterra:
Add a Typed Main proposal to the component model.

[Rendered](https://github.com/Kylebrown9/component-model/blob/typed-main/design/mvp/CLIEmbedding.md)

### DIFF
--- a/design/mvp/CLIEmbedding.md
+++ b/design/mvp/CLIEmbedding.md
@@ -1,10 +1,10 @@
 # Typed Main
 
-Previous WASI Command APIs were not able to strongly type the resource and data inputs to commands and instead relied on functions for reading command line arguments and environment variables.
+A high-level goal of the component model is the ability to run a single component portably on various hosts. One such embedding is a classic Unix-style Command-Line Interface (CLI) accepting positional/named arguments.
 
-Typed Main enables components to define value imports for their arguments (including file resources) and environment variables allowing the host to parse and validate them for the component.
+This document describes how a component can be generically interpreted by a component-enabled CLI to allow components to be directly instantiated and invoked from the command line in an idiomatic manner, corresponding to use case #2 of [invoking component exports](https://github.com/WebAssembly/component-model/blob/main/design/high-level/UseCases.md#invoking-component-exports-from-the-host). It also defines annotations which can be used to extend the default interpretation of a component to specify CLI-specific features (e.g. short flag names).
 
-This proposal defines a way to infer the command line arguments names, ordering, etc. from the component type (represented here in [WIT](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md)) and optionally augmenting this with annotated info (represented here as [Structured Annotations](https://github.com/WebAssembly/component-model/issues/58)). The anotation information will be encoded in a custom section in a not-yet-specified way.
+The component type is specified here in terms of a [`*.wit`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md) interface and annotations as [Structured Annotations](https://github.com/WebAssembly/component-model/issues/58).
 
 ## Arguments
 * Value imports are can be supplied as positional arguments in the order they are defined
--- a/design/mvp/TypedMain.md
+++ b/design/mvp/TypedMain.md
@@ -1,0 +1,115 @@
+# Typed Main
+
+Previous WASI Command APIs were not able to strongly type the resource and data inputs to commands and instead relied on functions for reading command line arguments and environment variables.
+
+Typed Main enables components to define value imports for their arguments (including file resources) and environment variables allowing the host to parse and validate them for the component.
+
+This proposal defines a way to infer the command line arguments names, ordering, etc. from the component type (represented here in [WIT](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md)) and optionally augmenting this with annotated info (represented here as [Structured Annotations](https://github.com/WebAssembly/component-model/issues/58)). The anotation information will be encoded in a custom section in a not-yet-specified way.
+
+## Arguments
+* Value imports are can be supplied as positional arguments in the order they are defined
+* Value imports can be supplied as long-form named arguments with their defined name
+* The ordering of named arguments with respect to each other and positional arguments is not constrained
+* If a named argument is present, parsing of positional argument skips over it to the next position
+
+```rust
+input: string
+pattern: string
+replace: string
+```
+
+### Usage
+```
+$ replace "some-thing" "-" "_"
+$ replace --input "some-thing" --pattern "-" --replace "_"
+$ replace --pattern "-" --replace "_" "some-thing"
+```
+
+## File Preopen Arguments
+* If the type of a value import is a WASI file, the host reads the corresponding argument as a path and preopens it.
+* The preopen must occur even for file paths which do not exist (as long as their parent does), so that output file paths can be supported.
+
+```rust
+import { file } from ...
+
+input: file
+output: file
+```
+
+### Usage
+```bash
+$ copy ./input.txt --output ./output.txt
+```
+
+## Optional Arguments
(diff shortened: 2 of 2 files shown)


lukewagner (commented):
Thanks for writing this up!  Quick high-level comment: what if we titled this document "Command-Line Interface Embedding" and named the file `CLIEmbedding.md` and slightly reframed the intro to say something to the effect of:
* A high-level goal of the component model is to be able to run a single component portably on multiple hosts.
* One such host is a classic Unix-style Command-Line Interface (CLI).
* This document describes how a component can be generically interpreted by a component-enabled CLI to allow components to be directly instantiated and invoked from the command line in an idiomatic manner, corresponding to use case #2 in [this section](https://github.com/WebAssembly/component-model/blob/main/design/high-level/UseCases.md#invoking-component-exports-from-the-host)

My thinking here is that, at the component-model level, what we're defining here is symmetric to the [JS Embedding](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#JavaScript-embedding) (which will likely, as it expands to go into more detail, get factored out into a symmetric `JSEmbedding.md`), treating the CLI (well, sh/bash/...) as just another programming language client.

What's neat about this context is that it's not WASI-specific.  The only WASI interaction here is for how files work, but we could even this part of the embedding to say that it works for any resource type import that is known to the host.

Kylebrown9:
Thanks, that makes sense to me. I've rendered your suggested changes.

design/mvp/CLIEmbedding.md
```diff
@@ -0,0 +1,115 @@
+# Typed Main
```
lukewagner, line 1:
```suggestion
# Command-Line Interface Embedding
```

design/mvp/CLIEmbedding.md
```diff
@@ -0,0 +1,115 @@
...
+    * Name is inferred to be the value import name
+    * Name can be explicitly set `@flag(long = "foobar")`
+    * Does not support grouping
+* Flags can annotated as "short"
+    * Using annotation `@flag(short)`
+    * Specified in short form with no associated value e.g.`-f`
+    * Name is inferred to be first character of value import name
+    * Name can be explicitly set `@flag(short = "f")`
+    * Supports grouping `-a -b -c` = `-abc`
+* Flag can be short and long `@flag(short, long)`, value `true` if either are present
```
lukewagner, line 77: It seems to me like the long name should *always* be accepted (since it's explicitly part of the component interface as an import) and that the `@flag(short)` annotation only adds an *additional* syntactic sugar option.  One can imagine a scenario where I use some general component tooling (say, a component discovery tool which doesn't know about the CLI structured annotation) to look at the signature of the component and I see a value import named `foo` so I'm unpleasantly surprised when I try to pass `--foo` on the CLI and it's rejected.  (And then if we agree on that, I think there's no need for a `@flag(long)`.)

design/mvp/CLIEmbedding.md
```diff
@@ -0,0 +1,115 @@
...
+* Value imports are can be supplied as positional arguments in the order they are defined
+* Value imports can be supplied as long-form named arguments with their defined name
+* The ordering of named arguments with respect to each other and positional arguments is not constrained
+* If a named argument is present, parsing of positional argument skips over it to the next position
+
+```rust
+input: string
+pattern: string
+replace: string
+```
```
lukewagner, line 19: Once @fibonacci1729 posts a sketch `.world` file syntax (soon, I think), it'd be very nice to update the syntax in all these wit blocks, which would have the nice effect of being clear what's an `import` and `export`.

lukewagner (commented):
Thanks a bunch!  Looks generally good to me with these tweaks but I'd be interested to see what other feedback comes up from the upcoming [subgroup meeting](https://github.com/WebAssembly/meetings/blob/main/wasi/2022/WASI-08-11.md) before final approval/merge.

npmccallum:
@Kylebrown9 Thanks for presenting today. I think this is a good direction overall.

However, I'm really uncomfortable with imports being both positional and named arguments by default. This has several drawbacks:

1. It isn't how people expect CLIs to work in existing programs. I cannot think of a single application that has this behavior.
2. Based on peoples' experience, the behavior that specifying a named argument changes the positional arguments will appear somewhat surprising. Although this is deterministic, it will be unexpected.
3. Because shifting positional arguments in the presence of named arguments will produce unexpected behavior, it will also potentially produce insecure or destructive behavior.

I propose instead the following:

1. Imports may be either named or positional; not both.
2. Imports are automatically treated as named arguments.
3. Developers may use an annotation to mark an import as a positional argument.

This does something useful by default and produces no unexpected behavior.

design/mvp/CLIEmbedding.md
ricochet: I'm leaning towards this answer to always export long but it would be great to hear other opinions. I almost always want a long-named flag when I have a short and it's difficult to come up with reasons for why that would be a problem. On the other hand, the way we would have to back out of this implementation once we standardize it is awkward in case we really do find a use-case where providing a long-named flag is not desired. A contrived example would be for a drop-in replacement of an existing CLI that only wants to accept inputs they have tests for (aka short-names), then we would need to support something like `@flag(short, no-long)`.

lukewagner:
That concern makes sense to me!  So then agreed to saying that, by default, all imports are named and you can't pass them positionally from the CLI; you have to opt-in to allowing an import to be supplied positionally with a CLI annotation.

Mirroring my comment [above](https://github.com/WebAssembly/component-model/pull/80#discussion_r937240981) regarding short/long, one property that I think is desirable is: if you can invoke a component successfully from the CLI without the CLI annotations, adding the CLI annotations shouldn't break you; it should only add new available syntax.  This seems valuable because it allows you to (1) add CLI annotations to any existing component without worrying that you're breaking someone, and (2) invoke a component from the CLI after looking at the component's type in a type-viewer tool/UI that doesn't understand the CLI annotations (i.e., you should be able to ignore the CLI annotations).  Based on this desired property, I'd propose that, when you opt-in to positional syntax, that doesn't disallow you from using the named syntax.  My expectation here is that the CLI Embedding can define what constitutes a valid combination of named/positional CLI arguments in an unsurprising and safe manner similar to Python.  Does that make sense?

design/mvp/CLIEmbedding.md
lukewagner: Yeah, that's a good case to think about.  My expectation here is that, for a large number of reasons, this automatic CLI Embedding won't be able to emulate the variety of CLI syntaxes in use for tools today, so we shouldn't worry too much about these exact-emulation scenarios.  On the flip side, thinking about this issue afresh, it seems like the "adding CLI annotations should always be backwards-compatible" principle mentioned [below](https://github.com/WebAssembly/component-model/pull/80#issuecomment-1216021411) is a useful one to preserve.

lukewagner:
Not quite: I'm suggesting that, when you opt-in to positional for a particular import (via an annotation), that doesn't disable the ability to name that same particular import.  The idea being that annotations should only allow additional syntax, not disable previously-working (before the annotation) syntax.

lukewagner:
Returning to this with the benefit of seeing how everything else has developed in the meantime: a different way to think about creating a "CLI Embedding" is to consider Bash to be just another programming language that binds to WIT, like JS, Python, Go, Rust, etc.  From this perspective, there doesn't need to be a special "CLI Embedding" in the Component Model spec but, rather, Bash toolchain support which can be built separately as with all other languages.  A neat consequence of this switch in perspective is that, just like there is both `jco transpile` and `jco componentize` (i.e., components-in-a-native-JS-runtime and JS-in-components) which implement the same bindings in two different directions, you could also imagine componentizing Bash (which actually sounds pretty cool).

Based on that understanding, maybe we don't need to develop this idea here, but in some other toolchain repo (which we could link to from here, if anyone starts working on it).

esoterra:
That makes sense to me. I'll shelve this PR and we

PR #91, opened 2022-08-15
Fix example which used one-level imports in a core module type

pl-semiotics:
This fixes a small typo in one of the examples in the explainer.


lukewagner:
Thanks!

Issue #90, opened 2022-08-15
Consistency in canonical definition indices


pl-semiotics:
Prior to #42, the Canonical ABI syntax used `core-prefix(core:<core:sort>idx)` for several `core:sort`s as part of the canon option syntax.   This was slightly strange, since the `core-prefix` meta-rule is only defined for rules which parse conses, whereas the various `core:...idx` rules in question parsed atoms, and additionally made things quite verbose, so #42 replaced them with `core:memidx` and `core:funcidx`, which makes a great deal of sense.

However, the explainer treats this as allowing either an actual identifier/number (as from `core:memidx`/`core:funcidx`) or an alias of syntax `$id "name"+`. This does not match how the inline export alias syntax is defined elsewhere in the explainer, which states: 
> For export aliases, the inline sugar has the form `(<sort> <instanceidx> <name>+)` and can be used in place of a `sortidx` or any sort-specific index (such as a `typeidx` or `funcidx`).

This gives rise to syntax such as
```
sortidx     ::= (<sort> <u32>)
              | (<sort> <instanceidx> <name>+)
core:memidx ::= <u32>
              | (<sort> <instanceidx> <name>+)  # Where <sort> must in fact be "core memory"
```

By that rule, the current grammar accepts `canonopt`s such as `(memory 0)` (meaning the first entry in the core memory index space of the current component), `(memory $M)` (the same, or an outer alias, via scoped identifier resolution), and `(memory (core memory $coremod "name"))`, which uses an inline export alias. (Whether this inline export alias would consult the core or component index space, or have syntax for both, is also an open question; see #89.)

I think this is a reasonable option, but it doesn't match the syntax in most of the explainer examples, which should perhaps be updated.  Another option would be to modify the way that we write export alias sugar everywhere to specify that when an export alias replaces a sort-specific index it (a) does not name the sort and (b) does not appear inside parentheses (in order to reach the precise syntax currently in the explainer); i.e. to specify something more like
> For export aliases, the inline sugar has the form `<instanceidx> <name>+` and can be used in place of any index into an index space.

Giving rise to syntax such as
```
sortidx     ::= (<sort> <u32>)
              | (<sort> <instanceidx> <name>+)
core:memidx ::= <u32>
              | <instanceidx> <name>+
```
However, this may cause issues with grammatical ambiguity.

lukewagner:
[...]

PR #94, opened 2022-08-16
Fix inline aliases used in 'memory' examples

lukewagner:
This PR fixes the examples to use the correct inline alias sugar inside the `memory` `canonopt` as highlighted by #90.


peterhuene (approved):

lukewagner:
I also noticed a bunch of missing `core` prefixes in `examples/SharedEverythingDynamicLinking.md`.

peterhuene (approved):

Issue #89, opened 2022-08-15
Revisiting core vs component aliases


pl-semiotics:
Currently, component definitions include both core-prefixed core aliases and component-level aliases. However, since the component-level sorts include core sorts (`sort ::= ... | core <core:sort>`), this leads to significant overlap.

In particular, most noticeably, core outer aliases such as `(core alias outer <u32> <u32> (<core:sort> <id>?))` are entirely equivalent to component outer aliases written `(alias outer <u32> <u32> (core <core:sort> <id>?))`.

More subtly, significant issues arise with regards to the alternative syntaxes for export aliases. Core export aliases are subtly different from component export aliases at the moment: a core export alias retrieves an export from a _core_ instance, while a component export alias retrieves an export from a _component_ instance. This similar syntax, however, gives rise to questions, especially about the inverted definition form: is `(core <core:sort> <id>? (alias export <idx> <name>))` equivalent to `(alias export <idx> <name> (core <core:sort> <id>?))` or to `(core alias export <idx> <name> (<core:sort> <id>?))`?  Furthermore, does the inline export alias syntax `(<sort> <idx> <name>)` desugar to a core or component export alias?

In the past, we resolved these ambiguities by removing the entire idea of a core alias from the spec, and noting that core export aliases were not useful at the time.  The former has been made untenable by the necessity, which has since become clear, of including core export aliases  in `core:moduledecl`; the latter is likely to change in the near future and is not a good assumption to make, so core aliases were re-added in their original form, giving rise to these questions.

I would suggest that we move `core:alias` to the discussion on core module types, and make it purely a part of `core:moduledecl`, and then enhance component aliases to include a new aliastarget making it possible to index into the core instance index space rather than the component instance index space.  This would nicely (a) remove the "two different syntax trees/operations that are identical" smell and (b) solve the issue with inverted out-of-line  aliases nicely.  The inline export alias syntax is another matter, and needs to either be fixed to always refer to one of component or core modules, or replaced with something that has an extra position allowing the distinction to be made (e.g. splitting it into `(<sort> <idx> <name>)` and `(<sort> core <idx> <name>)`.

lukewagner:
Great observations and suggestions; I'll give it a try.  As for what to do about inline alias sugar, I expect the vast majority of aliases will be satisfied if, in `(core <core:sort> $inst <name>)`, `$inst` indexes the core instance index space, so I think it's fine to just do that.

PR #95, opened 2022-08-17
remove component-level core aliases

lukewagner:
I think this addresses the comments in #89.  I believe this is a fairly minor change, tweaking the syntax and binary encoding, but not changing what's expressible.


peterhuene (approved):
The changes make sense to me.

pl-semiotics (approved):

Issue #25, opened 2022-04-21
Interface types don't always map to valid Web IDL types


Liamolucko:
Some of the conversions between JS values and interface types are defined in terms of the conversions between JS values and Web IDL types. However, not all of those interface types are valid Web IDL types.

For example:
- `option<option<T>>` - dissallowed, since JS `null` could map to either `none` or `some(none)`.
- `union { option<A>, option<B> }` - disallowed, since JS `null` could map to either a `none` of `option<A>` or a `none` of `option<B>`.
- `union { T, T }` - disallowed, since it's impossible to tell whether a value should be `T` variant 1 or `T` variant 2.

A more complex one:
```wit
record foo {
    a: u32,
}

record bar {
    b: s32,
}

union { foo, bar }
```
Even though `foo` and `bar` can technically be distinguished, Web IDL blanket disallows unions of dictionaries, presumably for simplicity.

How should these cases be handled? Should they be disallowed, or fall back on some less ergonomic representation?

lukewagner:
That's a great observation and question.  I'll preface this up-front by saying this answer isn't fully fleshed out and may need to be revised, but what I was thinking is that these problematic `union` cases (which we can statically recognize via a spec-defined predicate on the interface type) would fall back to the generic `variant` JS representation.  The JS binding for `variant` is currently marked as TBD b/c I didn't have time to fully look at the state of the art here (e.g. in TypeScript, the bindings of other statically-typed compile-to-JS languages, and any relevant TC39 proposals in-flight), but my current working idea is that a `(variant (case "a" string) (case "b" u32))` would produce JS values that looked like `{ a: "hi" }` or `{ b: 42 }`.

For your examples:
* `(option (option T))` would correspond to JS values looking like `{ some: t }`, `{ some: null }` or `{ none: undefined }`
* `(union (option A) (option B))` would correspond to JS values looking like `{ '0': null }`, `{ '0': a }`, `{ '1': null }`, `{ '1': b }` (since unions are specializations of variants with index-named cases)
* `(union T T)` would not be a valid interface type (validation/subtyping rules need to be better written up here)
* `(union (record (field "a" u32)) (record (field "b" s32)))` would correspond to JS values looking like `{ '0': { a: 42 } }` and `{ '1': { b: -1 } }`.

rossberg:
> * `(union T T)` would not be a valid interface type (validation/subtyping rules need to be better written up here)

[...]

PR #96, opened 2022-08-17
Add JS API notes about variants

lukewagner:
This PR adds notes to the JS API section speaking to the corner cases uncovered in #25.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -967,7 +967,7 @@ At a high level, the additional coercions would be:
 | `float32`, `float64` | as a Number, mapping the canonical NaN to [JS NaN] | `ToNumber` mapping [JS NaN] to the canonical NaN |
 | `char` | same as [`USVString`] | same as [`USVString`], throw if the USV length is not 1 |
 | `record` | TBD: maybe a [JS Record]? | same as [`dictionary`] |
-| `variant` | TBD | TBD |
+| `variant` | see below | see below |
 | `list` | create a typed array copy for number types; otherwise produce a JS array (like [`sequence`]) | same as [`sequence`] |
 | `string` | same as [`USVString`]  | same as [`USVString`] |
 | `tuple` | TBD: maybe a [JS Tuple]? | TBD |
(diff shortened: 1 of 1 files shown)


design/mvp/Explainer.md
```diff
@@ -985,6 +985,14 @@ Notes:
   the return value is specified by `ToJSValue` above. Otherwise, the function
   result is wrapped into a JS object whose field names are taken from the result
   names and whose field values are specified by `ToJSValue` above.
+* In lieu of an existing standard JS representation for `variant`, the JS API
+  would need to define its own custom binding built from objects. As a sketch,
+  the JS values accepted by `(variant (case "a" u32) (case "b" string))` could
+  include `{ a: 42 }` and `{ b: "hi" }`.
```
Liamolucko, line 991: Personally, I prefer the `{ tag, value }` encoding of variants like `wit-bindgen` currently uses, since it's easier to `switch` on the tag.

This is just a sketch though, so I don't think we have to decide on this right now.

design/mvp/Explainer.md
```diff
@@ -985,6 +985,14 @@ Notes:
...
+* For `union` and `option`, when Web IDL doesn't support particular type
+  combinations (e.g., `(option (option u32))`), the JS API would fall back to
+  the JS API of the unspecialized `variant` (e.g.,
+  `(variant (case "some" (variant (case "some" u32) (case "none"))) (case "none"))`).
```
programmerjake, line 995: imho it would make sense to unwrap variants until you hit something that does have a special case, so (assuming `(option u32)` has a special case where it would be encoded as a number or `null`, i didn't check) something like:
```
(variant (case "some" (option u32)) (case "none")) 
```

design/mvp/Explainer.md
lukewagner: Ohhh, I totally forgot about that.  So then would some concrete values be `{ tag:'a', value:42 }` and `{ tag:'b', value:'hi' }`?  Looking just now, I guess that corresponds more-closely to the [discriminated union support in TypeScript](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions).  So yeah, I'll switch to this.

lukewagner: [Fixed](https://github.com/WebAssembly/component-model/pull/96/commits/7339630b272ebb74c0ce828eb05d832dd042fcf5)

design/mvp/Explainer.md
lukewagner: Ah, good point, [fixed](https://github.com/WebAssembly/component-model/pull/96/commits/05bc3d6151173cd952657c45d559d10340b71028)

Liamolucko (approved):

PR #98, opened 2022-08-29
Remove the `async` keyword from WIT.md.

sunfishcode:
In the current [async proposal], there are no longer `async` functions; there
are instead functions that return `future` or `stream`.

[async proposal]: https://docs.google.com/presentation/d/1MNVOZ8hdofO3tI0szg_i-Yoy0N2QPU2C--LzVuoGSlE/edit#slide=id.g1270ef7d5b6_0_111

### DIFF
--- a/design/high-level/UseCases.md
+++ b/design/high-level/UseCases.md
@@ -180,7 +180,7 @@ use cases that require additional features:
    can encapsulate `i32` pointers to linear memory allocations that need to be
    safely freed when the last handle goes away.
 3. Developers import or export functions with signatures containing
-   concurrency-oriented types (e.g., async, future and stream) to address
+   concurrency-oriented types (e.g., future and stream) to address
    concurrency use cases like non-blocking I/O, early return and streaming. Both
    developers (the caller and callee) are able to use their respective languages'
    native concurrency support, if it exists, using the concurrency-oriented types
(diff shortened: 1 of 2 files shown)


lukewagner (approved):
LGTM, thanks!  (I'll wait a couple more days for comments, then merge.)

Kylebrown9:
Is there rationale for this change somewhere?

lukewagner:
Earlier versions of the async proposal ([e.g.](https://docs.google.com/presentation/d/1WtnO_WlaoZu1wp4gI93yc7T_fWTuq3RZp8XUHlrQHl4)) had `async` as an attribute on `func` whereas the [current proposal](https://docs.google.com/presentation/d/1MNVOZ8hdofO3tI0szg_i-Yoy0N2QPU2C--LzVuoGSlE) removed it in favor of more-general `future` and `stream` type constructors that can be used anywhere in `func` params and results, so the rationale is just keeping up-to-date with the async proposal.

Kylebrown9:
What was the motivation in the proposal for making that change?

lukewagner:
Oops, sorry, I missed the follow-up question.  The motivation in the original proposal that introduced `async` was that `async` represented an [effect type](https://en.wikipedia.org/wiki/Effect_system) saying "this function call can block" (allowing the caller to say "if the callee blocks, return the suspended computation and let me keep running".  `future` can say roughly the same thing, but is more expressive (it can be used anywhere in a param/result types) and is complemented by `stream`, so there is no need for the more-limited `async` effect type anymore.

PR #99, opened 2022-08-31
Remove a space from a canonical ABI mangling example.

sunfishcode:
Change `func (v1: string)` to `func(v1: string)` in an example to match the
mangling of the canonical ABI.


lukewagner:
Good catch, thanks!

PR #101, opened 2022-09-03
Add draft of the specification of validation

pl-semiotics:
This PR adds a spec/document folder with a draft of the beginnings of a spec document, currently providing a formalisation of validation rules for the component model.


lukewagner:
Woohoo, excited to see this; great work!  Just to lay out my plan, I'll be incrementally reviewing the PR in detail over the coming weeks, submitting comments piecemeal.

pl-semiotics:
@lukewagner Updated with the current structure of resource types.

I also removed the checkin of docs/, since it doesn't seem to have worked to provide an easy rendered version & it causes a lot of noise in git.  See the [rendered](https://ccs.neu.edu/home/semiotics/_component-model-draft/multihtml/) version for now; if this is merged, I'll set up CI hooks + GH Pages to get a CI rendered version.

pl-semiotics:
@lukewagner As discussed, this version has been de-Sphinx'd.  It's mostly in line with the reference implementation now. The changes from the previous version are mostly minor (but important) fixes; the largest presentational change is also the exception to that (i.e. is not yet done in the reference implementation), but should not actually change almost anything: existential quantifiers have been moved from `instancetype_e` to `componenttype_e` and `deftype_e`, since they are (and must be) always hoisted to one of them anyway.

lukewagner:
Sorry for the slow reply here, but excited to see this update!  Could you perhaps also include a rendered .pdf in the `document/` directory?  Maybe we remove it before things are final, but at least in the interim, it would be convenient.

PR #102, opened 2022-09-03
Allow using exported types in component and instance types

pl-semiotics:
This changes the syntax of component and instance types so that it is possible to name a type with a typebound when it is declared, rather than when it is exported, in order to create an entry for the abstract type so created in the type index space which may be used in the types of future exported definitions.  For uniformity's sake, the same treatment is given to all definition sorts.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -448,7 +448,7 @@ module types always start with an empty type index space.
 )
 ```
 
-Component-level type definitions are symmetric to core-level type definitions,
+Component-level type definitions are similar to core-level type definitions,
 but use a completely different set of value types. Unlike [`core:valtype`]
 which is low-level and assumes a shared linear memory for communicating
 compound values, component-level value types assume no shared memory and must
@@ -487,7 +487,7 @@ instancedecl  ::= core-prefix(<core:type>)
                 | <alias>
                 | <exportdecl>
 importdecl    ::= (import <externname> bind-id(<externdesc>))
-exportdecl    ::= (export <externname> <externdesc>)
+exportdecl    ::= (export <externname> bind-id(<externdesc>))
 externdesc    ::= (<sort> (type <u32>) )
                 | core-prefix(<core:moduletype>)
                 | <functype>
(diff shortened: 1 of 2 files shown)


design/mvp/Explainer.md
```diff
@@ -464,19 +464,25 @@ instancetype  ::= (instance <instancedecl>*)
 componentdecl ::= <importdecl>
                 | <instancedecl>
 instancedecl  ::= core-prefix(<core:type>)
-                | <type>
+                | (type <typebound>)
```
lukewagner, line 467:
```suggestion
                | <typedecl>
```
?  Otherwise, I don't see where `typedecl` is used

design/mvp/Explainer.md
```diff
@@ -464,19 +464,25 @@ instancetype  ::= (instance <instancedecl>*)
...
+                | (core module <typeidx>)
+                | (func <typeidx>)
+                | (value <typeidx>)
+                | (instance <typeidx>)
+                | (component <typeidx>)
```
lukewagner, line 472: If we add the syntactic sugar suggested below, I think perhaps we can hold off introducing these syntactic forms until they might actually be needed in the future.

design/mvp/Explainer.md
```diff
@@ -464,19 +464,25 @@ instancetype  ::= (instance <instancedecl>*)
...
                 | <alias>
                 | <exportdecl>
-importdecl    ::= (import <name> bind-id(<externdesc>))
-exportdecl    ::= (export <name> <externdesc>)
-externdesc    ::= (<sort> (type <u32>) )
+importdecl    ::= (import <name> bind-id(<importdesc>))
+typedecl      ::= (type <id>? <typebound>)
+exportdecl    ::= (export <name> <sortidx>)
```
lukewagner, line 477: I think it would be useful to keep allowing the same syntax as now by having:
```
exportdecl ::= (export <name> <exportdesc>)
exportdesc ::= bind-id(<externdesc>)
             | <sortidx>
```

design/mvp/Explainer.md
```diff
@@ -464,19 +464,25 @@ instancetype  ::= (instance <instancedecl>*)
...
+importdesc    ::= (<sort> (type <u32>) )
                 | core-prefix(<core:moduletype>)
                 | <functype>
                 | <componenttype>
                 | <instancetype>
                 | (value <valtype>)
                 | (type <typebound>)
-typebound     ::= (eq <typeidx>)
+typebound     ::= (eq <defype>)
```
lukewagner, line 485: Should `typebound` continue to also allow `typeidx`?

design/mvp/Explainer.md
```diff
@@ -464,19 +464,25 @@ instancetype  ::= (instance <instancedecl>*)
...
+                | (value <typeidx>)
+                | (instance <typeidx>)
+                | (component <typeidx>)
                 | <alias>
                 | <exportdecl>
-importdecl    ::= (import <name> bind-id(<externdesc>))
-exportdecl    ::= (export <name> <externdesc>)
-externdesc    ::= (<sort> (type <u32>) )
+importdecl    ::= (import <name> bind-id(<importdesc>))
+typedecl      ::= (type <id>? <typebound>)
```
lukewagner, line 476: Interesting; so to see if I'm reading correctly: do *all* type definitions inside a component-/instance-type happen inside a type bound?

lukewagner (commented):
Thanks for identifying this future hazard!  I need to ask a few clarifying questions first to see if I'm understanding the proposed change.  Could you also update Binary.md to match?

lukewagner:
Thanks for thinking through this!  Trying to keep the commit history minimal and targeted, I think it might be better to force-push a new commit that tries to make the minimal delta to Explainer.md.  As is, the diff contains both some substantial changes but also just some churn, so it's harder to see what the meaningful change is.  Also, it'd be nice to include a short example that illustrates clearly how exports introduce a new index and how subsequent definitions can reference both type definitions and exported type definitions, and how the latter implies a name.  Also, now that Wit is a thing in the component-model, it'd be useful to explain how these would be rendered differently (inline types vs. out-of-line type definitions).  Lastly, it'd be good to update the validation summary in Binary.md for exports.  Or, since that's a lot of asks, I'm happy to write the PR myself; either way.

design/mvp/Explainer.md
```diff
@@ -567,41 +567,81 @@ also be defined as standalone, human-friendly text files in the [`wit`](WIT.md)
...
+transparent to consumers of the component type, and any later
+references to theme could (semantically, if not syntactically) be
+replaced by the definition provided, as in fact they are during
+validation elaboration.
+
+Export declarators take a name and an extern descriptor describing the
+entry that they are exporting. However, they also create an entry in a
+local (to the current module or instance type declaration) index space
+for the appropriate sort, which may be used by later declarators to
+refer to the concept of this export. In the future, this will be used
```
lukewagner, line 601:
```suggestion
for the appropriate sort, which may be used by later declarators to
refer to the abstract named thing being exported. In the future, this will be used
```

design/mvp/Explainer.md
```diff
@@ -567,41 +567,81 @@ also be defined as standalone, human-friendly text files in the [`wit`](WIT.md)
...
+```wit
+type nanos = u32
+foo: func(in: u32) -> nanos
+```
+
+Component import and alias declarators are analagous to the core module
+declarators introduced above, and have essentially the same meaning, although
+they also (like the import and alias _definitions_ present inside a component)
+create entries in the local index spaces that may later be exported. As with
+core modules, `importdecl` classifies a component's `import` definitions.
```
lukewagner, line 641:
```suggestion
Component alias declarators are analagous to the core module
declarators introduced above, and have essentially the same meaning.
```
I think the parts I chopped out are already said above?

design/mvp/Explainer.md
```diff
@@ -567,41 +567,81 @@ also be defined as standalone, human-friendly text files in the [`wit`](WIT.md)
...
-be extended with a `sub` option (symmetric to the [type-imports] proposal) that
-allows importing and exporting *abstract* types.
+Note that both `(type ...)` and `({im,ex}port ... (type ...))`
+declarators introduce entries to the type index space. The difference
+is that the former are treated completely as aliases, with no identity
+of their own, while the latter are intended to associate future
+declarators with the exported type, so that the latter is correctly
+used in WIT/bindings/etc. When resource types are added, only the
+latter form will allow adding abstract resource types to the index
+space.
```
lukewagner, line 619:
```suggestion
declarators introduce entries to the type index space. Before resource
types, the only difference is that the latter are associated with a name.
(With resource types, only type imports/exports will be able to introduce
introduce resource types into the type index space since resource types.)
```
I think "alias" might be a confusing distinction to make here since, from another point of view, type exports are more like aliases (they add another index for the same thing, rather than defining a totally new thing).  This suggested alternative wording tries to avoid the confusion.

lukewagner:
#129 essentially slurped up this PR so, as discussed, I think we can close this and iterate in future issues/PRs.

PR #104, opened 2022-09-12
Special-case empty `{ ... }` sequences for prettier name mangling.

sunfishcode:
With the current mangling for eg. records:

```python
  return 'record { ' + ', '.join(mangled_fields) + ' }'
```

the mangling for a record with 0 fields is `record {  }`, with two spaces between the braces. That looks subjectively surprising.

This PR adds special cases to the mangling for `record` and similar types so that the mangling is `record {}`, with no spaces between the braces, which is subjectively prettier.

The argument against this is that empty records should be very rare, so prettiness isn't important, and every special case makes the system more complex. The argument for it is that a mangling with two spaces looks weird.

### DIFF
--- a/design/mvp/canonical-abi/definitions.py
+++ b/design/mvp/canonical-abi/definitions.py
@@ -1145,26 +1145,41 @@ def mangle_valtype(t):
 
 def mangle_recordtype(fields):
   mangled_fields = (f.label + ': ' + mangle_valtype(f.t) for f in fields)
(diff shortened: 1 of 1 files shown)


lukewagner:
This is ultimately a subjective matter, but to me this seems to add more special cases than warranted by the frequency and degree of benefit to humans. ¯\_(ツ)_/¯

sunfishcode:
¯_(ツ)_/¯

Issue #107, opened 2022-09-15
Clarifying anonymous parameters, import/export symmetry


kulakowski-wasm:
[Recently the component model was updated back to multi-return](https://github.com/WebAssembly/component-model/pull/69), along with a collection of related changes to WIT. Now, both function parameters and results are either a) a single, unnamed type or b) a list of all named types. This symmetry is intended to have WIT reflect import/export symmetry in the basic component model AST.

In practice, functions on resources (methods) like the following complicate this.

```
resource r {
  f: func u32 -> result
}
```

because there are in fact now 2 unnamed parameters, the `self` receiver and the anonymous u32 argument which language bindings need to handle. There's no case of an additional unnamed parameter type.

Luke wondered, in my conversation about this, if this is the missing argument for why we shouldn't have made imports and exports symmetric and instead say that imports always have param names.

lukewagner:
Thanks for noticing and filing!  If the alternative is to have different syntactic rules for functions based on whether they're methods or not (in which methods had to have all-named parameters but functions were allowed to have a single-unnamed-parameter), I think it'd be better to simply require all parameters to have names (in both Wit and component types).

kulakowski-wasm:
I wanted to double check if "parameters" means really just function parameters, or both parameters and results. I'm clarifying because the unification of the representation of parameters and results makes it tempting to just say "parameters" to mean both, especially in code.

Currently WIT.md says

```
func-item ::= id ':' 'func' func-vec '->' func-vec

func-vec ::= ty
           | '(' func-named-type-list ')'

func-named-type-list ::= nil
                       | func-named-type ( ',' func-named-type )*

func-named-type ::= id ':' ty
```

and so I'm trying to tell which is your suggestion:

```
func-item ::= id ':' 'func' func-named-type-list  '->' func-vec
func-item ::= id ':' 'func' func-named-type-list '->' func-named-type-list
```

(plus or minus some parens or whatever)

lukewagner:
Yes, I did mean "parameters" as distinct from results and so what you wrote is what I meant too (modulo parens).

skreborn:
I would think that anonymous parameters are really not worth the extra complication that they come with, and would much rather all parameters were named.

PR #108, opened 2022-09-16
Remove unnamed function parameter case

lukewagner:
As discussed in #107, this PR removes the unnamed function parameter case in the AST, text, binary and Wit formats.


peterhuene (approved):
👍 with regards to the binary / text / wit format changes.

lukewagner:
Thanks!

PR #112, opened 2022-10-03
Remove `handle` as a keyword

kulakowski-wasm:
Owned/unique handles to resources are now just directly named, and this keyword is unused.


kulakowski-wasm:
@lukewagner since I can't seem to add a reviewer in the ui

lukewagner:
lgtm, thanks!  And to provide a bit more context: handles will be re-added shortly, but the literal token won't be `handle` (and already isn't in wit-bindgen).

kulakowski-wasm:
Yeah, https://github.com/bytecodealliance/wit-bindgen/pull/343/files changes wit-bindgen to no longer accept `handle r`.

Issue #115, opened 2022-10-07
WIT Syntax: `interface`


fibonacci1729:
# WIT + `interface`

Given the proposal to introduce a new top-level [WIT](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md) syntax (i.e. [`worlds`](https://github.com/WebAssembly/component-model/pull/83)) for describing [component types](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#type-definitions), it has become clear that `*.wit` documents are evolving to be able to express more than just a singular interface per file.

Additionally, while conceptually defining an interface per file may be desired (or even idiomatic) for many repositories it is inconvenient for use-cases where inlining N interfaces into a single `*.wit` file is preferable; For example tests become unwieldy and verbose if strewn across many files. 

The `interface` syntax introduced below will enable such use-cases effectively loosening the currently imposed conventional organization of `*.wit` files.

See [motivating comments](https://github.com/WebAssembly/component-model/pull/83#issuecomment-1246882768).

---

From a component-model POV, an `interface` is `WIT` syntax for describing an [instance type](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#type-definitions). As an example consider how the following interfaces might be modeled with `WIT` today:

```
// http-handler.wit

use { request, response, error } from http-types

handle-http-request: func(req: request) -> result<response, error>
```

```
// http-types.wit

record request {
    // ...omitted
}

record response {
    // ...omitted
}

variant error {
    // ...omitted 
}
```

Enabling the `interface` keyword will allow the refactor of these `*.wit` documents into a single file, e.g.:

```
// http.wit

interface http-handler {
    use { request, response, error } from http-types

    handle-http-request: func(req: request) -> result<response, error>
}

interface http-types {
    record request {
        // ...omitted
    }

    record response {
        // ...omitted
    }

    variant error {
        // ...omitted 
    }
}
```

Concretely, the structure of an `interface` is:
```
interface-item ::= `interface` id  strlit? `{` interface-items* `}`
```

> NOTE: `wit-item` refers to the items that can be declared using [WIT](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#top-level-items) today.

lukewagner:
This change makes a lot of sense to me, thanks for writing it up!  Two follow-up comments:

[...]

PR #116, opened 2022-10-10
WIT Syntax: `interface`

fibonacci1729:
Signed-off-by: Brian H <brian.hardock@fermyon.com>

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -32,6 +32,7 @@ token ::= whitespace
         | operator
         | keyword
         | identifier
+        | strlit
 ```
 
 Whitespace and comments are ignored when parsing structures defined elsewhere
@@ -118,6 +119,50 @@ keyword ::= 'use'
 A `wit` document is a sequence of items specified at the top level. These items
 come one after another and it's recommended to separate them with newlines for
 readability but this isn't required.
+
+Concretely, the structure of a `wit` document is:
+```
+wit-document ::= interface-item*
+```
+
+## Item: `interface`
+
+Interfaces can be defined in a `wit` document. Interfaces have a name and a sequence of items and functions.
+
+```wit
+interface example {
+    thunk: func() -> ()
+    fibonacci: func(n: u32) -> u32
+}
+```
+
+Specifically interfaces have the structure:
+
+```wit
+interface-item ::= 'interface' id strlit? '{' interface-items* '}'
+
+interface-items ::= resource-item
+                  | variant-items
+                  | record-item
+                  | union-items
+                  | flags-items
+                  | enum-items
+                  | type-item
+                  | use-item
+                  | func-item
+
+func-item ::= id ':' 'func' param-list '->' result-list
+
+param-list ::= '(' named-type-list ')'
+
+result-list ::= ty
+              | '(' named-type-list ')
+
+named-type-list ::= nil
+                  | named-type ( ',' named-type )*
+
+named-type ::= id ':' ty
+```
 
 ## Item: `use`
 
@@ -329,32 +374,6 @@ union-items ::= 'union' id '{' union-cases '}'
 
 union-cases ::= ty,
               | ty ',' union-cases?
-```
-
-## Item: `func`
-
-Functions can also be defined in a `wit` document. Functions have a name,
-parameters, and results.
-
-```wit
-thunk: func() -> ()
-fibonacci: func(n: u32) -> u32
-```
-
-Specifically functions have the structure:
-
-```wit
-func-item ::= id ':' 'func' param-list '->' result-list
-
-param-list ::= '(' named-type-list ')'
-
-result-list ::= ty
-              | '(' named-type-list ')
-
-named-type-list ::= nil
-                  | named-type ( ',' named-type )*
-
-named-type ::= id ':' ty
 ```
 
 ## Item: `resource`


fibonacci1729:
closes #115

design/mvp/WIT.md
```diff
@@ -119,6 +120,50 @@ A `wit` document is a sequence of items specified at the top level. These items
 come one after another and it's recommended to separate them with newlines for
 readability but this isn't required.
 
+Concretely, the structure of a `wit` document is:
+```
+wit-document ::= wit-item*
+
+wit-item ::= interface-item
```
Mossaka, line 127: Corrent me if I am wrong, I remember we discussed to explictly use `interface <name> { }` as the top level item in last week's [component model sync](https://github.com/bytecodealliance/meetings/tree/main/component-model).  So I think the wit-item should only contain `interface-item` as this PR is concerned, and I believe the PR #83 will add another wit-item called `world-item`.  

When both PRs are merged in, I believe the syntax will end up looking like this:

```
wit-document ::= wit-item*
wit-item ::= interface-item | world-item
interface-item ::= func-item
           | resource-item
           | variant-items
           | record-item
           | union-items
           | flags-items
           | enum-items
           | type-item
           | use-item
```

design/mvp/WIT.md
```diff
@@ -119,6 +120,50 @@ A `wit` document is a sequence of items specified at the top level. These items
...
+interface example {
+    thunk: func() -> ()
+    fibonacci: func(n: u32) -> u32
+}
+```
+
+Specifically interfaces have the structure:
+
+```wit
+interface-item ::= 'interface' id strlit? '{' (wit-item | interface-func)* '}'
```
alexcrichton, line 152: Is it intentional that this allows for nested `interface`s? I don't see any particular reason why that wouldn't work but wanted to make sure that was deliberate.

alexcrichton (commented):
One thing I was reminded of when reading over this again, it might be good to start documenting a bit of name resolution semantics. For example should each `interface` start out with an empty namespace or should it "inherit" all the names from the enclosing file. 

For example in Rust each `mod` starts a fresh new namespace irrespective of the names within the module that defined the `mod`, and this is useful for refactoring code and moving it around since you can copy/paste without having to add more imports. Otherwise though it's less convenient if you're about to use a lot of names from the parent namespace.

design/mvp/WIT.md
fibonacci1729: @alexcrichton Deliberate choice! However, if we'd prefer, I see no reason why we couldn't start flat and introduce nesting later. Which do you prefer?

alexcrichton: In my [other comment](https://github.com/WebAssembly/component-model/pull/83#issuecomment-1274857126) I've seen how `wit-component` supports both "default" and "export an interface" style exports, and I think it would be best if that distinction could be removed. Using nested interfaces seems like it may be plausible for doing that, but otherwise I don't have much of a preference myself.

Overall I think it's best to solve the use cases we have today but try not to be too ambitious and solve too much from the outset. Along those lines I personally think a good line to draw is imports/exports of any combination of instances/functions at any level, and that naively seems like it should map well to nested interfaces (sort of at least). I think there's details to work out but I think nested interfaces are a promising line to solve this.

design/mvp/WIT.md
lukewagner: This sounds right to me too (modulo the fact that the `interface-item`s and `world-item`s would need to be wrapped in `interface { ... }` and `world { ... }` blocks, resp.  In the past we've also discussed allowing `use-item` at top-level, but it doesn't seem strictly necessary and it also raises some additional questions for name resolution.  Other than `use-item`, all the other items should I think produce named interface fields and thus belong inside an `interface`.  (I can imagine possible reasons why we might want to generalize this in the future, but it seems reasonable to just start with the two top-level items and see if we're happy with that.)

design/mvp/WIT.md
lukewagner: After #109, I think the way I'd frame "default"-style exports is: a component's "default" exports are those without a URL field in the name (since there is no external interface to refer to that is being implemented; it's just the component's own thing).  The way I think this look in `world` syntax is, e.g.:
```
world my-component {
  import upstreams: "wasi:http/upstreams"
  export http: "wasi:http/handler"  // (export "http" "wasi:http/handler" (instance ...))
  export foo: func(s: string) -> string // (export "foo" (func ...))
}
```

lukewagner (commented):
Good start!  Agreed with and added to the two previous comments.

design/mvp/WIT.md
fibonacci1729: @lukewagner @Mossaka I updated the proposed syntax to encapsulate the previous top-level items (i.e. `wit-item`) into  `interface-items` of `interface-item`. Let me know if this captures the above discussion.

Mossaka: lgtm!

design/mvp/WIT.md
```diff
@@ -119,6 +120,50 @@ A `wit` document is a sequence of items specified at the top level. These items
...
+    fibonacci: func(n: u32) -> u32
+}
+```
+
+Specifically interfaces have the structure:
+
+```wit
+interface-item ::= 'interface' id strlit? '{' interface-items* '}'
+
+interface-items ::= resource-item
```
Mossaka, line 144: Could interfaces be nested?

Mossaka (approved):
lgtm!

alexcrichton (approved):
👍

design/mvp/WIT.md
fibonacci1729: @Mossaka _eventually_, but currently there isn't an immediate use-case for nesting and it isn't clear how nested interfaces map onto the component model just yet.

PR #123, opened 2022-10-31
Add extra Canonical ABI note and variant clarification

### LINKED ISSUE #119: define enums explicitly, particularly starting ordinal
There are some enums that start negative, for example debug logging level is -1 and the rest start after that. other enums start at zero (ex errnosuccess in old wasi), others at 1. It would be nice to both have enum handling documented precisely, and also allow adjustment of the start ordinal similar to iota in go.

lukewagner:
This PR is just editorial and resolves #119 by explicitly noting that case indices start at 0.  Doing this, I noted that the Type Definitions section should probably hint that there is an explicit mapping to concrete core wasm values/memory coming later.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -116,9 +116,10 @@ def alignment_record(fields):
 ```
 
 As an optimization, `variant` discriminants are represented by the smallest integer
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
thanks!

PR #124, opened 2022-11-08
Add a custom `name` section specification

### LINKED ISSUE #14: Spec-defined name section for components
Similar to how core wasm has a well-defined `name` custom section I think it would also be good to define one for the component model with all of the types of items in the component model itself. Additionally similar to how core wasm has an [extended name section proposal](https://github.com/WebAssembly/extended-name-section) to extend the name setion to all nameable items in core wasm I think it would be nice to name all items in the component model as well (not just a subset).

One of the trickier things perhaps about this would be how to name types-in-types. For example a module type introduces a new index space for types. Core wasm doesn't provide clear precedent here to map from although the component model can perhaps create its own covnention for defining the names here.

alexcrichton:
This is intended to assist with debugging/reading/writing the text format of components by avoiding the need to have everything be numbers and allowing tools to annotate names of items optionally (or have the names preserved from the text format).

Closes #14

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -340,7 +340,7 @@ namesec    ::= section_0(namedata)
 namedata   ::= n:<name>                (if n = 'component-name')
                name:<componentnamesubsec>?
                decls*:<declnamesubsec>*
-namesubsection_N(B) ::= N: byte size:<u32> B     (if size == |B|)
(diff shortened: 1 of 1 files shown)


alexcrichton:
Sure thing, done 👍

lukewagner:
Thanks, merging!

PR #126, opened 2022-11-16
Add a `wit-version` keyword to the WIT syntax.

sunfishcode:
Add a `wit-version` keyword to the WIT syntax, which introduces a WIT version declaration, which currently must contain the string "0xa". In the future when WIT is more stable, this may be used to facilitate backwards-incompatible language changes.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -87,7 +87,8 @@ bare as an identifier. These are used to help parse the format, and the list of
 keywords is still in flux at this time but the current set is:
 
 ```wit
-keyword ::= 'use'
+keyword ::= 'wit-version'
+          | 'use'
           | 'type'
           | 'resource'
           | 'func'
@@ -116,13 +117,18 @@ keyword ::= 'use'
 
 ## Top-level items
 
-A `wit` document is a sequence of items specified at the top level. These items
+A `wit` document starts with a `wit-version` followed by a string specifying
+a version of the language. Currently the only supported version is "0xa".
+
+Next, it contains a sequence of items specified at the top level. These items
 come one after another and it's recommended to separate them with newlines for
 readability but this isn't required.
 
 Concretely, the structure of a `wit` document is:
 ```
-wit-document ::= interface-item*
+wit-document ::= wit-version-declaration interface-item*
+
+wit-version-declaration ::= 'wit-version' strlit
 ```
 
 ## Item: `interface`


alexcrichton:
Do you know if  other IDLs have a precedent for something like this? I'm not sure myself how this is envisioned to concretely work in a form of language change, since even during development I wouldn't expect that all changes would bump this number and enforce compatibility checks.

alexcrichton:
If the precedent here is editions in Rust then I'm not sure that simply adding a version to the top of every file is going to have the desired effect. Some features of editions that I don't think are reflected in a version at the top of every file are:

* Editions are largely "out of the way" where you have a small annotation in one location that is rarely updated or ever manually written. For example `cargo new` generates whatever the current edition is and then that's updated on-demand as necessary, if ever. In contrast when hand-writing a `*.wit` file if a version must be included at the top then users will always need to know what the current version of `*.wit` is.
* Editions still require that there's a sort of "IR" that all known editions compile down to. For Rust this is MIR more-or-less but it means that tooling still has to support all editions for a long time. If we're interested in doing breaking changes to wit when it's more stable then I'm not sure if this is something we'd want to commit to. Instead it seems like if this will be used when wit is more stable then it'd be best to plan for extending the format backwards-compatibly as best we can.

Personally I feel that pressure should be on getting `*.wit` to a stable format which can be evolved backwards compatibly rather than trying to have a version number as an "out" where it's not necessarily clear to me how such a version number would be used. If there was a rough plan for how this version number might be increased and what the timeline would look like for such a change it might help me at least see how this would be leveraged.

neal-borelli:
I would agree with the idea of having a syntax version in the WIT file as well. Another example of this is that might be a little closer to use in an IDL is gRPC `.proto` files that start with a syntax statement.  For example:
`syntax = "proto3";`

I am less sure about how the versions should be expressed especially in this early portion of the project. My concern is that using the year as a version might not be granular enough.  I do like the human readable token that @guybedford suggested, but the "wit-" portion may not be needed since it is syntax in a WIT file.  However, from the gRPC proto file example above, they did include "proto" in the human readable token.

autodidaddict:
>Personally I feel that pressure should be on getting *.wit to a stable format which can be evolved backwards compatibly rather than trying to have a version number as an "out" where it's not necessarily clear to me how such a version number would be used.

Agreed with @alexcrichton here. It feels early to be dropping a version number in there when wit hasn't made it to a stable first syntax yet.  When there's nothing to compare a version number against, it seems superfluous and potentially a source of confusion.

The lack of version number can always be reasonably interpreted as a default value by parsers from the future (this is actually what protobuf does, too - it assumes a value of `2` for syntax if you don't supply it).

lukewagner:
I can imagine a pre-standard point in time where we want to start encouraging Wit to be used more-widely to encourage more feedback and, at that point, it would make sense to stick in a Wit-version to set an expectation of stability while still allowing intentional breaking version changes going forward.  But it feels like we're not there yet.

sunfishcode:
The problem I'm aiming to solve here is that we're sometimes getting bogged down trying to think through all possible future use cases to ensure that we have the perfect solution. My idea is to give our future selves an escape hatch.

> Editions are largely "out of the way"

That's true. Unfortunately, we only have one file to work with here, the wit.

> Editions still require that there's a sort of "IR" 

I don't think that's necessary here. Editions in Rust support linking between crates with different editions, and all Rust toolchains support all prior editions, but neither of those would be a requirement here (unless we want them to be).

> The lack of version number can always be reasonably interpreted as a default value by parsers from the future (this is actually what protobuf does, too - it assumes a value of 2 for syntax if you don't supply it).

If we add a version now, then when we reach a point where we're confident we won't need further backwards-incompatible changes, we can make the version optional, and it'll just fade away. Everyone will get the good behavior by default.

If we don't add a version now, but we end up needing one, then we end up in a place where "no version" will mean "the old version", forever, like it is in protobuf, and everyone will have to add an explicit version forever.

autodidaddict:
>If we don't add a version now, but we end up needing one, then we end up in a place where "no version" will mean "the old version", forever, like it is in protobuf, and everyone will have to add an explicit version forever.
Good point. emulating protobuf isn't always a good thing :)

design/mvp/WIT.md
```diff
@@ -87,7 +87,8 @@ bare as an identifier. These are used to help parse the format, and the list of
 keywords is still in flux at this time but the current set is:
 
 ```wit
-keyword ::= 'use'
+keyword ::= 'wit-version'
```
eduardomourar, line 90: why not make this just `version` if you have wit in the file extension?
```suggestion
keyword ::= 'version'
```

design/mvp/WIT.md
```diff
@@ -116,13 +117,18 @@ keyword ::= 'use'
 
 ## Top-level items
 
-A `wit` document is a sequence of items specified at the top level. These items
+A `wit` document starts with a `wit-version` followed by a string specifying
```
eduardomourar, line 120: to make the transition smoother, the version can be optional with a default value

design/mvp/WIT.md
```diff
@@ -116,13 +117,18 @@ keyword ::= 'use'
...
+a version of the language. Currently the only supported version is "0xa".
```
eduardomourar, line 121: I would suggest we follow [semantic versioning](https://semver.org/) (at least in the WIT IDL), but could be something more concise in binary format. Similar examples can be found in the [OpenAPI](https://swagger.io/specification/#appendix-a-revision-history) and in the [Smithy IDL](https://smithy.io/2.0/spec/idl.html#smithy-version).

```suggestion
a version of the language. Currently the only supported version is "1.0.0".
```

lukewagner:
I'd like to close this, as least for now, since we're using the component-model defined binaries as our stable representation of packages of Wit documents, and they have a version built into them (in the preamble) that we can bump until the C-M hits stable.  But feel free to reopen if you'd like to continue discussing.

PR #129, opened 2022-11-18
Add resource, own and borrow types

lukewagner:
(I'm publishing in draft state in case anyone wants to look at it before all the TODOs are filled in.  I think the essential additions in the AST and binary format are roughly right, I'm just filling in a bunch of explanatory prose/examples and the Canonical ABI Python definitions.)

This PR adds `resource`, `own` and `borrow` type constructors to the component model AST, binary format, text format which can serve as the underlying component-model semantics targeted by the current Wit syntax for resources.  The biggest change here is that, instead of having one shareable ref-counted handle type, we have two handle types: `own` representing unique ownership (like a `std::unique_ptr` in C++) and `borrow` representing a reference that can only be held onto for the duration of an export call (like a reference in C++ or Rust).  Based on feedback from the previous experiments, this unique ownership is important for performance, allowing host implementations to know, when they are passed a host resource (e.g., an HTTP Request) that the guest wasm code is no longer holding onto a handle, thereby allowing the host to do whatever it wants with the memory without first having to preserve a copy.  This also has software-engineering/composability benefits, preventing accidental interference when two components are linked together and pass handles between themselves.  There's more explanation of this in the PR itself (well, currently as TODOs, but there will be soon).

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -260,7 +260,7 @@ Notes:
 * Validation rejects `resourcetype` type definitions inside `componenttype` and
   `instancettype`. Thus, handle types inside a `componenttype` can only refer
   to resource types that are imported or exported.
-* Validation requires the all types used in the type of an export are introduced
+* Validation requires all types used in the type of an export are introduced
   by an `exportdecl` (not an `importdecl`).
 * The uniqueness validation rules for `externname` described below are also
   applied at the instance- and component-type level.
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -260,8 +260,8 @@ Notes:
 * Validation rejects `resourcetype` type definitions inside `componenttype` and
   `instancettype`. Thus, handle types inside a `componenttype` can only refer
   to resource types that are imported or exported.
-* Validation requires all types used in the type of an export are introduced
-  by an `exportdecl` (not an `importdecl`).
+* Validation requires that all resource types transitively used in the type of an
+  export are introduced by a preceding `exportdecl`.
 * The uniqueness validation rules for `externname` described below are also
   applied at the instance- and component-type level.
 * Validation of `externdesc` requires the various `typeidx` type constructors
@@ -334,8 +334,8 @@ URL        ::= b*:vec(byte)                    => char(b)*, if char(b)* parses a
 Notes:
 * All exports (of all `sort`s) introduce a new index that aliases the exported
   definition and can be used by all subsequent definitions just like an alias.
-* Validation requires that any handle type that is part of the type of any
-  export must refer to a resource type whose index was produced by an `export`.
+* Validation requires that all resource types transitively used in the type of an
+  export are introduced by a preceding `exportdecl`.
 * The "parses as a URL" condition is defined by executing the [basic URL
   parser] with `char(b)*` as *input*, no optional parameters and non-fatal
   validation errors (which coincides with definition of `URL` in JS and `rust-url`).
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -80,8 +80,12 @@ sort                ::= 0x00 cs:<core:sort>                                => co
                       | 0x05                                               => instance
 inlineexport        ::= n:<name> si:<sortidx>                              => (export n si)
 name                ::= len:<u32> n:<name-chars>                           => n (if len = |n|)
-name-chars          ::= w:<word>                                           => w
-                      | n:<name> 0x2d w:<word>                             => n-w
+name-chars          ::= l:<label>                                          => l
+                      | '[constructor]' r:<label>                          => [constructor]r
+                      | '[method]' r:<label> '.' m:<label>                 => [method]r.m
+                      | '[static]' r:<label> '.' s:<label>                 => [static]r.s
+label               ::= w:<word>                                           => w
+                      | l:<label> '-' w:<word>                             => l-w
 word                ::= w:[0x61-0x7a] x*:[0x30-0x39,0x61-0x7a]*            => char(w)char(x)*
                       | W:[0x41-0x5a] X*:[0x30-0x39,0x41-0x5a]*            => char(W)char(X)*
 ```
@@ -99,8 +103,21 @@ Notes:
   sort, but would be extended to accept other sorts as core wasm is extended.
 * Validation of `instantiate` requires that `name` is present in an
   `externname` of `c` (with a matching type).
+* When validating `instantiate`, after each individual type-import is supplied
+  via `with`, the actual type supplied is immediately substituted for all uses
+  of the import, so that subsequent imports and all exports are now specialized
+  to the actual type.
 * The indices in `sortidx` are validated according to their `sort`'s index
   spaces, which are built incrementally as each definition is validated.
+* Validation requires that all annotated `name`s only occur on `func` `export`s
+  and that the `r` `label` matches the `name` of a preceding `resource` export.
+* Validation of `[constructor]` names requires that the `func` returns a
+  `(result (own $R))`, where `$R` is the resource labeled `r`.
+* Validation of `[method]` names requires the first parameter of the function
+  to be `(param "self" (borrow $R))`, where `$R` is the resource labeled `r`.
+* Validation of `[method]` and `[static]` names ensures that all field names
+  are disjoint.
+
 
 ## Alias Definitions
 
@@ -120,7 +137,8 @@ Notes:
   index in the `sort` index space of the `i`th enclosing component (counting
   outward, starting with `0` referring to the current component).
 * For `outer` aliases, validation restricts the `sort` to one
-  of `type`, `module` or `component`.
+  of `type`, `module` or `component` and additionally requires that the
+  outer-aliased type is not a `resource` type (which is generative).
 
 
 ## Type Definitions
@@ -174,25 +192,28 @@ primvaltype   ::= 0x7f                                    => bool
                 | 0x74                                    => char
                 | 0x73                                    => string
 defvaltype    ::= pvt:<primvaltype>                       => pvt
-                | 0x72 nt*:vec(<namedvaltype>)            => (record (field nt)*)
+                | 0x72 lt*:vec(<labelvaltype>)            => (record (field lt)*)
                 | 0x71 case*:vec(<case>)                  => (variant case*)
                 | 0x70 t:<valtype>                        => (list t)
                 | 0x6f t*:vec(<valtype>)                  => (tuple t*)
-                | 0x6e n*:vec(<name>)                     => (flags n*)
-                | 0x6d n*:vec(<name>)                     => (enum n*)
+                | 0x6e l*:vec(<label>)                    => (flags l*)
+                | 0x6d l*:vec(<label>)                    => (enum l*)
                 | 0x6c t*:vec(<valtype>)                  => (union t*)
                 | 0x6b t:<valtype>                        => (option t)
                 | 0x6a t?:<valtype>? u?:<valtype>?        => (result t? (error u)?)
-namedvaltype  ::= n:<name> t:<valtype>                    => n t
-case          ::= n:<name> t?:<valtype>? r?:<u32>?        => (case n t? (refines case-label[r])?)
+                | 0x69 i:<typeidx>                        => (own i)
+                | 0x68 i:<typeidx>                        => (borrow i)
+labelvaltype  ::= l:<label> t:<valtype>                   => l t
+case          ::= l:<label> t?:<valtype>? r?:<u32>?       => (case l t? (refines case-label[r])?)
 <T>?          ::= 0x00                                    =>
                 | 0x01 t:<T>                              => t
 valtype       ::= i:<typeidx>                             => i
                 | pvt:<primvaltype>                       => pvt
+resourcetype  ::= 0x3f 0x7f f?:<funcidx>?                 => (resource (rep i32) (dtor f)?)
 functype      ::= 0x40 ps:<paramlist> rs:<resultlist>     => (func ps rs)
-paramlist     ::= nt*:vec(<namedvaltype>)                 => (param nt)*
+paramlist     ::= lt*:vec(<labelvaltype>)                 => (param lt)*
 resultlist    ::= 0x00 t:<valtype>                        => (result t)
-                | 0x01 nt*:vec(<namedvaltype>)            => (result nt)*
+                | 0x01 lt*:vec(<labelvaltype>)            => (result lt)*
 componenttype ::= 0x41 cd*:vec(<componentdecl>)           => (component cd*)
 instancetype  ::= 0x42 id*:vec(<instancedecl>)            => (instance id*)
 componentdecl ::= 0x03 id:<importdecl>                    => id
@@ -210,17 +231,37 @@ externdesc    ::= 0x00 0x11 i:<core:typeidx>              => (core module (type 
                 | 0x04 i:<typeidx>                        => (instance (type i))
                 | 0x05 i:<typeidx>                        => (component (type i))
 typebound     ::= 0x00 i:<typeidx>                        => (eq i)
+                | 0x01                                    => (sub resource)
 ```
 Notes:
 * The type opcodes follow the same negative-SLEB128 scheme as Core WebAssembly,
   with type opcodes starting at SLEB128(-1) (`0x7f`) and going down,
   reserving the nonnegative SLEB128s for type indices.
 * Validation of `valtype` requires the `typeidx` to refer to a `defvaltype`.
-* Validation of `instancedecl` (currently) only allows `outer` `type` `alias`
-  declarators.
+* Validation of `own` and `borrow` requires the `typeidx` to refer to a
+  resource type.
+* Validation only allows `borrow` to be used inside the `param` of a `functype`.
+  (This is likely to change in a future PR, converting `functype` into a
+  compound type constructor analogous to `moduletype` and `componenttype` and
+  using scoping to enforce this constraint instead.)
+* Validation of `resourcetype` requires the destructor (if present) to have
+  type `[i32] -> []`.
+* Validation of `instancedecl` (currently) only allows the `type` and
+  `instance` sorts in `alias` declarators.
 * As described in the explainer, each component and instance type is validated
   with an initially-empty type index space. Outer aliases can be used to pull
   in type definitions from containing components.
+* `exportdecl` introduces a new type index that can be used by subsequent type
+  definitions. In the `(eq i)` case, the new type index is effectively an alias
+  to type `i`. In the `(sub resource)` case, the new type index refers to a
+  *fresh* abstract type unequal to every existing type in all existing type
+  index spaces. (Note: *subsequent* aliases can introduce new type indices
+  equivalent to this fresh type.)
+* Validation rejects `resourcetype` type definitions inside `componenttype` and
+  `instancettype`. Thus, handle types inside a `componenttype` can only refer
+  to resource types that are imported or exported.
+* Validation requires the all types used in the type of an export are introduced
+  by an `exportdecl` (not an `importdecl`).
 * The uniqueness validation rules for `externname` described below are also
   applied at the instance- and component-type level.
 * Validation of `externdesc` requires the various `typeidx` type constructors
@@ -239,6 +280,9 @@ Notes:
 ```
 canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift f opts type-index-space[ft])
            | 0x01 0x00 f:<funcidx> opts:<opts>                   => (canon lower f opts (core func))
+           | 0x02 t:<typeidx>                                    => (canon resource.new t (core func))
+           | 0x03 t:<valtype>                                    => (canon resource.drop t (core func))
+           | 0x04 t:<typeidx>                                    => (canon resource.rep t (core func))
 opts     ::= opt*:vec(<canonopt>)                                => opt*
 canonopt ::= 0x00                                                => string-encoding=utf8
            | 0x01                                                => string-encoding=utf16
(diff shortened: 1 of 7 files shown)


design/mvp/Binary.md
```diff
@@ -80,8 +80,12 @@ sort                ::= 0x00 cs:<core:sort>                                => co
                       | 0x05                                               => instance
 inlineexport        ::= n:<name> si:<sortidx>                              => (export n si)
 name                ::= len:<u32> n:<name-chars>                           => n (if len = |n|)
-name-chars          ::= w:<word>                                           => w
-                      | n:<name> 0x2d w:<word>                             => n-w
+name-chars          ::= l:<label>                                          => l
+                      | '[constructor]' l:<label>                          => [constructor]l
+                      | '[method]' l:<label> '.' m:<label>                 => [method]l.m  
+                      | '[static]' l:<label> '.' s:<label>                 => [static]l.s
```
guybedford, line 86: Can we add `[get]` and `[set]` to the options here?

guybedford (commented):
Very excited to see this, seems like it will solve a lot of use cases!

design/mvp/Binary.md
lukewagner: Yeah, I was wondering about those too; I think they'd make sense are are a natural-enough completion of this set of standard OOP idioms.

lukewagner: (Included in the most recent draft, although I'm not quite yet ready to undertake how this should look in Wit.)

lukewagner: On second thought, since there's probably some interesting discussion to be had for how this should look in Wit and how the language bindings should work, I think this addition belongs in a separate PR, so we can avoid making this already-a-bit-big PR bigger.

lukewagner:
Alright, all TODOs have been filled in, along with the CanonicalABI Python code/tests, so I think this is ready to review.

design/mvp/Binary.md
```diff
@@ -210,17 +231,37 @@ externdesc    ::= 0x00 0x11 i:<core:typeidx>              => (core module (type
...
+* `exportdecl` introduces a new type index that can be used by subsequent type
+  definitions. In the `(eq i)` case, the new type index is effectively an alias
+  to type `i`. In the `(sub resource)` case, the new type index refers to a
+  *fresh* abstract type unequal to every existing type in all existing type
+  index spaces. (Note: *subsequent* aliases can introduce new type indices
+  equivalent to this fresh type.)
+* Validation rejects `resourcetype` type definitions inside `componenttype` and
+  `instancettype`. Thus, handle types inside a `componenttype` can only refer
+  to resource types that are imported or exported.
+* Validation requires the all types used in the type of an export are introduced
```
ricochet, line 263: Validation requires ~~the~~ all

ricochet:
Looks great!

> Handle types (`own` and `borrow`) are purely structural, but, since they refer
to resource types, transitively "inherit" the freshness of abstract resource
types.

If `own` and `borrow` resource handles are purely structural, does that mean a list could contain a mix of `own` and `borrow` resource handles? I don't think this is the intention and would loose helpful semantics that hosts could optimize around.

I think practically with how the AST is constructed in the example, the answer to above is that this spec treats `own` and `borrow` resource handles as not equal and could not be added to the same list. The examples with lists as resources are all homogenous. In this case the list only contains owned handles to a resource:
 
```text
    (import "T" (type $T (sub resource)))
    (import "i" (func (param (list (own $T)))))
```

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+    (export "r1" (type (sub $r)))
+    (export "r2" (type (sub $r)))
+  )
+  (instance $c (instantiate $C))
+  (type $r1 (alias export $c "r1"))
+  (type $r2 (alias export $c "r2"))
+)
+```
+The types `$r1` and `$r2` are unequal. These type-checking rules for
+type exports mirror the *introduction* rule of [existential types]  (∃T).
```
rossberg, line 932: Yeah, here I'm not really sure why you tie this one to introduction and the earlier to elimination. I think they're pretty much symmetric, except that one uses export abstraction and the other "definitional" abstraction. In both cases the "magic" already happens on the intro side.

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+
+When supplying a resource type (imported *or* defined) to a type import via
+`instantiate`, type checking performs a substitution, replacing all uses of the
+`import` in the instantiated component with the actual type supplied via
+`with`. This means that, from the parent's point of view, the type imports stop
+being abstract. For example, the following component validates:
```
rossberg, line 938: Well, unless they are substituted with other abstract types. ;) But I suppose it's fine to gloss over that.

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+  (alias export $c1 "foo" (func $foo))
+  (instance $c2 (instantiate $C2 (with "T" (type $R)) (with "foo" (func $foo))))
+)
+```
+This depends critically on the `T` imports of `$C1` and `$C2` having been
+replaced by `$R` when validating the instantiations of `$c1` and `$c2`. These
+type-checking rules for instantiating type imports mirror the *elimination*
+rule of [universal types]  (∀T).
+
+As a final constraint on both the `export` declarators of a `componenttype` and
```
rossberg, line 960: I think this applies not just to component types, but also components themselves.

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+the `export` definitions of a `component`, all transitive uses of abstract
+types in the types of exported functions or values must refer to *exported*
```
rossberg, line 962: ...and in the case of components, "abstract types" indeed needs to include generative types here.

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+bound are *abstract (resource) types* and follow the standard introduction and
+elimination rules of universal and existential types.
+
+Lastly, since "nominal" is sometimes intended to mean "the opposite of
+structural", a valid question is whether any of this is "nominal typing".
+Unfortunately, "nominal" has a fuzzy meaning in modular contexts like the
+Component Model where there is [necessarily](../high-level/Choices.md) no
+shared global namespace. Instead, the main use cases for nominal types are
+independently addressed by the addition of abstract and generative types
+mentioned above. In particular, while importing or exporting an abstract type
```
rossberg, line 993: You could phrase that slightly differently, to perhaps make the connection clearer. First, I would say that the common interpretation of nominal types corresponds most closely to generative types above. However, the notion of generative _generalises_ nominal in a way that still makes sense when a component can be instantiated multiple times, where the same local "name" now denotes multiple distinct types.

rossberg (commented):
Overall, the type checking explanations read really good, nicely done! I spotted a few inaccuracies, but I think they are not difficult to tweak.

rossberg:
One point I'm still not entirely clear about is why we actually need bounds on type exports. This is (for components as opposed to component types) inconsistend with other exports. Moreover, it introduces the whole machinery of ML-style "sealing", which is complex and may be overkill here. Do you have a particular use case for it? (FWIW, I specifically tried to avoid that complexity and asymmetry with the design sketch for private types in core Wasm.)

Without those bounds, there would be just one form of existential, and that's a resource. And the need for exports to bind indices (another asymmetry with core Wasm) would go away as well, since they are always equivalent to some other type.

design/mvp/Binary.md
```diff
@@ -174,25 +192,28 @@ primvaltype   ::= 0x7f                                    => bool
...
-case          ::= n:<name> t?:<valtype>? r?:<u32>?        => (case n t? (refines case-label[r])?)
+                | 0x69 i:<typeidx>                        => (own i)
+                | 0x68 i:<typeidx>                        => (borrow i)
+labelvaltype  ::= l:<label> t:<valtype>                   => l t
+case          ::= l:<label> t?:<valtype>? r?:<u32>?       => (case l t? (refines case-label[r])?)
 <T>?          ::= 0x00                                    =>
                 | 0x01 t:<T>                              => t
 valtype       ::= i:<typeidx>                             => i
                 | pvt:<primvaltype>                       => pvt
+resourcetype  ::= 0x3f 0x7f f?:<funcidx>?                 => (resource (rep i32) (dtor f)?)
```
alexcrichton, line 212: Should this perhaps be prefixed with `0x3f 0x01 0x7f` instead of just the bare `0x7f`?

design/mvp/Binary.md
```diff
@@ -210,17 +231,37 @@ externdesc    ::= 0x00 0x11 i:<core:typeidx>              => (core module (type
...
+  definitions. In the `(eq i)` case, the new type index is effectively an alias
+  to type `i`. In the `(sub resource)` case, the new type index refers to a
+  *fresh* abstract type unequal to every existing type in all existing type
+  index spaces. (Note: *subsequent* aliases can introduce new type indices
+  equivalent to this fresh type.)
+* Validation rejects `resourcetype` type definitions inside `componenttype` and
+  `instancettype`. Thus, handle types inside a `componenttype` can only refer
+  to resource types that are imported or exported.
+* Validation requires the all types used in the type of an export are introduced
+  by an `exportdecl` (not an `importdecl`).
```
alexcrichton, line 264: I think that this rule may be a bit over-restrictive because while we want to name `record foo` I don't think it's necessary to name a `string`, for example. I think instead this would need to categorize what can be named and what's allowed to not be named?

design/mvp/Binary.md
```diff
@@ -210,17 +231,37 @@ externdesc    ::= 0x00 0x11 i:<core:typeidx>              => (core module (type
...
+  `instance` sorts in `alias` declarators.
 * As described in the explainer, each component and instance type is validated
   with an initially-empty type index space. Outer aliases can be used to pull
   in type definitions from containing components.
+* `exportdecl` introduces a new type index that can be used by subsequent type
+  definitions. In the `(eq i)` case, the new type index is effectively an alias
+  to type `i`. In the `(sub resource)` case, the new type index refers to a
+  *fresh* abstract type unequal to every existing type in all existing type
+  index spaces. (Note: *subsequent* aliases can introduce new type indices
+  equivalent to this fresh type.)
```
alexcrichton, line 259: I think this is where what we were talking about yesterday would be inserted, namely the export of an instance type would create an instance so types could be aliased from it? (or have I perhaps missed this from elsewhere?)

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+When supplying a resource type (imported *or* defined) to a type import via
+`instantiate`, type checking performs a substitution, replacing all uses of the
+`import` in the instantiated component with the actual type supplied via
+`with`. This means that, from the parent's point of view, the type imports stop
+being abstract. For example, the following component validates:
+```wasm
+(component $P
+  (import "C1" (component $C1
+    (import "T" (type $T (sub resource)))
+    (export "foo" (param (own $T)))
```
alexcrichton, line 943: I think this is missing a `(func ...)` wrapper

design/mvp/canonical-abi/definitions.py
```diff
@@ -290,7 +306,117 @@ class CanonicalOptions:
...
+    self.array = []
+    self.free = []
+
+  def insert(self, cx, h):
+    match h:
+      case OwnHandle():
+        assert(len(h.resource.borrowers) == 0)
+      case BorrowHandle():
+        if cx.inst in h.resource.borrowers:
+          return h.resource.borrowers[cx.inst]
```
alexcrichton, line 373: Is it intended to update the `borrow_count` here? Without doing that I'm worried that languages need to maintain their own mirrored borrow count as opposed to always executing the same destructor given the structure of a type.

alexcrichton (approved):
This all looks reasonable to me, thanks again for spearheading this @lukewagner! 

I do think that this is going to require significant thought/design/etc on the implementation side of things, both for guest languages in addition to hosts. Nothing seems too intractable per se but an implementation will be necessary to suss out anything concrete.

design/mvp/Binary.md
lukewagner: Good catch, [fixed](https://github.com/WebAssembly/component-model/pull/129/commits/08cbd80a652b26103a90a1fe5e0d9efe70374280)

lukewagner:
@ricochet Thanks for the review!  You're right, `own` and `borrow` handles are completely separate types and thus can't be mixed in a single list.  I expanded the example you're referring to [here](https://github.com/WebAssembly/component-model/pull/129/commits/fafa1f26fdbbc5f5db2f76e61b058a162b5ef161) to include `borrow`s and describe their inequality with the `own`s.

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+```wasm
+(component $A
+  (type $ListString1 (list string))
+  (type $ListListString1 (list $ListString1))
+  (type $ListListString2 (list $ListString1))
+  (component $B
+    (type $ListString3 (list string))
+    (type $ListListString3 (list $ListString3))
+    (type $ListString4 (alias outer $A $ListString))
+    (type $ListListString4 (list $ListString4))
```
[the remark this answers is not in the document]
lukewagner: Good idea, [added](https://github.com/WebAssembly/component-model/pull/129/commits/89724c317e6ddf82466b3dd8e22a70f32f1bfe50)

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+following component:
+```wasm
+(component
+  (import "T1" (type $T1 (sub resource)))
+  (import "T2" (type $T2 (sub resource)))
+)
+```
+the types `$T1` and `$T2` are not equal. If type imports worked purely
+structurally, then they would be equal because structurally they have the same
+AST.
```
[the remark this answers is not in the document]
lukewagner: Yeah, I felt awkward writing this for the same reason.  When I said "have the same AST", I meant "literally written the same" (treating the sub bound as just part of the AST), but perhaps this all confuses more than it helps, so [removed](https://github.com/WebAssembly/component-model/pull/129/commits/253dfd7594e97e81c04c19e0ff32048f2375fcba).

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+    (export "T3" (type $T3 (eq $T2)))
+  ))
+  (type $T1 (alias export $i "T1"))
+  (type $T2 (alias export $i "T2"))
+  (type $T3 (alias export $i "T3"))
+)
+```
+the types `$T2` and `$T3` are equal to each other but not to `$T1`. These
+type-checking rules for aliases of type exports mirror the *elimination* rule
+of [existential types]  (∃T).
```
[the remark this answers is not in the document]
lukewagner: Great points!  I forgot that exports of imported instances turn into imports, so I made it a component import as you suggested [here](https://github.com/WebAssembly/component-model/pull/129/commits/475c8553bfff4295ffd8cfdd46dd5b63bf10caea).

For the existential intro rule, does [this revised example](https://github.com/WebAssembly/component-model/pull/129/commits/60df3c79969345e90a770f18b9c13f1f2c7f17d3) work?

design/mvp/Explainer.md
lukewagner: Oh hah, right.  Yeah, I can't think of a succinct replacement, so perhaps better just to [nix](https://github.com/WebAssembly/component-model/pull/129/commits/a85fe110e675b800c1516c611082577013b4406a) the line.

design/mvp/Explainer.md
lukewagner: I might be missing your point, but I think this sentence says that this requirement applies to both type-level and component-level exports?

design/mvp/Explainer.md
lukewagner: Ah, yes, good point!  If we classify resource types as "abstract" (as you half-suggest above), this clause grows to cover generative types, I think?

design/mvp/Explainer.md
lukewagner: Ooh, yeah, I like that perspective of "generative generalizes nominal".  Does [this](https://github.com/WebAssembly/component-model/pull/129/commits/62a94bbe10f81b15cef43f36ac3fc69013077863) sound right to you and capture what you were going for?

design/mvp/Binary.md
lukewagner: That's a great question and I've flip-flopped on this (particularly since showing it last time).  Thinking about the implementation, I think there's a real chance we won't ever want >1 elements (or even scalar `v128`) since it'll ruin the otherwise dense-array-friendliness of `i32|i64` and force the host to do some out-of-line allocation that may end up being slower than if the `rep` was an `i32` pointer into linear memory.  So I thought I'd just remove the pretense of maybe doing this any time soon and if, in the future, we do want to do it, we can always do so in a backwards-compatible manner by introducing a second opcode in the binary format for `resource` that has a `vec`.  But it's just a byte so I could be convinced either way; I just think there's a 95% chance it's a wart :)  WDYT?

design/mvp/Explainer.md
lukewagner: Oops, right, thanks, [fixed](https://github.com/WebAssembly/component-model/pull/129/commits/bfc2ba739c5ad93d856d10c3c2b0154a0ba77770)!

design/mvp/canonical-abi/definitions.py
lukewagner: That's a great question and one I've wondered about for that exact reason.  My current thinking is: to support this, we'd need to add a new reference count in `BorrowHandle` (the `OwnHandle.lend_count` isn't quite the right place because a single `OwnHandle` can be lent to many instances) but then some guest code will end up needing to duplicate this refcount in the guest code *anyway* because they'll care about when the index is officially removed from the `HandleTable` so that they can clear out associated state.  Thus, I was thinking that having the host do this is either a wash or net-loss, but this is based on some extrapolations and assumptions, so I might be wrong and I'd be interested to hear your take (now or later during impl).

design/mvp/Explainer.md
rossberg: Ah yes, scrap that, I was somehow missing half the sentence while reading.

design/mvp/Explainer.md
rossberg: That's right.

design/mvp/Explainer.md
```diff
@@ -996,25 +996,19 @@ In summary: all type constructors are *structural* with the exception of
...
+valid question is whether any of the above "nominal typing". Inside a
+component, resource types act "nominally": each resource type definition
+produces a new local "name" for a resource type that is distinct from all
+preceding resource types. The interesting case is when resource type equality
+is considered from *outside* the component, particularly when a single
+component is instantiated multiple times. In this case, a single resource type
+definition that is exported with a single `externname` will get a fresh type
+with each component instance, with the abstract typing rules mentioned above
+ensuring that each of the component's instance's resource types are kept
+distinct. Thus, in a sense, the generativity of resource types *generalize*
```
rossberg, line 1009: *generalizes*

design/mvp/Explainer.md
rossberg: Yes, sgtm.

lukewagner:
@rossberg 

> One point I'm still not entirely clear about is why we actually need bounds on type exports. This is (for components as opposed to component types) [...]

I might be missing your point, but I think the proposal already does what you're suggesting: top-level export definitions of types do not have bounds ([here](https://github.com/WebAssembly/component-model/blob/add-resource-types/design/mvp/Explainer.md#import-and-export-definitions)); only type-level export declarators do.

design/mvp/Explainer.md
lukewagner: Thanks, [fixed](https://github.com/WebAssembly/component-model/pull/129/commits/778297ea206ecd40d2fb8d91c701619ea49d8f7b)

design/mvp/Binary.md
alexcrichton: Sounds reasonable to me 👍

design/mvp/Binary.md
alexcrichton: Indeed! 👍

design/mvp/canonical-abi/definitions.py
alexcrichton: From a Rust guest perspective what I'm worried about is that something which naively follows a mostly RAII-style paradigm is no longer compatible here. For example with an owned handle we can wrap it up in a struct with a destructor and I think that makes sense. For a borrow handle I would expect to do the same but that subtly can't be done and require global tracking of how many borrows are active for a specific handle, and the close function is only invoked when that global tracking reaches 0.

So in some sense preserving the exact same borrow here, while good for JS, is bad for RAII-style resource management. RAII, however, is bad for JS where it wants to map to the same JS object. (or at least that's my read on this)

One possible idea would be to have a canoical ABI option controlling the lowering which could possibly differ between JS and Rust. There could also be another intrinsic for "get unique id" where two different integer descriptors (representing two separate borrows) could point to the same unique id to handle the JS case. Neither of those feels that great though.

design/mvp/Explainer.md
```diff
@@ -624,12 +706,304 @@ and out-of-line type definitions:
...
+the types `$T2` and `$T3` are equal to each other but not to `$T1`. These
+type-checking rules for aliases of type exports mirror the *elimination* rule
+of [existential types]  (∃T).
+
+Next, we consider resource type definitions, which are half-like abstract types
+in that they also produce a *fresh* type that is unequal to every preceding
+type. However, a resource type definition is **not** abstract; it has a
+concrete representation that is known to the containing component (and *only*
+the containing component). Thus, local resource type definitions are neither
+structural nor abstract: they are "[generative]". For example, in the following
```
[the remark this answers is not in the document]
lukewagner: Hm, I guess that makes sense.  It is indeed rather awkward to succinctly describe, here's my [attempt](https://github.com/WebAssembly/component-model/pull/129/commits/1eb69c1ce65c57b0b65adaa92404e8d9a725920d), lmkwyt.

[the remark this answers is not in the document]
rossberg: sgtm

rossberg:
@lukewagner:
> I might be missing your point, but I think the proposal already does what you're suggesting: top-level export definitions of types do not have bounds ([here](https://github.com/WebAssembly/component-model/blob/add-resource-types/design/mvp/Explainer.md#import-and-export-definitions)); only type-level export declarators do.

You're right, I was confused. :)

design/mvp/canonical-abi/definitions.py
lukewagner: Ah, interesting idea!  I hadn't thought about having a `canonopt` to opt out of the per-inst handle uniqueness.  I'll need to think a bit more about what it means to not have this property and whether it causes any problems.

lukewagner: Thinking about this some more, it definitely makes sense to have the option to unconditionally lower borrows into new handle-table entries (without de-duplication) since many cases simply don't need the de-duplication and it has non-trivial cost.  There are a number of possibilities for how to add a de-duplication option, but it's not clear to me yet which exactly is the right one.  In part, the question is coupled to how exactly the forthcoming "child" handles work and how we want object-wrappers in dynamic languages to work with both `borrow` and `child` handles.  So for now, I've simply [removed](https://github.com/WebAssembly/component-model/pull/129/commits/e2cfe7ba89559fee0c5c2921367742c9c3b68160) the `borrowers` map from this PR with a tentative plan to include it back in conjunction with adding `child` in a subsequent PR.

lukewagner:
Thanks everyone for the feedback so far!  Since it's been quiet for a few days, I'll wait for any more feedback until the end of the week and if things still seem settled, merge this PR and continue to iterate in subsequent issues/PRs.

PR #139, opened 2022-12-02
Enums to be 1-indexed

guybedford:
In the guest C generator, it currently utilizes a pattern where a function that returns a `-> result<ok, code>` where `code` is an enum, will support outputting the common C-style error code function:

```c
typedef code_t uint8_t;
code_t func(ok_t* ret);
```

The success check on this function then effectively becomes - `code == -1` currently due to the `0` enum value being used, which is what is implemented currently in wit-bindgen.

If instead we start enums at 1, this check can be come a more standard `code == 0` success check.

I'm not sure the current ABI spec documents this coalescing fully yet, but at the very least it could be good to update the enum definition.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -74,11 +74,14 @@ def despecialize(t):
   match t:
     case Tuple(ts)         : return Record([ Field(str(i), t) for i,t in enumerate(ts) ])
     case Union(ts)         : return Variant([ Case(str(i), t) for i,t in enumerate(ts) ])
-    case Enum(labels)      : return Variant([ Case(l, None) for l in labels ])
+    case Enum(labels)      : return Variant([ Case("", None), ...Case(l, None) for l in labels ])
     case Option(t)         : return Variant([ Case("none", None), Case("some", t) ])
     case Result(ok, error) : return Variant([ Case("ok", ok), Case("error", error) ])
     case _                 : return t
 ```
+Enums start at the first index, permiting singular result code return patterns
+in corresponding core function ABIs.
+
 The specialized value types `string` and `flags` are missing from this list
 because they are given specialized canonical ABI representations distinct from
 their respective expansions.


sunfishcode:
In Unix API circles, `code == -1` is the standard error check. I fear changing this would trade one set of confused C programmers for another.

alexcrichton:
To clarify, right now in C bindings enums are 0-based which means that for the above API `-1` actually means success rather than failure, where any other code means failure. This is contradictory to at least Windows/Unix semantics AFAIK.

This change would allow 0 to be a "niche" where 0 could mean success and anything nonzero could be interpreted as failure. That's still not quite Unix semantics but otherwise changing that I think would require having a special `error-enum` type which is only represented with negative integers.

sunfishcode:
I'm unclear about what this is doing. Where in the CanonicalABI does it pack the `result` discriminant into the same integer as the `result`'s error code?

The most likely place would seem to be `flatten_variant` in the [flattening section](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#flattening), but it always uses a separate result for the discriminant.

peterhuene:
I feel compelled to voice an opposition to this change. I don't think we need to modify the ABI to facilitate a particular language binding's (idiomatic?) conventions.

Personally, I think the code generated for C bindings should always return a tagged union to model the `result` type rather than special casing enum error types and "flattening" the result discriminate in with the error enum values.

peterhuene:
Even if we don't want to model the result with a tagged enum specifically in the C bindings, I think even a binding of `bool func(..., ok_t*, err_t*)` would be preferable to reserving 0 for all enums in the ABI _just in case_ an enum might be used to convey the semantic meaning of an error in a result type.

guybedford:
The flattening here is not reflected in the canonical ABIs, it is a specific behaviour of the C generator. I appreciate not wanting to implement something for such a specific use case though. I wonder, are there benefits in other languages for starting at 1?

@alexcrichton as far as I'm aware, the current approach of using `-1` in the C generator effectively reserves the enum max - 1 value for a given enum size, _which I believe **is** part of the canonical ABI_, so I think we currently might have an overflow against the spec without something like this?

alexcrichton:
That was [my concern here](https://github.com/bytecodealliance/wit-bindgen/pull/440#issuecomment-1329277218) but you mentioned you tested and it worked out ok?

I think though that it's correct that returning `error_t` wouldn't work though. I also suspect that errors for most `*.wit` APIs are going to want to be more rich than just an error code, so I like @peterhuene's suggestion of `bool func(..., ok_t*, err_t*) would` for C APIs. that return results.

guybedford:
@alexcrichton yes I believe in theory that is an implementation bug that this doesn't overflow currently as there is no extra slot reserved by the enum sizing here.

Sounds like we have agreement then to update the C gen ABI as suggested and close this, unless there are any other arguments we should be considering in reserving enum space for other purposes?

guybedford:
WIT bindgen has been updated to avoid this case now in https://github.com/bytecodealliance/wit-bindgen/pull/450.

PR #141, opened 2022-12-06
Add `use` and a binary format to WIT.md

### LINKED ISSUE #140: Adding a `use` directive to `*.wit`
Currently as specified the `*.wit` format doesn't have a `use` directive or any similar ability to import/share definitions from other files. A historical implementation of this existed in `wit-bindgen` but was removed since it was never really fully fleshed out and working as one might expect. I wanted to open this issue, though, to serve as a place to talk about it.

@lukewagner, @fibonacci1729, and I just had a long chat about a possible design here and I wanted to additionally write down our current thinking for what this might look like. The current plan is for @fibonacci1729 and I to pair up and implement this both in `wit-parser` in the `wasm-tools` repository as well as through a PR here describing the semantics.

## Interfaces and `use`

```
interface http-types {
  resource request
  resource response
  type headers = list<tuple<string, string>>
  record error {
    kind: error-kind
    message: string
  }
  enum error-kind {
    too-sunny-outside,
    too-cold-outside,
    too-many-cats,
  }
}

interface http-backend {
  use { request, response, error } from http-types

  fetch: func(r: request) -> result<response, error>
}

world my-world {
  import backend1: http-backend
  import backend2: http-backend
}
```

Here the `http-backend` interface has a `use` directive which refers to types defined in the previous interface, `http-types`. The ordering of the interfaces doesn't matter here so you could also do:

```
interface http-backend {
  use { request, response, error } from http-types

  // ...
}

interface http-types {
  // ...
}
```

although `use` statements do not support cyclic references, they must form a DAG still. The `use` directive can, for now, only appear in `interface` and `world` blocks, not the top-level. This may be added at a future date, however. Names can also be renamed in a `use`:

```
interface foo {
    use { bar as different-name } from other-interface 
}
```

Additionally anything `use`'d is considered an export of the interface which means that `use`-s can be chained together:

```
interface i1 {
  type a = u32
}

interface i2 {
  use { a } from i1
}

interface i3 {
  use { a } from i2
}
```

This will also enable splitting one `*.wit` file into multiple `*.wit` files:

```
// types.wit
default interface types {
  enum errno {
    // ...
  }
}

// fs.wit
interface fs {
  use { errno } from "types"

  exists: func(path: string) -> result<bool, errno>
}
```

Here the `from` target is quoted to represent any filename being loaded from. The precise contents of the string are sort of nebulous for now and are eventually intended to be URL-like but for now will just be path names for local resolution. Note the usage of `default interface` in `types.wit` which indicates that when `"types"` is used from it knows which interface to pick by default. A non-default interface can be selected with:

```
use { errno } from some-non-default-interface in "types"
```

For now only types can be referred to by `use`. Eventually it might make sense to add functions as well but that's seen as a possible future feature.

## Worlds and `use`

World directives can be specified with strings now to refer to other files:

```
world foo {
    import wasi-fs: "wasi-fs" // requires `wasi-fs.wit` to exist
}
```

and this is equivalent to:

```
world foo {
    use { wasi-fs } from "wasi-fs"
    import wasi-fs: wasi-fs
}
```

In a `world` its `use` statements must refer to interfaces as opposed to types. Unlike `interface`s they're not imported/export but rather are just available for use in `import` or `export` directives.

## Translation to a component type

The intention is that a `use` statement basically equates to an `alias` in a component. For example:

```
interface types {
  enum errno {
    i-am-a-teapot,
  }
}

interface thread {
  use { errno } from types
  sleep: func(dur: u64) -> result<_, errno>
}

world my-world {
  import thread: thread
}
```

would translate to:

```
(component $C
  (import "types" (instance $types
    (type $errno (enum "i-am-a-teapot"))
    (export "errno" (type (eq $errno)))
  ))
  (alias export $types "errno" (type $errno))

  (import "thread" (instance $thread
    (alias outer $C $errno (type $errno))
    (export "errno" (type (eq $errno)))
    (export "sleep" (func (param "dur" u64) (result (result (error $errno)))))
  ))
)
```

Note here that `my-world` only imported `thread` but ended up also importing `types` in the final component. This is due to the `use` of the `types` interface in `thread`. Note that the string `"types"` comes from the name of the interface `types`. For cases where conflicts arise the import strings must be explicitly disambiguated:

```
// dir1/foo.wit
default interface foo {
    use { /* ... */ } from "shared"
}

// dir1/shared.wit
default interface shared {
    // ...
}

// dir2/bar.wit
default interface bar {
    use { /* ... */ } from "shared"
}

// dir2/shared.wit
default interface shared {
    // ...
}

// world.wit
world world-with-bug {
  // This generates an error since it requires `import "shared"` in the component for two 
  // different interfaces
  import foo: "dir1/foo"
  import bar: "dir2/bar"
}

world world-that-works {
  // explicitly disambiguate here and this works ok
  import shared1: "dir1/shared"
  import shared2: "dir2/shared"
  import foo: "dir1/foo"
  import bar: "dir2/bar"
}
```

## Resolution

Resolution of `use` statements is intended to follow a rough algorithm that looks like:

1. First resolve all quoted `use` statements and interface dependencies in worlds. This will involve reading external files, recursively, and collecting everything together. Note that at this point it should be possible to emit another `*.wit` file equivalent to the first. Note that cycles are disallowed here
2. Next resolve all `use` statements now that all `interface`s are known. Cycles are again disallowed here and connections are made.
3. Finally resolve each `interface individually, disalling name conflicts within an `interface`. For now everything in an `interface` will share a namespace, meaning types, functions, and interfaces all live in the same namespace and can have no duplicates.

The final resulting `*.wit` data structure still semantically has `use` statements inside of it. This is useful for generators such as:

* For `*.wasm` generation a `use` means "go alias something from somewhere else"
* For code generators printing a type definition for exported types a `use` means "skip this, import the name from somewhere else"
* For code generators printing the name of a type, a `use` means "print the name as used into this module"
* For lifting/lowering `use` is "unwrapped" to look at the structure of the type to determine how to lift and lower

This should provide a straightforward way for, even in code generators, to share types between interfaces since all that's necessary is to `use` between them.

alexcrichton:
This commit brings the `WIT.md` file up-to-date with recent developments in WIT, chifely the `use` syntax. I've added a new "introduction section" which is a higher-level explanation of the format than the lexical/syntactic structure. I've also edited the syntactic structure to clarify a few old references and remove `resource`, `future`, and `stream` types for now. (resources to be added soon I suspect)

Additionally I've added a description of a binary format for WIT based on the component model. This binary format is intended to be the "types only" mode of the `wasm-tools component` tooling, although it doesn't match the current implementation and the `wasm-tools` repo will need to be updated. This representation, however, provides the means by which to understand what it means for a component to have the type of a `world`, namely it's a subtype of the binary representation's component type.

Closes #140

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -102,10 +102,11 @@ not need to be specified as it's the `default`.
 ## WIT Worlds
 [worlds]: #wit-worlds
 
-WIT documents can contain a `world` annotation at the top-level in addition to
-[`interface`][interfaces]. A world is a complete description of both imports and
-exports of a component. A world can be thought of as an equivalent of a
-`component` type in the component model. For example this world:
+WIT documents can contain a `world` definition at the top-level in addition to
+[`interface`][interfaces] definitions. A world is a complete description of
+both imports and exports of a component. A world can be thought of as an
+equivalent of a `component` type in the component model. For example this
+world:
 
 ```wit
 world my-world {
@@ -136,13 +137,13 @@ Worlds can contain any number of imports and exports, and can be either a
 function or an interface.
 
 ```wit
-world wasi {
-  import wasi-fs: wasi.fs
-  import wasi-random: wasi.random
-  import wasi-clock: wasi.clock
+world command {
+  import fs: wasi-fs.fs
+  import random: wasi-random.random
+  import clock: wasi-clock.clock
   // ...
 
-  export command: func(args: list<string>)
+  export main: func(args: list<string>)
 }
 ```
 
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -513,30 +513,6 @@ characters and numbers that are `-` separated.
 
 For more information on this see the [binary format](./Binary.md).
 
-## WIT in Markdown
-[markdown]: #wit-in-markdown
-
-The WIT text format can also additionally be parsed from markdown files with the
-extension `wit.md`, for example `foo.wit.md`:
-
-    # This would be
-
-    Some markdown text
-
-    ```wit
-    // interspersed with actual `*.wit`
-    interface my-interface {
-    ```
-
-    ```wit
-    // which can be broken up between multiple blocks
-    }
-    ```
-
-Triple-fence code blocks with the `wit` marker will be extracted from a markdown
-file and concatenated into a single string which is then parsed as a normal
-`*.wit` file.
-
 # Lexical structure
 [lexical-structure]: #lexical-structure
 
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -160,7 +160,7 @@ interface out-of-line {
 
 world your-world {
   import out-of-line: self.out-of-line
-  // ... is roughtly equivalent to ...
+  // ... is roughly equivalent to ...
   import out-of-line2: interface {
     the-function: func()
   }
@@ -259,7 +259,7 @@ interface my-host-functions {
 }
 ```
 
-Here `more-types in the `use` path indicates that it's the specific interface
+Here `more-types` in the `use` path indicates that it's the specific interface
 being referenced. Documents in a WIT package must be named after a [valid
 identifier][identifiers] and be unique within the package. Documents cannot
 contain cycles between them as well with `use` statements.
@@ -303,7 +303,7 @@ do not need to conflict with dependency names in other packages. This enables
 each package to be resolved entirely separately and, if necessary, the name
 `foo` could mean different things to different packages.
 
-> **Note**: The tooling for and mechanism for precisly how these external names
+> **Note**: The tooling for and mechanism for precisely how these external names
 > are defined is not specified here. This is something that will be iterated on
 > to create documentation of the tooling in question and community standards
 > about how best to do this.
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -1124,12 +1124,13 @@ would correspond to:
 (component
   (type (export "host") (component
     (type $types (instance
-      (export "level" (type (enum "info" "debug")))
+      (type $level (enum "info" "debug"))
+      (export "level" (type (eq $level)))
     ))
     (export $types "types" (instance (type $types)))
     (alias export $types "level" (type $level))
     (type $console (instance
-      (alias outer 1 $level (type $level'))
+      (export $level' "level" (type (eq $level)))
       (export "log" (func (param "level" $level') (param "msg" string)))
     ))
     (export "console" (instance (type $console)))
@@ -1155,11 +1156,10 @@ would correspond to:
 ```wasm
 (component
   (type (export "host") (component
-    (type $the-world (component
-      (import "test" (func))
+    (export "the-world" (component
+      (export "test" (func))
       (export "run" (func))
     ))
-    (export "the-world" (type $the-world))
   ))
 )
 ```
@@ -1189,12 +1189,11 @@ would correspond to:
       (export "log" (func (param "arg" string)))
     ))
     (export "console" (instance (type $console)))
-    (type $the-world (component
+    (export "the-world" (component
       (import "console" (instance
         (export "log" (func (param "arg" string)))
       ))
     ))
-    (export "the-world" (type $the-world))
   ))
 )
 ```
@@ -1243,12 +1242,11 @@ would correspond to:
   (type (export "foo") (component
     (import "types" "URL" (instance $types
       (type $some-type ...)
-      (export "some-type" (type $some-type))
+      (export "some-type" (type (eq $some-type)))
     ))
     (alias export $types "some-type" (type $some-type))
     (type $foo (instance
-      (alias outer 1 $some-type (type $some-type'))
-      (export "some-type" (type $some-type'))
+      (export "some-type" (type (eq $some-type)))
     ))
     (export "foo" (instance (type $foo)))
   ))
@@ -1291,26 +1289,30 @@ and its corresponding binary encoding would be:
   ;; corresponds to `wit/types.wit`
   (type (export "types") (component
     (export "types" (instance
-      (export "request" (type (sub resource)))
-      (export "response" (type (sub resource)))
+      (type $request (record))
+      (type $response (record))
+      (export "request" (type (eq $request)))
+      (export "response" (type (eq $response)))
     ))
   ))
   ;; corresponds to `wit/handler.wit`
   (type (export "handler") (component
     ;; interfaces not required in a document are imported here. The name "types"
     ;; with no URL refers to the `types` document in this package.
-    (import "types" (instance $types
-      (export "request" (type (sub resource)))
-      (export "response" (type (sub resource)))
+    (import "types" "pkg:/types/types" (instance $types
+      (type $request (record))
+      (type $response (record))
+      (export "request" (type (eq $request)))
+      (export "response" (type (eq $response)))
     ))
 
     ;; aliases represent `use` from the imported document
     (alias export $types "request" (type $request))
     (alias export $types "response" (type $response))
     (export "handler" (instance
-      (export $request' "request" (type $request))
-      (export $response' "response" (type $response))
-      (export "handle" (func (param (own $request')) (result (own $response'))))
+      (export $request' "request" (type (eq $request)))
+      (export $response' "response" (type (eq $response)))
+      (export "handle" (func (param "request" $request') (result $response')))
     ))
   ))
   ;; corresponds to `wit/proxy.wit`
@@ -1318,27 +1320,29 @@ and its corresponding binary encoding would be:
     (export "proxy" (component
       ;; This world transitively depends on "types" so it's listed as an
       ;; import.
-      (import "types" (instance $types
-        (export "request" (type (sub resource)))
-        (export "response" (type (sub resource)))
+      (import "types" "pkg:/types/types" (instance $types
+        (type $request (record))
+        (type $response (record))
+        (export "request" (type (eq $request)))
+        (export "response" (type (eq $response)))
       ))
       (alias export $types "request" (type $request))
       (alias export $types "response" (type $response))
 
       ;; This is filled in with the contents of what `wasi-logging.backend`
       ;; resolved to
-      (import "console" "URL-for-wasi-logging.backend" (instance
-        ...
+      (import "console" "dep:/foo/bar/baz" (instance
+        ;; ...
       ))
-      (import "origin" (instance
+      (import "origin" "pkg:/handler/handler" (instance
         (export $request' "request" (type (eq $request)))
         (export $response' "response" (type (eq $response)))
-        (export "handle" (func (param (own $request)) (result (own $response))))
+        (export "handle" (func (param "request" $request') (result $response')))
       ))
-      (export "handler" (instance $handler
+      (export "handler" "pkg:/handler/handler" (instance
         (export $request' "request" (type (eq $request)))
         (export $response' "response" (type (eq $response)))
-        (export "handle" (func (param (own $request')) (result (own $response'))))
+        (export "handle" (func (param "request" $request') (result $response')))
       ))
     ))
   ))
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -34,7 +34,7 @@ would correspond to:
 
 ```wasm
 (component
-  (import "host" (insance $host
+  (import "host" (instance $host
     (export "log" (func (param "msg" string)))
   ))
   ;; ...
@@ -157,7 +157,7 @@ reusability. This enables a sort of module system for WIT syntax where files may
 import from one another.
 
 > **Note**: The precise semantics of imports and how everything maps out is
-> still being design. Basic filesystem-based organization works but it's
+> still being designed. Basic filesystem-based organization works but it's
 > intended to extend to URL-based organization in the near future. For example
 > the strings below are intended to integrate into a registry-based workflow as
 > well in addition to looking up files on the filesystem.
@@ -343,7 +343,7 @@ world my-world {
 }
 ```
 
-This is an invalid WIT document due because `my-world` needs to import two
+This is an invalid WIT document because `my-world` needs to import two
 unique interfaces called `shared`. To disambiguate a manual import is required:
 
 ```
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -63,7 +63,7 @@ and [function][functions] definitions. For example:
 
 ```wit
 interface wasi-fs {
-  use { errno } from .types
+  use pkg.types.{errno}
 
   record stat {
     ino: u64,
@@ -137,9 +137,9 @@ function or an interface.
 
 ```wit
 world wasi {
-  import wasi-fs: "wasi-fs"
-  import wasi-random: "wasi-random"
-  import wasi-clock: "wasi-clock"
+  import wasi-fs: wasi.fs
+  import wasi-random: wasi.random
+  import wasi-clock: wasi.clock
   // ...
 
   export command: func(args: list<string>)
@@ -159,8 +159,8 @@ interface out-of-line {
 }
 
 world your-world {
-  import out-of-line: out-of-line
-  // ... is equivalent to ...
+  import out-of-line: self.out-of-line
+  // ... is roughtly equivalent to ...
   import out-of-line2: interface {
     the-function: func()
   }
@@ -177,6 +177,7 @@ an explicitly named world must be chosen:
 
 ```wit
 default world my-world {
+  // ...
 }
 ```
 
@@ -203,20 +204,21 @@ interface types {
 }
 
 interface my-host-functions {
-  use { errno, size } from types
+  use self.types.{errno, size}
 }
 ```
 
-Here the destination of the `from` is resolved to the `types` interface defined
-locally within the file. The interface `types` may come either after or before
-the `use` directive's `interface`. Interfaces linked with `use` are not allowed
-to be cyclic.
+Here the `use` starts with `self` which indicates that something from the
+current document is being used. Then the interface named `types` is listed,
+followed by a list of type names to use from the interface. The interface
+`types` may textually come either after or before the `use` directive's
+`interface`.  Interfaces linked with `use` are not allowed to be cyclic.
 
 Names imported via `use` can be renamed as they're imported as well:
 
 ```wit
 interface my-host-functions {
-  use { errno as my-errno } from types
+  use self.types.{errno as my-errno}
 }
 ```
 
@@ -233,14 +235,14 @@ default interface types {
 
 // host.wit
 interface my-host-functions {
-  use { errno, size } from .types
+  use pkg.types.{errno, size}
 }
 ```
 
-Note the `.` in the `.types` destination of the `from` here, which indicates
-that a sibling document is being imported from. Additionally note the usage of
-`default interface` in the `types.wit` file which simplifies the `from`
-directive of the `use`. WIT documents can also import from any interface defined
+The `pkg` keyword indicates that the `use` statement starts at the package root,
+as opposed to the `self` keyword described above starting in the local document.
+Additionally note the usage of `default interface` in the `types.wit` file which
+simplifies the `use`. WIT documents can also import from any interface defined
 within another document, however:
 
 ```wit
@@ -253,23 +255,23 @@ interface more-types {
 
 // host.wit
 interface my-host-functions {
-  use { another-type } from more-types in .types
+  use pkg.types.more-types.{another-type}
 }
 ```
 
-Here the `more-types in ...` indicates that a specific non-`default` interface
-is being chosen to import from. Documents in a WIT package must be named after a
-[valid identifier][identifiers] and be unique within the package. Documents
-cannot contain cycles between them as well with `use` statements.
+Here `more-types in the `use` path indicates that it's the specific interface
+being referenced. Documents in a WIT package must be named after a [valid
+identifier][identifiers] and be unique within the package. Documents cannot
+contain cycles between them as well with `use` statements.
 
 When importing or exporting an [interface][interfaces] in a [world][worlds]
-the same syntax is used after the `:` as what's after a `from` in a `use`:
+the same syntax is used after the `:` as `use`:
 
 ```wit
 world my-world {
-  import host: host
-  import other-functionality: .sibling-file
-  import more-functionality: specific-interface in .sibling-file
+  import host: self.host
+  import other-functionality: pkg.sibling-file
+  import more-functionality: pkg.sibling-file.specific-interface
 }
 
 interface host {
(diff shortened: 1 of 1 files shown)


alexcrichton:
I realized after opening that the original title didn't even mention `use`, which is one of the primary purposes of this PR, so I've updated it.

design/mvp/WIT.md
```diff
@@ -1,28 +1,518 @@
...
+}
+```
+
+represents an interface called `host` which provides one function, `log`, which
+takes a single `string` argument. If this were imported into a component then it
+would correspond to:
+
+```wasm
+(component
+  (import "host" (insance $host
```
ricochet, line 37: instance

design/mvp/WIT.md
```diff
@@ -1,28 +1,518 @@
...
+
+## WIT File Organization
+[use]: #wit-file-organization
+
+WIT files can be organized into separate files for maintainability and
+reusability. This enables a sort of module system for WIT syntax where files may
+import from one another.
+
+> **Note**: The precise semantics of imports and how everything maps out is
+> still being design. Basic filesystem-based organization works but it's
```
ricochet, line 160: designed*

design/mvp/WIT.md
```diff
@@ -1,28 +1,518 @@
...
+  import foo: interface {
+    use { a-type } from "shared1"
+  }
+  import bar: interface {
+    use { other-type } from "shared2"
+  }
+}
+```
+
+This is an invalid WIT document due because `my-world` needs to import two
```
ricochet, line 346: ~~due~~ because

design/mvp/WIT.md
```diff
@@ -376,42 +845,6 @@ union-cases ::= ty
               | ty ',' union-cases?
 ```
 
-## Item: `resource`
```
ricochet, line 379: Given https://github.com/WebAssembly/component-model/pull/129 is nearly ready, should this section be added back?

ricochet:
A few minor nits and one q, otherwise this looks good to me.

Out of scope and not a blocker to merge: It'll be interesting to see the next round of this with URL's. Something we can (one day) do with components that we can't do with many other dependency resolution technologies is link multiple versions. This MVP reasonably doesn't grapple with that.

design/mvp/WIT.md
```diff
@@ -112,6 +602,7 @@ keyword ::= 'use'
           | 'tuple'
           | 'future'
           | 'stream'
+          | 'default'
```
fibonacci1729, line 605: I think we'll need to add `in` to the list of keywords as well.

fibonacci1729 (approved):

design/mvp/WIT.md
```diff
@@ -1,28 +1,518 @@
...
+```wit
+interface out-of-line {
+  the-function: func()
+}
+
+world your-world {
+  import out-of-line: out-of-line
+  // ... is equivalent to ...
+  import out-of-line2: interface {
+    the-function: func()
```
lukewagner, line 131: When we get to how URLs are produced by Wit, I expect these two will be different (the former will get a URL whereas the latter wouldn't).

design/mvp/WIT.md
```diff
@@ -1,28 +1,518 @@
...
+  type size = u32
+}
+
+interface my-host-functions {
+  use { errno, size } from types
+}
+```
+
+Here the `from` directive of the `use` is not quoted which means that it's
+resolved relative to the WIT file itself. The `interface types` may come either
```
lukewagner, line 181: Not sure why "interface types" is given the backticks?

design/mvp/WIT.md
```diff
@@ -577,9 +1002,122 @@ record bar2 {
...
+would correspond to:
+
+```wasm
+(component
+  (type $wit (instance
+    (type $shared (instance
+      (type $metadata (record (; ... ;)))
+      (export "metadata" (type (eq $metadata)))
+    ))
+    (export $shared "shared" (type $shared))
```
lukewagner, line 1094: I think you'll have to rename one of these `$shared` ids to avoid a conflict.

lukewagner (approved):
On first pass, lgtm, nice job writing this up!

design/mvp/WIT.md
alexcrichton: Ah from reading https://github.com/WebAssembly/component-model/pull/129 I see now that it's updating the `WIT.md` as part of that PR.

design/mvp/WIT.md
```diff
@@ -1,28 +1,518 @@
...
+## WIT Identifiers
+[identifiers]: #wit-identifiers
+
+Identifiers in WIT documents are required to be valid component identifiers,
+meaning that they're "kebab cased". This currently is restricted to ascii
+characters and numbers that are `-` separated.
+
+For more information on this see the [binary format](./Binary.md).
+
+## WIT in Markdown
```
guybedford, line 483: This seems very useful, but I'm not sure it should be _specified_ here? As in, effectively that makes this proposal dependent on a markdown specification? I wonder if there's a way to refer to it more vaguely as a possible feature as opposed to an explicit part of the specification.

alexcrichton:
I talked with Luke a bit more day, and we concluded that the current idea of slurping up everything into one WIT document is probably not actually going to work. This loses filesystem structure when merging into a document and doesn't actually provide a means by which to name everything in the binary representation within a component. An existing bug from my description here is that there's not a great way to name conflicting interfaces when they are placed in the binary representation.

Another downside is that the source of an import is lost. This is significant when a registry is considered, for example, because when you import something from `wasi:random` the fact you imported it from there is lost and no longer meaningful by the time it's merged into one document. From a code-generation perspective this is ok but from a "module ecosystem of wits" this is not ok.

I wanted to basically post this here as a status update. I would like to try to come up with a design to fix this but it will require some thinking about how this all works out.

design/mvp/WIT.md
```diff
@@ -577,9 +1002,122 @@ record bar2 {
...
+    ))
+    (export "host" (type $host))
+
+    (type $the-world (component
+      (import "host" (instance (type $host)))
+      (export "run" (func))
+    ))
+    (export "the-world" (type $the-world))
+  ))
+  (export "wit" (type $wit))
```
peterhuene, line 1048: I'm having trouble understanding the necessity of having the `wit` export.

Would the following not convey the same semantic meaning?

```wit
(component
  (type $host (instance
    (export "log" (func (param "arg" string)))
  ))
  (export "host" (type $host))
  (type $the-world (component
    (import "host" (instance (type $host)))
    (export "run" (func))
  ))
  (export "the-world" (type $the-world))
)
```

Where the root `component` represents the document, which declares an instance type (interface) named `host` and a component type (world) named `the-world`?

Are we expecting potentially more exports than just `wit` from the root component in the future, perhaps?

alexcrichton: Ah this is actually a mistake in my example which was from a misunderstanding I had. Currently the example has this for example:

```wasm
    (type $host (instance
      (export $metadata "metadata" (type $outer-metadata))
      (export "get" (func (result $metadata)))
    ))
    (export $host "host" (type $host))
```

when actually it needs to be 

```wasm
    (type $host (instance
      (export $metadata "metadata" (type $outer-metadata))
      (export "get" (func (result $metadata)))
    ))
    (export $host "host" (instance (type $host)))
```

(note the type-change of `"host"`) 

Doing this requires being in a type context because otherwise an export can't be conjured out of nowhere with a particular type. Essentially the answer is that this is switching to 100% the type idnex space quickly by defining everything within a type. 

In talking with Luke this is also likely to change a bit from this PR as-specified, and is something I'd like to finish fleshing out before merging this. One example is that right now the `"wit"` export container is an `instance` but instead it's going to change to a `component` so imports to the component type model "I'm importing this WIT, aka WASI" and exports are "I defined these things".

alexcrichton:
Ok I've pushed a relatively large update here. After talking with @lukewagner we've settled on something which I think makes sense for all known and predicted use cases. There were a lot of subtle updates here and there so another once-over would be much appreciated!

design/mvp/WIT.md
peterhuene: Ok, that makes sense and was further clarified during discussion on the component model impl weekly meeting. Thanks!

alexcrichton:
Another weekend, more thinking, another Monday, yet more thinking. I've pushed an update to the `use` syntax now to address some concerns I have plus concerns from the last wit-component meeting last Friday:

* The "thing that's being used from" is now syntactically placed first. Previously it was `use { ... } from ...` which isn't amenable to IDE auto-completion, whereas reversing the order should make it easier on IDEs since the set of what-to-autocomplete is narrowing down as you type characters. This is a far-future concern since IDE support will take awhile but seems as good enough a reason as any to not paint the bikeshed that particular shade of colors.

* I've tried to game out what this looks like in the low-level tooling and have opted to completely remove strings from the `use` syntax. Syntactically this changed but semantically you can do the same thing. Previously what was written as `use { a-type } from "wasi:types"` is now written as `use wasi.types.{a-type}`. From a parsing and tooling perspective this clearly identifiers that `wasi` is an identifer that needs to be resolved to a package by some outer context. For example registry tooling would be externally configured to say what the `wasi` identifier points to and that configuration would be communicated to the WIT parsing phase.

design/mvp/WIT.md
```diff
@@ -1,28 +1,567 @@
...
+
+This syntax allows `use` paths starting with any identifier other than `self`
+and `pkg` to create a set of names that the document needs to be resolved
+correctly. These names are considered the dependency packages of the current
+package being parsed. Note that the set of names here are local to this package
+do not need to conflict with dependency names in other packages. This enables
+each package to be resolved entirely separately and, if necessary, the name
+`foo` could mean different things to different packages.
+
+> **Note**: The tooling for and mechanism for precisly how these external names
```
ricochet, line 306: precisely*

design/mvp/WIT.md
```diff
@@ -1,28 +1,567 @@
...
+it at the top-level
+
+```wit
+interface out-of-line {
+  the-function: func()
+}
+
+world your-world {
+  import out-of-line: self.out-of-line
+  // ... is roughtly equivalent to ...
```
ricochet, line 163: roughly

ricochet:
The syntactic change from string to package identifiers makes a ton of sense from an implementation perspective. My initial take was that it loses a little bit of the human readability we were hoping to get out of wit. Additionally, too many years in java gave me an initial knee jerk reaction to seeing `pkg.types.foo` as looking a lot like `com.mycompany.types.foo`. But I got past that :)

The indication of current path via `self` vs `pkg` as the root is an unfamiliar paradigm to me. Something else to consider is a dot prefix to indicate local self, e.g. `use .types.{errno as my-errno}`.

alexcrichton:
Yeah the prior draft here used `.types` for "use from a sibling file called `types.wit`" which was Python-inspired. The current use of `pkg` and `self` is Rust-inspired where Rust has `use crate::...` and `use self::...` for similar constructs where `crate` is the root of the crate and `self` is the current module.

Once strings were removed from the prior draft then there was not distinction between using from a dependency and using from an interface within the same document, both would have looked like `use foo.bar` which is something I wanted to at least initially avoid to prevent confusion if possible. In Rust one of the major things in the 2018 edition, at least from my perspective, was "things make a lot more sense if `use` paths are always anchored at a known set of roots" for example your dependencies or the crate root.

One of the difficulties with a Python-like interface is name resolution within a file itself, for example:

```
interface foo {
    type a = u8
}

interface bar {
    use self.foo.a
}
```

Previously this would have been `use foo.a` but that can possibly be confused with a `foo` dependency as well. If the `use .foo.a` syntax were taken to indicate a sibling `foo.wit` file I'm not sure what would be left for specifying a local intra-file use.

design/mvp/WIT.md
```diff
@@ -1,28 +1,567 @@
...
+  ))
+  (alias export $shared "metadata" (type $metadata))
+  (import "host" (instance $host
+    (export "get" (func (result $metadata)))
+  ))
+)
+```
+
+Here it can be seen that despite the `world` only listing `host` as an import
+the component additionally import a `shared` instance. This is due to the fact
```
guybedford, line 393:
```suggestion
the component additionally imports a `shared` instance. This is due to the fact
```

design/mvp/WIT.md
```diff
@@ -1,28 +1,567 @@
...
+This is an invalid WIT document because `my-world` needs to import two
+unique interfaces called `shared`. To disambiguate a manual import is required:
+
+```
+world my-world {
+  import shared1: pkg.shared1
+  import shared2: pkg.shared2
+
+  import foo: interface {
+    use pkg.shared1.{a-type}
```
guybedford, line 427: Should this be `self.shared1`?

design/mvp/WIT.md
```diff
@@ -1,28 +1,567 @@
...
+  (import "host" (instance $host
+    (export "get" (func (result $metadata)))
+  ))
+)
+```
+
+Here it can be seen that despite the `world` only listing `host` as an import
+the component additionally import a `shared` instance. This is due to the fact
+that the `use { ... } from shared` implicitly requires that `shared` is imported
+to the component as well.
```
guybedford, line 395:
```suggestion
into the component as well.
```

design/mvp/WIT.md
```diff
@@ -1,28 +1,567 @@
...
+
+Note that the name `"shared"` here is derived from the name of the `interface`
```
guybedford, line 397: Ideally could such use-based derivation also follow the same kind of "externalized package management handling" semantics you describe earlier on?

design/mvp/WIT.md
```diff
@@ -1,28 +1,567 @@
...
+  (import "shared" (instance $shared
+    (type $metadata (record (; ... ;)))
+    (export "metadata" (type (eq $metadata)))
+  ))
+  (alias export $shared "metadata" (type $metadata))
+  (import "host" (instance $host
+    (export "get" (func (result $metadata)))
+  ))
+)
+```
```
guybedford, line 390: This generation is interesting indeed. Apologies if I'm repeating discussions here, but could we not restrict that all "use" statements in imported interfaces should reference imported interfaces themselves?

design/mvp/WIT.md
```diff
@@ -577,9 +1049,298 @@ record bar2 {
...
+}
+```
+
+would correspond to:
+
+```wasm
+(component
+  (type (export "host") (component
+    (type $types (instance
+      (export "enum" (type (enum "info" "debug")))
```
guybedford, line 1127:
```suggestion
      (export "level" (type (enum "info" "debug")))
```

design/mvp/WIT.md
```diff
@@ -577,9 +1049,298 @@ record bar2 {
...
+      (alias outer 1 $level (type $level'))
+      (export "log" (func (param "level" $level') (param "msg" string)))
+    ))
+    (export "console" (instance (type $console)))
+  ))
+)
+```
+
+Here the `alias` of `"level"` indicates that the exact type is being used from a
+different interface.
```
guybedford, line 1141: Is it valid to say that while this is the direct representation, if the interface type was fully _internalized_ (so that `"types"` is not exported), a "semantically equivalent" simplification for the specific execution as a compaction operation could be performed into something like:

```wasm
(component
  (type $level (enum "info" "debug"))
  (type $console (instance
    (alias outer 1 $level (type $level'))
    (export "log" (func (param "level" $level') (param "msg" string)))
  ))
  (export "console" (instance (type $console))))
)
```

Or is the whole point that in the component model execution one cannot inline types multiple times, but must always uniquely refer to them?

guybedford: Given the above, could it be worth having a better way of marking interfaces as private / local to benefit from this kind of encapsulation above, where importers of the component type are forced to go through the "public" console interface to reach types, as opposed to having two options (`use host.types.level` versus `use host.console.level`)?

design/mvp/WIT.md
```diff
@@ -577,9 +1049,298 @@ record bar2 {
...
+      (import "console" (instance
+        (export "log" (func (param "arg" string)))
+      ))
+    ))
+    (export "the-world" (type $the-world))
+  ))
+)
+```
+
+Each WIT document in a package becomes a top-level export of the component:
```
guybedford, line 1202: It would be good to define "package" here more clearly. As @ricochet mentions, defining the root is an important part of that, but also which files constitute the public WIT in this conventional manner. For example distinguishing WIT files that indicate dependencies or private files from the public WIT interface. Perhaps it's as simple as something like "WIT tooling is provided a set of public WIT file paths for a component package, where the shared root is treated as the package root, and all files are then combined into their logical `.` separated naming based on that file structure".

design/mvp/WIT.md
```diff
@@ -577,9 +1049,298 @@ record bar2 {
...
+
+// wit/proxy.wit
+default world proxy {
+  import console: wasi-logging.backend
+  import origin: pkg.handler
+  export handler: pkg.handler
+}
+```
+
+and its corresponding binary encoding would be:
```
guybedford, line 1287: In this generation, could one imagine that the generation is passed `createComponentFor(wit/handler.wit, wit/proxy.wit)` as the public interface, with `wit/types.wit` not treated as an explicit exported interface, but rather an internal interfacing, which is then made available indirectly eg via `use wasi.handler.request` etc? This way the first `(type (export "types"))` could be treated as unnecessary.

Secondly, as a simplification could one define `(type sub resource)` as a private shared internal outer type which is aliased within the instance types?

guybedford (commented):
There's a lot to digest here, but it's a fascinating writeup! Hope I'm not rehashing too much in the comments!

design/mvp/WIT.md
guybedford: Alternatively, can this be done with an explicit export pattern in a top-level package WIT somehow?

design/mvp/WIT.md
alexcrichton: I don't think so in that this is referencing the `shared1.wit` file (although it may not be entirely clear here)

design/mvp/WIT.md
alexcrichton: I think that may be required since if an import refers to an item it can't refer to something in the local component as that would create a cyclic relation, so for imports the only choice of where to get their imports from are from other imports.

design/mvp/WIT.md
alexcrichton: The current thinking is that the component you listed is not a valid component due to the restriction that `enum` types appearing in function parameters or results (transitively or directly) must be referenced as an `export` from some instance somewhere. That's what the translation here is trying to capture is that the name is exported somewhere and we're translating where it was defined.

design/mvp/WIT.md
alexcrichton: Right now the thinking is that there's a directory and all `*.wit` files in the directory make-up the package (conventionally), but it's theoretically possible to configure things differnetly. Structurally though packages are a flat list of documents to get `use` between them working.

I'm not sure what you mean by private WIT files, though, because right now I don't think there's any affordance anywhere to a private interface detail since all of WIT is dealing with the interfaces of a component which is something that can't be internalized as it's public by definition.

design/mvp/WIT.md
alexcrichton: My above responses may allude to this, but I think there's a higher level disconnect perhaps in our understandings of WIT/components in that it seems like you're thinking of a situation where private implementation details are a thing but I haven't been considering that a possibility at least myself. 

At least for me it's hard to answer this question since many of the structures here in WIT and plan for WIT encodings are reflexive of the validation rules of compnents themselves, and using semantics like you're describing I think would require changing validation semantics.

design/mvp/WIT.md
guybedford: I read this as that a feature of the disambiguation was that one could use any name here, unrelated to the actual WIT name, along the lines of:

```
world my-world {
  import whatever: pkg.shared1

  import foo: interface {
    use self.whatever.{a-type}
  }
}
```

design/mvp/WIT.md
guybedford: I see, so is the idea that the type must be "public" effectively by being accessible through exports? What if the component that the type existed on were a nested internal component?

design/mvp/WIT.md
guybedford: Consider the example of WASI structuring, where you might have a shared file that defines your "common" records. But you don't necessarily want that shared WIT document to be part of the final API. In this way there is a distinction between code structure and the public "flat listing". Flat listing with private ignores can be one way to capture it.

design/mvp/WIT.md
guybedford: Yeah this is the same topic as the previous comment - files used for code structure versus public interface boundaries.

alexcrichton:
The changes described in this PR should now be fully implemented as of https://github.com/bytecodealliance/wasm-tools/pull/867 and will be released to crates.io shortly.

design/mvp/WIT.md
lukewagner: This is a good question, and a bit subtle.  What's happening with `use` is that you're explicitly reaching *outside* the current `world` to select an interface and then relying on the machinery of `use` to add it to the containing `world`.

If you want to refer to an already imported interface in your *same* world, I *think* (but might be wrong, or this might not be implemented yet, so Alex check me on this) you could just use the identifier of the import directly, so, e.g., you could write:
```wit
world my-world {
  import shared1: pkg.shared1
  import shared2: pkg.shared2

  import foo: interface {
    type a-type = shared1.a-type
  }
  import bar: interface {
    type other-type = shared2.other-type
  }
}
```
If that's right, it might be useful to include both examples?

design/mvp/WIT.md
lukewagner: So perhaps then keep the `resource` section (but proceed with killing `future`/`stream` until they are re-added in the future)?

lukewagner (commented):
Re-reading this, it still looks good, just a few comments/nits/questions from me.  I see there are still some outstanding questions from @guybedford that it'd be good to close out.  Once those are done, perhaps we can have a last call for comments, wait a few days, then merge this and iterate in future PRs/issues like usual.

design/mvp/WIT.md
alexcrichton: It's correct yeah that reaching and importing from a specific named import isn't supported yet. Currently a path of the form `foo.bar` refers to the `foo` dependency and the `bar` document within there, so I think there will be some nuance implementing this.

That being said there's a fair amount of inference about how precisely a world is structured, so it's hopefully unlikely that the desire to do this comes up any time soon. For example unless an interface is being both imported and exported any references to it are unambiguously an import.

design/mvp/WIT.md
alexcrichton: Hm I'm not quite sure I'm following this question (sorry for the delay getting to it). At least in the initial implementation this isn't something that can be left to some future scheme since resolving this is required to encode worlds into wasm.

design/mvp/WIT.md
alexcrichton: Sorry for the delay in getting to this, but I don't fully understand the "this type must be exported" rules myself. @lukewagner has a better idea about this I think (e.g. the when is a referenced resource type valid problem). In that sense I don't know if it's valid or not if it's just exported from some internal component. Also "exported from some internal component" could mean a lot of things I think, so I don't know how to get more precise than that.

design/mvp/WIT.md
alexcrichton: Currently there's no plan to support "private" wit definitions, everything's public and everyone can access everything. I'm not sure myself though if there's much precedence in IDLs for schemes like this which also have a notion of privacy.

design/mvp/WIT.md
alexcrichton: I think this is still present, it's just in a section called `Handles` now

lukewagner: Ah, right, I forgot about that, thanks.

design/mvp/WIT.md
lukewagner: Thinking about how this could be integrated in the future, we'd probably naturally want to allow `foo.bar` (and `shared1.a-type` in the example above), which raises the ambiguity Alex mentions.  Given that inter-document-intra-package references are qualified with `pkg.`, perhaps it makes sense to prefix inter-document-inter-package references as well, e.g. with `ext.` or `dep.`.  Once we have prefixed external references, then I think that frees up intra-document references to optionally drop the `self.` prefix when it is unambiguous (i.e., when there is not shadowing by an interface- or world-local name), keeping `self.` as an option for when there is ambiguity (shadowing).

design/mvp/WIT.md
```diff
@@ -1,28 +1,568 @@
...
+```wit
+interface shared {
+  record metadata {
+    // ...
+  }
+}
+
+world my-world {
+  import host: interface {
+    use self.shared.{metadata})
```
guybedford, line 371:
```suggestion
    use self.shared.{metadata}
```

design/mvp/WIT.md
guybedford: This thread I guess is mainly just about the question of public versus private types. That if a type is never exposed across the component membrane (ie in any way transitively through an export or import), it could be useful to keep it private and unexported.

I understand everything is public though currently, and that seems great for this PR, but it could be worth continuing that discussion.

design/mvp/WIT.md
guybedford: This was very much asked based on the concept of public v private, which is not relevant given everything is public.

guybedford (approved):

design/mvp/WIT.md
```diff
@@ -1,28 +1,568 @@
...
+The `use` statement so far has always started with `self` or `pkg`, but it may
+also start with any valid identifier:
+
+```wit
+interface foo {
+  use package.other-document.{some-type}
+}
+```
+
+This form indicates that the identifier `package` corresponds to some externally
```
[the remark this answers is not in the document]
alexcrichton: FWIW ther `deps/*` folder is more of a temporary convention of sorts as implemented by `wit-component` rather than something intended to be codified here. It's envisioned that this is where registry tooling comes into play and figures things out for you. For example a bindings generation tool would have external configuration about dependency information that wouldn't require a `deps/*` folder.

design/mvp/WIT.md
alexcrichton: Ok I've been trying to think about this a good deal recently, and I'd like to propose an alternative. I actually had a really long comment written up about all the various design with `use` and trying to rationalize that with my historical experience with the Rust module system, but I ended up throwing it away when I realized my conclusion didn't make sense. The crux of it was that, at least in Rust, names aren't inherited between module namespaces. For example in Rust you have:

```rust
type SomeType = u32;

mod foo {
    pub fn foo() -> SomeType { 2 } // ERROR: no name `SomeType` defined
}
```

where each `mod` has its own fresh new namespace. Personally I think this is somewhat important because it prevents names from spilling over everywhere and raising the risk of "action at a distance" where an edit in one location breaks something far away. 

Now a resonable question is what does any of this have to do with worlds here. My thinking is that if we allowed:

```
world my-world {
  import shared: pkg.shared1

  import foo: interface {
    use shared.{a-type}
  }
}
```

then why not _also_ allow something along the lines of 

```
interface foo { type t = u32 }

interface bar {
    use foo.{t} // as-is this PR otherwise requires `self.foo.{t}`
}
```

This feels like it's re-introducing the problem of inheriting namespaces again, however, which I'm worried about enabling. For example if `interface` blocks inherit their document's outer namespace, why not have documents inehrit their package's outer namespace? That gets confusing though because then you wouldn't know at-a-glance what `use foo.bar.{t}` refers to -- is `foo` a document in your package, a dependency, or an interface `foo` in this document?

So one idea I had to perhaps solve this, instead of the world I wrote above consider this instead:

```
world my-world {
  import shared: pkg.shared1

  import foo: interface {
    use world.shared.{a-type}
  }
}
```

basically all paths in `use` start today with either `pkg`, `self`, or an identifier which is a dependency. An addition, for worlds, would be a `world` keyword which refers to the current world. Or we could add something like `import.foo.bar` or `export.foo.bar` as well since imports/exports have disjoint namespaces.

That would mean no change from what's implemented today, but planning to use `import` and `export` prefixes on imports-from-worlds in the future. How does that sound?

lukewagner: Having explicit `import`/`export` prefix be the plan of record makes sense too, so I'm fine just going with that, which I guess resolves this issue.

lukewagner:
Ok, looks like there aren't any open issues and things have been quiet; I guess I'll merge at the end of the week.

alexcrichton:
Ok I'm going to go ahead and merge this. I'm going to follow up with a few more words about types-in-worlds but that mostly only affects the wasm-encoding. Thanks again everyone for the discussion here!

PR #155, opened 2023-01-24
add optional export type ascription

### LINKED ISSUE #151: Cannot faithfully represent type aliases in exported instances from WIT
Given a WIT document of the form:

```
interface foo {
  type a = u8
  type b = a
  c: func(a: a) -> b
}

default world the-world {
  export foo: self.foo
}
```

when this is actually transformed into a component that implements `the-world` the current translation of `wasm-tools` looks like:

```wasm
(component
  (core module (;0;) (;... core wasm module here ...;) )
  (core instance (;0;) (instantiate 0))
  (type (;0;) u8)
  (type (;1;) (func (param "a" 0) (result 0)))
  (alias core export 0 "foo#c" (core func (;0;)))
  (func (;0;) (type 1) (canon lift (core func 0)))
  (instance (;0;)
    (export "a" (type 0))
    (export "b" (type 0))
    (export "c" (func 0))
  )
  (export (;1;) "foo" (instance 0))
)
```

The problem with this format, however, is that the type alias cannot faithfully be represented in this format. In the exported instance it has

```
    (export "a" (type 0))
    (export "b" (type 0))
```

which loses the information that "b" is a reexport of "a". This is due to the fact that the export introduced in "create an instance from this bag of exports" doesn't get its own index for "b" to refer to.

This will furthermore be a problem with validation of resources/records/etc where it's not possible to execute the `canon lift` with a parameter that has a type which requires a name. The type shouldn't be exported from the root of the component, only a sub-instance, so there's no way to introduce the type index before the `canon lift` happens.

I believe @lukewagner has an idea of how to fix this but I wanted to open an issue since I'm about to write a FIXME in code that needs to reference this problem.

lukewagner:
This PR resolves #151 by adding the optional ability to attach an explicit type to an `export` definition.  This allows a function to be initially defined using a private/non-exported type and, when the function is later exported (possibly in a different instance), for the public/exported type to be swapped in.

This PR ended up being a bit different than what I proposed in #151:
* Since it's optional, the `externdesc` needs to go at end of the `export` to avoid text-format ambiguity.
* Since `externdesc` is really optional (in both text and binary), we can't use it to bind the identifier (via `bind-id(<externdesc>)`, as we do with imports); we need to keep the `<id>?` where it is now.  (Otherwise, if you wanted an identifier, you'd be forced to add an explicit type to the binary format and identifiers are supposed to be a purely text-format concern that doesn't perturb the binary format.)
* The inline "bag-of-exports" form for defining instances isn't extended.  It's possible, and we can always add it backwards-compatibly in the future, but it seemed like a rather non-trivial addition and amount of work that's ultimately just convenience for what you can do with a tiny helper component (that imports and re-exports).

With these choices, the PR and size of change is quite small.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -324,10 +324,10 @@ flags are set.
 (See [Import and Export Definitions](Explainer.md#import-and-export-definitions)
 in the explainer.)
 ```
-import     ::= en:<externname> ed:<externdesc> => (import en ed)
-export     ::= en:<externname> si:<sortidx>    => (export en si)
-externname ::= n:<name> u?:<URL>?              => n u?
-URL        ::= b*:vec(byte)                    => char(b)*, if char(b)* parses as a URL
+import     ::= en:<externname> ed:<externdesc>                => (import en ed)
+export     ::= en:<externname> si:<sortidx> ed?:<externdesc>? => (export en si ed?)
+externname ::= n:<name> u?:<URL>?                             => n u?
+URL        ::= b*:vec(byte)                                   => char(b)*, if char(b)* parses as a URL
 ```
 Notes:
 * All exports (of all `sort`s) introduce a new index that aliases the exported
@@ -338,7 +338,10 @@ Notes:
   parser] with `char(b)*` as *input*, no optional parameters and non-fatal
   validation errors (which coincides with definition of `URL` in JS and `rust-url`).
 * Validation requires any exported `sortidx` to have a valid `externdesc`
-  (which disallows core sorts other than `core module`).
+  (which disallows core sorts other than `core module`). When the optional
+  `externdesc` immediate is present, validation requires it to be equal to
+  the inferred `externdesc` of the `sortidx` (where equality judges a type and
+  the `typeidx` of an export of that type (via `eq` or `sub`) equivalent).
 * The `name` fields of `externname` must be unique among imports and exports,
   respectively. The `URL` fields of `externname` (that are present) must
   independently unique among imports and exports, respectively.
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -1299,7 +1299,7 @@ of core linear memory.
 Lastly, imports and exports are defined as:
 ```
 import     ::= (import <externname> bind-id(<externdesc>))
-export     ::= (export <id>? <externname> <sortidx>)
+export     ::= (export <id>? <externname> <sortidx> <externdesc>?)
 externname ::= <name> <URL>?
 ```
 Both import and export definitions append a new element to the index space of
@@ -1313,7 +1313,11 @@ exported).
 Validation of `export` requires that all transitive uses of resource types in
 the types of exported functions or values refer to resources that were either
 imported or exported (concretely, via the type index introduced by an `import`
-or `export`). For example, in the following component:
+or `export`). The optional `<externdesc>?` in `export` can be used to ascribe
+an equivalent-but-different type to an exported definition, allowing a private
+type definition to be replaced with a public (exported) type definition.
+
+For example, in the following component:
 ```wasm
 (component
   (import "R1" (type $R1 (sub resource)))
@@ -1324,6 +1328,7 @@ or `export`). For example, in the following component:
   (func $f2' (result (own $R2')) (canon lift ...))
   (export "f1" (func $f1))
   ;; (export "f2" (func $f2)) -- invalid
+  (export "f2" (func $f2) (func (result (own $R2'))))
   (export "f2" (func $f2'))
 )
 ```


alexcrichton:
Thinking a bit more about this, is the type ascription still necessary? I opened #151 because the current bag-of-exports is what's used right now during copmonent generation and that's what doesn't work, but if bag-of-exports isn't being updated here that'll need to change to instantiation of a nested component. Given that at least my personal motivation for #151 I believe is solved and wouldn't require the addition of type ascription

lukewagner:
Without type ascription, the nested component instance wouldn't be able to assign a new type to the original function definition (using the type export of the nested component instance), so I think #151 would still be a problem.

lukewagner:
That's a really good question and indeed the situation feels fuzzy with type aliases of value types.  For starters, things get a bit more clear though when working through an example with resource types.  So when validating this component:
```wasm
(component
  (type $R (resource (rep i32)))
  (func $f (result (own $R)) (canon lift ...))
  (component $Helper
    (import "R" (type $R1 (sub resource)))
    (import "f" (func $f1 (result (own $R1))))
    (export "R" (type $R1))
    (export "f" (func $f1))
  )
  (instance $helper (instantiate $Helper (with "R" (type $R)) (with "f" (func $f))))
  (export "some-interface" (instance $helper))
)
```
the validation rule for `instantiate` substitutes the local resource type `$R` for all occurrences of the import `R` in `$Helper`.  Thus, the resulting type of `$helper` is effectively:
```wasm
(instance
  (alias outer 0 $R (type $R1))
  (export "R" (type (eq $R1)))
  (export "f" (func (result (own $R1))))
)
```
Here, the export `f` is still referring to the outer (non-exported) `$R` via alias, not the exported `R` we need it to.  Thus, when we try to export `$helper`, validation will give an error because `f` is referring to a privately-defined type.  To fix this, we need to use type ascription on the export of `f`:
```wasm
  (component $Helper
    (import "R'" (type $R1 (sub resource)))
    (import "f'" (func $f1 (result (own $R1))))
    (export $R2 "R" (type $R1))
    (export "f" (func $f1) (func (result (own $R2))))
  )
```
Based on this understanding, it's easier to see the problem if you change the example from using resource types to aliased value types: without type ascription in `$Helper`, the type of `f` won't end up referring to the exported type name; it'll be referring to something internal.

alexcrichton:
That makes sense, and honestly I haven't really thought much about resources specifically. I would natively expect the intermediate module you listed to be:


```wasm
  (component $Helper
    (import "R" (type $R1 (sub resource)))
    (export $R "R" (type $R1))
    (import "f" (func $f1 (result (own $R))))
    (export "f" (func $f1))
  )
```

which would solve the problem, I believe. I've been thinking this way because no types currently require an import like a resource does, but I also thought that resources here would be required to be named so the `$Helper` you listed above wouldn't validate because the resource is only named through the import, not the export. There's a lot I'm missing about resources though...

alexcrichton:
> but unfortunately imports cannot refer to the type-indices produced by exports

Oh dear, I was slow in getting to digesting this again and this is not what I expected. I [just implemented types in worlds](https://github.com/bytecodealliance/wasm-tools/pull/901) and explicitly had to write a fair amount to support what I thought was going to need to work along the lines of:

```
world foo {
    record r {}
    import foo: func() -> t
}
```

but if imports can't refer to exports then this can't be supported? I think this means that imported functions directly into a world can't refer to any nontrivial types?


---

Also I meant that for this wasm:

```wasm
  (component $Helper
    (import "R" (type $R1 (sub resource)))
    (import "f" (func $f1 (result (own $R1))))
    (export "R" (type $R1))
    (export "f" (func $f1))
  )
```

I thought that would fail to validate. The import of `"f"` refers to `$R1` which is not an exported name, just an imported name. I thought the exported name was going to be required.

Overall it sounds like I'm even more strongly disconnected from how you envisioned this all working than I originally thought...

lukewagner:
> but if imports can't refer to exports then this can't be supported?

I think the way to capture all the name information to be able to reproduce the original Wit is:
```wasm
(type (component
  (type $r (record))
  (import "r" (type $r' (eq $r)))
  (import "foo" (func (result $r'))
))
```
Note: the component validation rules don't require this; the import could just refer directly to `$r` and the `$r'` import could be dropped.

-----------------

> I thought that would fail to validate. The import of "f" refers to $R1 which is not an exported name, just an imported name.

In the first draft of the validation rules, that's the stricter rule we started with.  Then, we discussed and I filed #145 to loosen these rules to say that that exported functions can refer to export *or* imported types, since imported types are also knowable externally.  And concretely the motivation is, e.g.: if I import `wasi:http/types` so that I can export an HTTP `handle` function, it's really not necessary for me to re-export `wasi:http/types` since my clients can also import `wasi:http/types`.

alexcrichton:
Hm ok. Sorry but I'm basically totally lost at this point. It sounds like this type ascription needs to be added, but I don't know where it would get integrated into the component creation process or translation from a core wasm to a component. I also seemingly implemented types-in-worlds entirely differently than what was to be expected, and decoding of types is also very different from what I expected if types are going to be imported instead of exported. 

If you get a chance I'd really appreciate a review on what's implemented and suggestions for how to change things. I'll need to set aside a good chunk of time to rethink how to implement all of this.

alexcrichton (approved):
Ok I talked with Luke today and things are straight in my head now at least. In short:

* Types-in-worlds are going to always be modeled as type imports, both in the wasm-encoding of wit and the final component itself.
* Export-an-instance will use type-ascription to "re-type" imports within a component. This means instance exports will always synthesize a "helper component" and then immediately instantiate it to reexport its imports. The imports work because "this must be named" is only a restriction for exports.


In light of that this all looks good to me. Thanks again for taking the time @lukewagner to help me through the questions I had!

PR #163, opened 2023-02-11
Move definition of `Flags`

Kylebrown9:
This makes the ordering of definitions and match branches more consistent.


lukewagner:
lgtm, thanks!

PR #164, opened 2023-02-14
Tighten validation to require uniqueness between import and export kebab-names

lukewagner:
Recently, I've seen a few cases where the possibility of overlap between the kebab-names of imports and exports has led to additional complexity that seems unnecessary given that the whole point of kebab-names is to provide a nice local developer-friendly names that can be renamed in cases of collision without impacting meaning (which is where the `(id <URL>)` subfield come in instead).  Thus, this PR suggests tightening the validation rules for component-level `externname`s to rule out these cases of overlap.  E.g., with this change, the names of imports and exports in a Wit `world` can share a single namespace.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -343,9 +343,12 @@ Notes:
   (which disallows core sorts other than `core module`). When the optional
   `externdesc` immediate is present, validation requires it to be a supertype
   of the inferred `externdesc` of the `sortidx`.
-* The `name` fields of `externname` must be unique among imports and exports,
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
This definitely simplifies the bindings generation side of things, but are there use cases that this might otherwise affect such as "I import and export $FOO" where slightly different names now need to be assigned? Those cases can always continue to work with slight renamings which doesn't sound like the end of the world to me, but I haven't thought too far ahead to those cases myself.

lukewagner:
Ok, async feedback seems positive and no issues raised, so merging.

PR #170, opened 2023-02-25
format: Clarify what `nil` represents

### LINKED ISSUE #152: Missing nonterminal definitions in the WIT EBNF
Hi,
whilst reading WIT's EBNF I noticed multiple missing nonterminal definitions, namely: `nil`, `strlit`, `id`

No, the [Identifier](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#identifiers) section of the markdown does not constitute a nonterminal nor can act as one.

If there's reasoning behind any, I would love to know.
I'm not an expert of any kind in (E)BNF nor do I have experience writing an (E)BNF, so do feel free to correct me if I'm wrong.

Thank you

r-muhairi:
This change replaces nil with the Epsilon symbol and adds a note further clarifying what it is, in this case a empty string

Closes #152

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -683,6 +683,8 @@ sequence of items and functions.
 
 Specifically interfaces have the structure:
 
+> **Note**: The symbol `ε`, also known as Epsilon, denotes an empty string.
+
 ```ebnf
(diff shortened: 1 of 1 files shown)


lukewagner:
lgtm, thanks!

Issue #169, opened 2023-02-24
Proposal: Union of Worlds


Mossaka:
I want to raise this issue to propose a new syntax for the union of Worlds. The primary motivation for this proposal is that we want to be able to form a more compresive world by unionizing / combining multiple worlds together to form a bigger one. Concretly, this is motivated by the [`wasi-cloud-core`](https://github.com/WebAssembly/WASI/issues/520) proposal, which tries to define a new World that includes all the wasi-cloud worlds as to provide a feature set that allows a serverless / edge function to do state management or message exchanging. 

## Proposal

This proposal adds a new syntax `include <world-path>` in World definition. Below are a few examples/sketches showing how the `include` syntax would work. 

### Union of two Worlds

```fsharp
// worlds.wit
world my-world-1 {
    import a: self.a
    import b: self.b
    export c: self.c
}

world my-world-2 {
    import foo: self.foo
    import bar: self.bar
    export baz: self.baz
}

world union-my-world {
     include self.my-world-1
     include self.my-world-2
}

is equivalent to

world union-my-world {
    import a: self.a
    import b: self.b
    export c: self.c
    import foo: self.foo
    import bar: self.bar
    export baz: self.baz
}
```

### Include a World that has package and inline imports

> Notice that the inline export has been "copied" over to the World that includes this World. 

```fsharp
// b.wit
interface b { ... }

// a.wit
interface a { ... }

world my-world-1 {
    import a: self.a 
    import b: pkg.b
    import c: io.c // external package
    export d: interface exp { ... }
}

// union.wit

world union-my-world-1 {
    include pkg.my-world-1
}

is equivalent to

world union-my-world-1 {
    import a: pkg.a
    import b: pkg.b
    import c: io.c
    export d: interface exp { ... }
}
```
### Name Conflicts

This is a more challenging example where the two Worlds being included have name conflicts. The solution to this is to use `with` syntax and aliasing to avoid conflicts of names. The `with` syntax is only used to avoid name conflicts for those names that are in conflict with another World. The names that are not in conflicts should stay the same and need not to be specified in the `with` scope.

```fsharp
// my-world-1.wit
world my-world-1 {
    import a: self.a
    import b: self.b
    export d: self.d
}

// my-world-2.wit
world my-world-2 {
    import a: self.a
    import b: self.b
    export c: self.c
}

// union.wit
world union-my-world {
    include pkg.my-world-1 with { a as a1, b as b1 }
    include pkg.my-world-2
}

[...]

PR #174, opened 2023-03-03
Add an `include` keyword

Mossaka:
As proposed by #169 , this PR adds an `include` keyword to the WIT.md. It also sorts the keyword list alphabetically.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -954,7 +954,7 @@ include-item ::= 'include' use-path
                | 'include' use-path 'with' '{' include-names-list '}'
 
 include-names-list ::= include-names-item
-                 | include-names-item ',' include-names-item?
+                     | include-names-list ',' include-names-item
 
 include-names-item ::= id 'as' id
 ```
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -287,7 +287,7 @@ The `include` statement also works with [WIT package](#wit-packages-and-use) def
 interface b { ... }
 
 // a.wit
-interface a {}
+interface a { ... }
 
 world my-world-a {
     import a
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -338,10 +338,7 @@ world union-my-world-b {
 
 ### Name Conflicts and `with`
 
-When two or more included Worlds have the same name for an import or export, the name is considered to be in conflict. The conflict needs to be explicitly resolved by the world author using the `with` keyword. `with` allows the world author to rename the import or export to a different name.
-
-Notice that when import or export names are IDs and since IDs are unique, there is no need to resolve name conflicts. Thus the `with` syntax is invalid when used with `include <ID>`. Only when import or export names are kebab names, name conflicts need to be resolved.
-
+When two or more included Worlds have the same name for an import or export that does *not* have an ID, automatic de-duplication cannot be used (because the two same-named imports/exports might have different meanings in the different worlds) and thus the conflict has to be resolved manually using the `with` keyword:
 The following example shows how to resolve name conflicts where `union-my-world-a` and `union-my-world-b` are equivalent:
 
 ```wit
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -283,10 +283,9 @@ world union-my-world {
 The `include` statement also works with [WIT package](#wit-packages-and-use) defined below with the same semantics. For example, the following World `union-my-world-a` is equivalent to `union-my-world-b`:
 
 ```wit
-// b.wit
-interface b { ... }
+package local:demo
 
-// a.wit
+interface b { ... }
 interface a { ... }
 
 world my-world-a {
@@ -295,9 +294,6 @@ world my-world-a {
     import wasi:io/c
     export d: interface { ... }
 }
-
-// union.wit
-package local:demo
 
 world union-my-world-a {
     include my-world-a
@@ -344,7 +340,7 @@ world union-my-world-b {
 
 When two or more included Worlds have the same name for an import or export, the name is considered to be in conflict. The conflict needs to be explicitly resolved by the world author using the `with` keyword. `with` allows the world author to rename the import or export to a different name.
 
-Notice that when import or export names are IDs and since IDs are unique, there is no need to resolve name conflicts. Thus the `with` syntax is a no-op in this case. Only when import or export names are kebab names, name conflicts need to be resolved.
+Notice that when import or export names are IDs and since IDs are unique, there is no need to resolve name conflicts. Thus the `with` syntax is invalid when used with `include <ID>`. Only when import or export names are kebab names, name conflicts need to be resolved.
 
 The following example shows how to resolve name conflicts where `union-my-world-a` and `union-my-world-b` are equivalent:
 
@@ -365,9 +361,9 @@ world union-my-world-b {
 }
 ```
 
-The following example shows that `with` is a no-op when the import or export name is an ID:
+The following example shows an invalid example that `with` is used when the import or export name is an ID:
 
-```wit
+```wi
 package local:demo
 
 world my-world-a {
@@ -383,11 +379,6 @@ world my-world-b {
 world union-my-world-a {
     include my-world-a with { a1 as a3 }
     include my-world-b with { a1 as a2 }
-}
-
-world union-my-world-a {    
-  import a1
-  import b1
 }
 ```
 
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -358,26 +358,21 @@ world union-my-world-b {
 }
 ```
 
-The following example shows an invalid example that `with` is used when the import or export name is an ID:
-
-```wi
+`with` cannot be used to rename IDs, however, so the following world would be invalid:
+```wit
 package local:demo
 
-world my-world-a {
-    import a1
-    import b1
+interface a {
+    foo: func()
 }
 
-world my-world-b {
-    import a1
-    import b1
+world world-using-a {
+    import a
 }
 
-world union-my-world-a {
-    include my-world-a with { a1 as a3 }
-    include my-world-b with { a1 as a2 }
+world invalid-union-world {
+    include my-using-a with { a as b }  // invalid: 'a', which is short for 'local:demo/a', is an ID
 }
-```
 
 ### A Note on SubTyping
 
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -312,32 +312,9 @@ world union-my-world-b {
 }
 ```
 
-### Name Conflicts
+### De-duplication of IDs
 
-When two or more included Worlds have the same name for an import or export, the name is considered to be in conflict. The conflict needs to be explicitly resolved by the world author using the `with` keyword. `with` allows the world author to rename the import or export to a different name.
-
-The following example shows how to resolve name conflicts where `union-my-world-a` and `union-my-world-b` are equivalent:
-
-```wit
-package local:demo
-
-world world-one { import a: func() }
-world world-two { import a: func() }
-
-world union-my-world-a { 
-    include foo
-    include bar with { a as b }
-}
-
-world union-my-world-b {
-  import a: func()
-  import b: func()
-}
-```
-
-### De-duplication
-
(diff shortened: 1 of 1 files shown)


alexcrichton (commented):
Could you expand a bit about the semantic meaning of the `include` statement? For example when `with` is required, what `with` does, when `include` might fail, etc.

Mossaka:
> Could you expand a bit about the semantic meaning of the `include` statement? For example when `with` is required, what `with` does, when `include` might fail, etc.

Good suggestion. I have thought about that but then I saw the section I've modified is for Lexical Structure. Is there a dedicated section for semantic structure of WIT? If not, I can expand the the semantic meaning in the part where I modified. @alexcrichton

alexcrichton:
The top of the document is probably the best place for that right now, probably around the area that `world`s are explained.

alexcrichton (commented):
Looking good, but I might recommend trying to run some examples through `wit-parser` to sort out `self`/`pkg` paths since right now they're relatively imprecise and can lead to some ambiguities in the examples I think. I realize that `include` isn't implemented yet but I think most other pieces should be able to be fleshed out to make sure they're valid.

As I think about this as well though the "de-duplication" section I think will actually need to be required for the first implementation. It's not valid to import the same interface twice under two names, so if the same interface is imported into two worlds which are union'd then it's required that a deduplication happens rather than emitting two separate imports (e.g. resolving with `with` isn't valid)

Mossaka:
> so if the same interface is imported into two worlds which are union'd then it's required that a deduplication happens rather than emitting two separate imports (e.g. resolving with with isn't valid)

That's a really good point. Transient deps are always hard haha. I will update the doc to include de-dep semantics.

design/mvp/WIT.md
```diff
@@ -185,6 +185,194 @@ default world my-world {
...
+world union-my-world-3 {
+    include pkg.my-world-1
+    include pkg.my-world-2
+}
+
+world union-my-world-4 {
+    import a1: pkg.a
+    import b1: pkg.b
+    export c1: pkg.c
+}
```
alexcrichton, line 330: This I think runs a risk of being surprising because `my-world-2` is effectively ignored here. Could this perhaps start out with a rule that if two interfaces are imported they must be imported under the same name? That means that the union-world above would be an error by default, but it could be fixed with `with` to do renaming to make sure the names line up.

Mossaka: > if two interfaces are imported they must be imported under the same name?

that seems reasonable to me.

design/mvp/WIT.md
```diff
@@ -185,6 +185,194 @@ default world my-world {
...
+    export c: pkg.d
+}
+
+```
+
+### A Note on SubTyping
+
+As of now, host bindings are required to implicitly interpret all exports as optional to make a component targeting an included World a subtype of the union World. This [comment](https://github.com/WebAssembly/component-model/issues/169#issuecomment-1446776193) describes the reasoning behind this decision.
+
+In the future, when `optional` export is supported, the world author may explicitly mark exports as optional to make a component targeting an included World a subtype of the union World.
```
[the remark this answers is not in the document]
Mossaka: Oh I had a different interpretation of Luke's comment. I thought the goal is that we want to produce subtyped worlds and a solution to that is described by Luke as "In the short term, I expect we could get away with simply defining include to take the union of exports (just like imports) and having host bindings implicitly interpret all exports as optional"

@lukewagner please correct me if this is wrong.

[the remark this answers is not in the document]
alexcrichton: Currently there's no support for optional exports, nor is there any planned in the near future at this time. If the goal of `include` is to produce subtyped worlds then this will need to be put on hold until optional exports are considered.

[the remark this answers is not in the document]
Mossaka: @alexcrichton ACK, I am okay with proceeding without subtyped worlds

design/mvp/WIT.md
Mossaka: I have updated the doc to address your comment.

design/mvp/WIT.md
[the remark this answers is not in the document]
Mossaka: I have updated the doc to address your comment.

Mossaka:
@alexcrichton thanks for reviewing. Could you please take another look?

alexcrichton (approved):
Looks good to me, thanks!

design/mvp/WIT.md
```diff
@@ -185,6 +185,216 @@ default world my-world {
...
+    import a: self.a 
+    import b: pkg.b
+    import c: io.c // external package
+    export d: interface exp { ... }
+}
+
+// union.wit
+
+world union-my-world-1 {
+    include pkg.my-world-1
```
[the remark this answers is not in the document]
Mossaka: Yes that's right. Thanks!

design/mvp/WIT.md
```diff
@@ -185,6 +185,216 @@ default world my-world {
...
+```
+
+The `include` statement also works with [WIT package](#wit-packages-and-use) defined below with the same semantics. For example, the following World `union-my-world-1` is equivalent to `union-my-world-2`:
+
+```wit
+// b.wit
+interface b { ... }
+
+// a.wit
+interface a { ... }
```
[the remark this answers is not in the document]
Mossaka: Fixed it. Thank you!

Mossaka:
I will rebase once #198 is merged in.

Mossaka:
Okay, I think this PR is ready to review @guybedford

design/mvp/WIT.md
```diff
@@ -778,6 +912,25 @@ from the root of a component and used within functions imported and exported.
 The `interface` item here additionally defines the grammar for IDs used to refer
 to `interface` items.
 
+## Item: `include`
+
+A `include` statement enables union the current world with another world. The structure of an `include` statement is:
```
guybedford, line 917:
```suggestion
A `include` statement enables the union of the current world with another world. The structure of an `include` statement is:
```

design/mvp/WIT.md
```diff
@@ -778,6 +912,25 @@ from the root of a component and used within functions imported and exported.
...
+
+```wit
+include pkg.my-world-1 with { a as a1, b as b1 }
+include self.my-world-2
```
guybedford, line 921: Note these are still using the dated syntax.

design/mvp/WIT.md
```diff
@@ -236,6 +236,139 @@ Kebab names cannot overlap and must be unique, even between imports and exports.
...
+    include bar with { a as b }
+}
+
+world union-my-world-b {
+  import a: func()
+  import b: func()
+}
+```
+
+### De-duplication
```
guybedford, line 338: Perhaps this should go before the name conflicts section? It would also be worth clarifying how deduplication and name conflicts can be both supported. For example, can `with` syntax be used in cases where deduplication would otherwise have occured? I'd assume so but it would be good to clarify this.

guybedford (commented):
It would be great to see this moving forward, thanks for the update.

design/mvp/WIT.md
Mossaka: Ah missed it. Thanks!

design/mvp/WIT.md
Mossaka: I should clarify. Deduplication only applies to IDs, and `with` syntax only applies to kebab-names.

design/mvp/WIT.md
```diff
@@ -236,6 +236,167 @@ Kebab names cannot overlap and must be unique, even between imports and exports.
...
+    include bar with { a as b }
+}
+
+world union-my-world-b {
+  import a: func()
+  import b: func()
+}
+```
+
+The following example shows that `with` is a no-op when the import or export name is an ID:
```
guybedford, line 368: Rather than a no-op can we just make this throw - that is, it is an error to refer to ids at all in this process? Also helps ensure any future compat concerns.

guybedford (approved):
LGTM

danbugs (approved):
Awesome work, Joe!

design/mvp/WIT.md
lukewagner: +1

design/mvp/WIT.md
```diff
@@ -236,6 +236,167 @@ Kebab names cannot overlap and must be unique, even between imports and exports.
...
+```
+
+The `include` statement also works with [WIT package](#wit-packages-and-use) defined below with the same semantics. For example, the following World `union-my-world-a` is equivalent to `union-my-world-b`:
+
+```wit
+// b.wit
+interface b { ... }
+
+// a.wit
+interface a {}
```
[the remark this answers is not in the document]
lukewagner: Also, a suggestion: with the new Wit syntax, the filenames no longer matter, so it's just as well to remove the "// a.wit" comments since they're not relevant

design/mvp/WIT.md
```diff
@@ -778,6 +940,25 @@ from the root of a component and used within functions imported and exported.
...
+include wasi:io/my-world-1 with { a as a1, b as b1 }
+include my-world-2
+```
+
+```ebnf
+include-item ::= 'include' use-path
+               | 'include' use-path 'with' '{' include-names-list '}'
+
+include-names-list ::= include-names-item
+                 | include-names-item ',' include-names-item?
```
lukewagner, line 957:
```suggestion
                     | include-names-item ',' include-names-item?
```

design/mvp/WIT.md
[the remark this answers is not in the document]
Mossaka: Done

design/mvp/WIT.md
Mossaka: Okay, made the `include <ID> with { ... }` case invalid.

Mossaka: I am happy to send a PR to wasm-tools to adopt this decision.

design/mvp/WIT.md
```diff
@@ -236,6 +236,158 @@ Kebab names cannot overlap and must be unique, even between imports and exports.
...
+    import b1
+}
+```
+
+### Name Conflicts and `with`
+
+When two or more included Worlds have the same name for an import or export, the name is considered to be in conflict. The conflict needs to be explicitly resolved by the world author using the `with` keyword. `with` allows the world author to rename the import or export to a different name.
+
+Notice that when import or export names are IDs and since IDs are unique, there is no need to resolve name conflicts. Thus the `with` syntax is invalid when used with `include <ID>`. Only when import or export names are kebab names, name conflicts need to be resolved.
+
```
lukewagner, line 344:
```suggestion
When two or more included Worlds have the same name for an import or export that does *not* have an ID, de-duplication cannot be used (because the two same-named imports/exports might have different meanings in the different worlds) and thus the conflict has to be resolved using the `with` keyword:
```

design/mvp/WIT.md
```diff
@@ -236,6 +236,158 @@ Kebab names cannot overlap and must be unique, even between imports and exports.
...
+world my-world-b {
+    import a1
+    import b1
+}
+
+world union-my-world-a {
+    include my-world-a with { a1 as a3 }
+    include my-world-b with { a1 as a2 }
+}
+```
```
lukewagner, line 383:
```suggestion
`with` cannot be used to rename IDs, however, so the following world would be invalid:
```wit
package local:demo

interface a {
  foo: func()
}

world world-using-a {
    import a
}

world invalid-union-world {
    include my-using-a with { a as b }  // invalid: 'a', which is short for 'local:demo/a', is an ID
}
```
```

lukewagner (commented):
Thanks for the updates, and great work.  Regarding the most recent change, which says that you can't rename an ID, unfortunately we can't just rule it out syntactically, because `include x with {a as b }` may or may not be allowed depending on `x`.  Here's a suggested rewording, but feel free to reject and do it your own way.

design/mvp/WIT.md
lukewagner: Thanks!

Mossaka:
Thank you, @lukewagner !! I have committed your suggestions.

lukewagner (approved):
Thanks Joe, overall lgtm, thanks for spearheading this!  I'm thinking I'll merge early next week if there aren't any more comments from anyone.

Mossaka:
Thanks everyone for reviewing this!! This meant a lot to me as it marks the first contribution I've contributed to a Wasm proposal. Cheers 🙌 🥂

PR #188, opened 2023-04-22
Add missing paren in explainer example

rylev:
There is a missing paren in the alias example in the explainer. This adds that paren and adjusts the comments so that they remain on the same level.


lukewagner (approved):
Thanks!

PR #192, opened 2023-04-27
Initial implementation of resource type validation

pl-semiotics:
This adds a relatively direct OCaml implementation of the current formalisation (essentially that given in #101), focusing on the rules around how resource types are treated as creating quantified component types.

It is designed to layer cleanly around the Core WebAssembly reference interpreter, but presently requires [minor changes](https://github.com/pl-semiotics/spec/commit/7008d7dfea9f8298f385ed5e3c9857e5fcf02a77) to expose certain parsing-related information and entrypoints.

### DIFF
--- a/interpreter/test/resource/resource-simple.wast
+++ b/interpreter/test/resource/resource-simple.wast
@@ -1,0 +1,271 @@
+(assert_invalid
+  (component
+    (core module $cm
+      (func (export "one") (param i32) (result i32) (i32.const 1)))
+    (core instance $ci (instantiate $cm))
+    (type $t (resource (rep i32)))
+    (export $et "t" (type $t)))
+  "Cannot export type containing bare resource type")
+
+(assert_invalid
+  (component
+    (core module $cm
+      (func (export "one") (param i32) (result i32) (i32.const 1)))
+    (core instance $ci (instantiate $cm))
+    (type $t (resource (rep i32)))
+    (export $et "t" (type $t) (type (sub resource)))
+    (alias core export $ci "one" (core func $cf1))
+    (canon lift $cf1 (func $f1 (param (own $t)) (result (own $t))))
+    (export "one" (func $f1)))
+  "Cannot export type containing bare resource type")
+
+(assert_invalid
+  (component $C
+    (type $t (resource (rep i32)))
+    (component
+      (alias outer $C $t (type))))
+  "Cannot export type containing bare resource type")
+
+(assert_invalid
+  (component $C
+    (type $t (resource (rep i32)))
+    (export $et "t" (type $t) (type (sub resource)))
+    (component
+      (alias outer $C $et (type))))
+  "Cannot export type containing bare resource type")
+
+(assert_invalid
+  (component $C
+    (import "t" (type $it (sub resource)))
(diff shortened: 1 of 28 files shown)


lukewagner:
Sorry for the slow reply, but excited to check this out!  Could you perhaps add a README.md to the `interpreter/` directory giving step-by-step instructions (for Linux and MacOS, ideally) to go from "I've checked out this repo" to "I'm running the test suite"?

lukewagner:
One other thing that would be super-useful is to see if this reference interpreter can run the tests in [`wasm-tools/tests/local/component-model`](https://github.com/bytecodealliance/wasm-tools/tree/main/tests/local/component-model) and [`wasmtime/tests/misc-testsuite/component-model`](https://github.com/bytecodealliance/wasmtime/tree/main/tests/misc_testsuite/component-model).  The error strings won't match of course, so we can start by ignoring error string equality.  But once the reference interpreter agrees, we'll get a bunch of coverage from importing them into the reference test suite.

interpreter/test/resource/resource-simple.wast
```diff
@@ -0,0 +1,271 @@
...
+  (alias export $itwo "two" (func $f2))
+  (export "one" (func $f1))
+  (export "two" (func $f2))
+  )
+
+(component
+  (instance $deftype (instantiate $dt))
+  (instance (instantiate $use2c
+    (with "deftype" (instance $deftype))
+    (with "two" (component $use1)))))
```
lukewagner, line 113: It seems like this test is missing a final part that, e.g., requires `one` and `two` to have the same function type.

interpreter/test/resource/resource-simple.wast
```diff
@@ -0,0 +1,271 @@
...
+    (export "t1" (type $t1 (sub resource)))
+    (export "t2" (type $t2 (sub resource)))
+    (export "one" (func (param (own $t2)) (result (own $t2))))
+    (export "two" (func (param (own $t1)) (result (own $t1)))))))
+
+(assert_invalid
+  (component
+    (instance (instantiate $needs_dt2i_backwards
+      (with "deftype" (component $dt2)))))
+  "Type variable u0.0 is not u0.1")
```
lukewagner, line 151: Perhaps also add an invalid testcase that tries to import `one` and `two` with the same resource type?

interpreter/test/resource/resource-simple.wast
```diff
@@ -0,0 +1,271 @@
...
+  (instance $ii (instantiate $in)))
+
+(assert_invalid
+  (component
+    (import "in" (component $in
+      (export "t" (type (sub resource)))))
+    (instance $ii (instantiate $in))
+    (alias export $ii "t" (type $t))
+    (export "t" (type $t)))
+  "Component type may not refer to non-imported uvar")
```
lukewagner, line 236: I would think that the problem here was the lack of ascribed type, but if you did ascribe a `(sub resource)` type, it would be valid, and thus the error of "may not refer to non-imported uvar" doesn't make sense to me.

lukewagner (commented):
Lots of great tests in there!  Just a few nits and a question.

Issue #193, opened 2023-05-03
Proposal: Update WIT syntax/semantics for packages/dependencies/use and alter binary format for externnames


alexcrichton:
I've been talking with @lukewagner and @guybedford recently about #177 and how
best to tackle that in components and WIT, and I'm opening this issue to propose
changes to both WIT and the component model binary format to solve the issue at
hand and a few neighboring concerns as well. As a heads up what I'm proposing
here is a breaking change relative to the currently implemented and accepted WIT
syntax, and it's an addition of the binary syntax (while adding some ignores of
some preexisting binary syntax).

## Motivation and Existing Problems

The main motivation of these changes is that if you have a world such as:

```wit
// today's syntax
world {
    import foo: self.foo
}
```

then any dependency of `foo` will be "inlined" into the world as well:

```wit
// today's syntax
world {
    import foo-dep1: self.foo-dep1
    import foo-dep2: self.foo-dep2
    import foo: self.foo
}
```

This interacts poorly with transitive dependencies where, for example the
`proxy` world for WASI depends on a `types` interface in the `wasi-http` package
so the world ends up having `import types: ...` which is quite a "bland" name
without any context in it.

One way to fix this is to consider an "id" more often than a kebab-name. For
example the "id" of the relevant interface is `wasi:http/types` which is a much
more contextually relevant name. The problem with this, however, is that WIT
doesn't have a great means by which to discover its own package name or
namespace. For example when developing the `wasi-http` package itself you'd
simply have:

```wit
// wit/types.wit

default interface types {
  // ...
}
```

So there's no context available to say wither `wasi` or `http`. This is sort of
solve with the `deps` folder the `wit-parser` crate implements along the lines
of:

```wit
// wit/deps/http/types.wit

default interface types {
  // ...
}
```

where here the parser at least knows the package name is `http`. Still, however,
it doesn't know anything about `wasi`.

Another problem with the `deps/*` folder syntax is that packages are named after
their directory names meaning that there's no way to have two packages of the
same name. For example two packages from different registries or two packages
from different versions of the same registry.

In general many of the above problems are solvable in isolation somewhat but
I'm hoping to solve them all at once with updates to the WIT syntax, binary
format, and translation to wasm. One final point worth addressing here, or at
least planning to be addressed here, is at least somewhat introducing the
concept of versioning for future use. This isn't intended to be comprehensive
just yet, but should help set up some foundations hopefully.

## Changes to WIT

[...]

PR #198, opened 2023-05-17
Update WIT and binary import/export syntax

alexcrichton:
This is an update to `WIT.md` and a minor update to `Binary.md` to reflect the changes discussed in #193. Lots more discussion is in the issue, and a summary of the changes is:

* WIT `use` statements are almost entirely replaced. Old `pkg` and `self`-style paths are all gone.
* WIT documents are gone as a concept, a package is a collection of interfaces/worlds now.
* The `externname` binary format production has been updated to either be a kebab-name or an "ID". The URL field has been dropped.
* Imports and exports of interfaces in WIT worlds is now done through IDs instead of kebab-names.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -34,22 +34,28 @@ three components:
 
 * A namespace, for example `foo` in `foo:bar`. This namespace is intended to
   disambiguate between registries, top-level organizations, etc. For example
-  WASI interfaces use the `wasi:` namespace.
+  WASI interfaces use the `wasi` namespace.
 
-* A package name, for example `bar` in `foo:bar`. This provides a further
-  description of the contained functionality within the namespace. For example
-  `wasi:clocks` has the package name `clocks`.
+* A package name, for example `clocks` in `wasi:clocks`. A package name groups
+  together a set of interfaces and worlds that would otherwise be named with a
+  common prefix.
 
-* An optional version, specified as a major and minor number. Currently full
-  semver isn't allowed and instead only `1.0` is allowed, for example. Both
-  version numbers must be unsigned integers. Note that the version field is
-  optional, though.
+* An optional version, specified as a major and minor number. Because WIT
+  packages define *interfaces*, not *implementations*, there is not a patch
+  number, as in full semver. Both version numbers must be unsigned integers.
+  Note that the version field is optional, though.
 
 Package identifiers are specified at the top of a WIT file via a `package`
 declaration:
 
 ```wit
 package wasi:clocks
+```
+
+or
+
+```wit
+package wasi:clocks@1.2
 ```
 
 WIT packages can be defined in a collection of files and at least one of them
@@ -97,16 +103,17 @@ and [function][functions] definitions. For example:
 ```wit
 package wasi:filesystem
 
-interface filesystem {
-  use types.{errno}
+interface types {
+  use wasi:clocks.wall-clock.{datetime}
 
   record stat {
     ino: u64,
     size: u64,
+    mtime: datetime,
     // ...
   }
 
-  stat-file: func(path: string) -> result<stat, errno>
+  stat-file: func(path: string) -> result<stat>
 }
 ```
 
@@ -199,9 +206,8 @@ world your-world {
 ```
 
 The kebab name of the `import` or `export` is the name of the corresponding item
-in the final component. This can be different from the [`interface`][interfaces]
-name but must be a [valid identifier][identifiers]. Note that kebab names cannot
-overlap and must be unique, even between imports and exports.
+in the final component. Note that kebab names cannot overlap and must be unique,
+even between imports and exports.
 
 Note that in the component model imports to a component either use an ID or a
 kebab-name, and in WIT this is reflected in the syntax:
@@ -217,8 +223,8 @@ world command {
   // generates an import of the ID `wit:demo/my-interface`
   import my-interface
 
-  // generates an import of the ID `wasi:random/random`
-  import wasi:random/random
+  // generates an import of the ID `wasi:filesystem/types`
+  import wasi:filesystem/types
 
   // generates an import of the kebab-name `foo`
   import foo: func()
@@ -450,8 +456,8 @@ would generate this component:
 
 Here it can be seen that despite the `world` only listing `host` as an import
 the component additionally imports a `wit:demo/shared` interface. This is due
-to the fact that the `use { ... } from shared` implicitly requires that
-`shared` is imported into the component as well.
+to the fact that the `use shared.{ ... }` implicitly requires that `shared` is
+imported into the component as well.
 
 Note that the name `"wit:demo/shared"` here is derived from the name of the
 `interface` plus the package ID `wit:demo`.
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -14,10 +14,10 @@ The Wasm Interface Type (WIT) format is an [IDL] to provide tooling for the
   on a WIT definition of a shared set of APIs between platforms.
 
 A WIT package is a collection of WIT [`interface`s][interfaces] and
-[`world`s][worlds] defined in files that that uses the file extension `wit`,
-for example `foo.wit`, and is encoded as valid utf-8 bytes. Types can be
-imported between interfaces within a package and additionally from other
-packages through IDs.
+[`world`s][worlds] defined in files in the same directory that that all use the
+file extension `wit`, for example `foo.wit`. Files are encoded as valid utf-8
+bytes. Types can be imported between interfaces within a package and
+additionally from other packages through IDs.
 
 This document will go through the purpose of the syntactic constructs of a WIT
 document, a pseudo-formal [grammar specification][lexical-structure], and
@@ -206,10 +206,9 @@ world your-world {
 ```
 
 The kebab name of the `import` or `export` is the name of the corresponding item
-in the final component. Note that kebab names cannot overlap and must be unique,
-even between imports and exports.
+in the final component.
 
-Note that in the component model imports to a component either use an ID or a
+In the component model imports to a component either use an ID or a
 kebab-name, and in WIT this is reflected in the syntax:
 
 ```wit
@@ -236,6 +235,10 @@ world command {
 }
 ```
 
+Kebab names cannot overlap and must be unique, even between imports and exports.
+IDs, however, can be both imported and exported. The same interface cannot be
+explicitly imported or exported twice.
+
 ## WIT Packages and `use`
 [use]: #wit-packages-and-use
 
@@ -252,7 +255,8 @@ and at the top-level of a WIT file.
 
 #### Interfaces, worlds, and `use`
 
-The `use` statement here can be used to import types between interfaces:
+A `use` statement inside of an `interface` or `world` block can be used to
+import types:
 
 ```wit
 package local:demo
@@ -269,9 +273,9 @@ interface my-host-functions {
 ```
 
 The `use` target, `types`, is resolved within the scope of the package to an
-`interface`, in this case defined prior. Afterwards a list of types are provided
+interface, in this case defined prior. Afterwards a list of types are provided
 as what's going to be imported with the `use` statement. The interface `types`
-may textually come either after or before the `use` directive's `interface`.
+may textually come either after or before the `use` directive's interface.
 Interfaces linked with `use` must be acyclic.
 
 Names imported via `use` can be renamed as they're imported as well:
@@ -398,8 +402,28 @@ world my-world {
 The meaning of this and the previous world are the same, and `use` is purely a
 developer convenience for providing smaller names if necessary.
 
-One possible use case for this is importing the same interface with different
-versions:
+The interface referred to by a `use` is the name that is defined in the current
+file's scope:
+
+```wit
+package local:demo
+
+use wasi:http/types   // defines the name `types`
+use wasi:http/handler // defines the name `handler`
+```
+
+Like with interface-level-`use` the `as` keyword can be used to rename the
+inferred name:
+
+```wit
+package local:demo
+
+use wasi:http/types as http-types
+use wasi:http/handler as http-handler
+```
+
+Note that these can all be combined to additionally import packages with
+multiple versions and renaming as different identifiers.
 
 ```wit
 package local:demo
@@ -446,10 +470,10 @@ would generate this component:
     (type $metadata (record (; ... ;)))
     (export "metadata" (type (eq $metadata)))
   ))
-  (alias export $shared "metadata" (type $metadata'))
+  (alias export $shared "metadata" (type $metadata_from_shared))
   (import "host" (instance $host
-    (export "metadata" (type $metadata (eq $metadata')))
-    (export "get" (func (result $metadata)))
+    (export $metadata_in_host "metadata" (type (eq $metadata_from_shared)))
+    (export "get" (func (result $metadata_in_host)))
   ))
 )
 ```
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -77,7 +77,7 @@ belong to an interface.
 An example of an interface is:
 
 ```wit
-package wit:demo
+package local:demo
 
 interface host {
   log: func(msg: string)
@@ -90,7 +90,7 @@ would correspond to:
 
 ```wasm
 (component
-  (import (interface "wit:demo/host") (instance $host
+  (import (interface "local:demo/host") (instance $host
     (export "log" (func (param "msg" string)))
   ))
   ;; ...
@@ -138,7 +138,7 @@ equivalent of a `component` type in the component model. For example this
 world:
 
 ```wit
-package wit:demo
+package local:demo
 
 world my-world {
   import host: interface {
@@ -168,7 +168,7 @@ Worlds can contain any number of imports and exports, and can be either a
 function or an interface.
 
 ```wit
-package wit:demo
+package local:demo
 
 world command {
   import wasi:filesystem/filesystem
@@ -190,7 +190,7 @@ Additionally interfaces can be defined inline with an explicit kebab-name that
 avoids the need to have an out-of-line definition.
 
 ```wit
-package wit:demo
+package local:demo
 
 interface out-of-line {
   the-function: func()
@@ -213,14 +213,14 @@ Note that in the component model imports to a component either use an ID or a
 kebab-name, and in WIT this is reflected in the syntax:
 
 ```wit
-package wit:demo
+package local:demo
 
 interface my-interface {
   // ..
 }
 
 world command {
-  // generates an import of the ID `wit:demo/my-interface`
+  // generates an import of the ID `local:demo/my-interface`
   import my-interface
 
   // generates an import of the ID `wasi:filesystem/types`
@@ -255,7 +255,7 @@ and at the top-level of a WIT file.
 The `use` statement here can be used to import types between interfaces:
 
 ```wit
-package wit:demo
+package local:demo
 
 interface types {
   enum errno { /* ... */ }
@@ -277,7 +277,7 @@ Interfaces linked with `use` must be acyclic.
 Names imported via `use` can be renamed as they're imported as well:
 
 ```wit
-package wit:demo
+package local:demo
 
 interface my-host-functions {
   use types.{errno as my-errno}
@@ -299,7 +299,7 @@ interface types {
 }
 
 // host.wit
-package wit:demo
+package local:demo
 
 interface my-host-functions {
   use types.{errno, size}
@@ -314,7 +314,7 @@ the same syntax is used in `import` and `export` directives:
 
 ```wit
 // a.wit
-package wit:demo
+package local:demo
 
 world my-world {
   import host
@@ -336,7 +336,7 @@ When referring to an interface an ID form can additionally be used to refer to
 dependencies. For example above it was seen:
 
 ```wit
-package wit:demo
+package local:demo
 
 world my-world {
   import wasi:clocks/monotonic-clock
@@ -349,7 +349,7 @@ This is the package identified by `wasi:clocks` and the interface
 well:
 
 ```wit
-package wit:demo
+package local:demo
 
 interface my-interface {
   use wasi:http/types.{request, response}
@@ -362,7 +362,7 @@ If a package being referred to has a version number, then using the above syntax
 so far it can get a bit repetitive to be referred to:
 
 ```wit
-package wit:demo
+package local:demo
 
 interface my-interface {
   use wasi:http/types@1.0.{request, response}
@@ -380,7 +380,7 @@ interfaces within the scope of the file itself. For example the above could be
 rewritten as:
 
 ```wit
-package wit:demo
+package local:demo
 
 use wasi:http/types@1.0
 use wasi:http/handler@1.0
@@ -402,7 +402,7 @@ One possible use case for this is importing the same interface with different
 versions:
 
 ```wit
-package wit:demo
+package local:demo
 
 use wasi:http/types@1.0 as http-types1
 use wasi:http/types@2.0 as http-types2
@@ -421,7 +421,7 @@ component.
 For example this document:
 
 ```wit
-package wit:demo
+package local:demo
 
 interface shared {
   record metadata {
@@ -442,7 +442,7 @@ would generate this component:
 
 ```wasm
 (component
-  (import (interface "wit:demo/shared") (instance $shared
+  (import (interface "local:demo/shared") (instance $shared
     (type $metadata (record (; ... ;)))
     (export "metadata" (type (eq $metadata)))
   ))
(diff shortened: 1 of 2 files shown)


alexcrichton:
I'll also note that I'm in the process of implementing these changes. Current work is:

* https://github.com/bytecodealliance/wasm-tools/pull/1027
* https://github.com/bytecodealliance/wasmtime/pull/6390

design/mvp/WIT.md
```diff
@@ -28,6 +27,38 @@ package suitable for distribution.
...
 [components]: https://github.com/webassembly/component-model
 
+## Package identifiers
+
+All WIT packages are assigned an "ID". IDs look like `foo:bar@1.0` and have
+three components:
+
+* A namespace, for example `foo` in `foo:bar`. This namespace is intended to
+  disambiguate between registries, top-level organizations, etc. For example
+  WASI interfaces use the `wasi:` namespace.
```
lukewagner, line 37:
```suggestion
  WASI interfaces use the `wasi` namespace.
```

design/mvp/WIT.md
```diff
@@ -28,6 +27,38 @@ package suitable for distribution.
...
+
+* A package name, for example `bar` in `foo:bar`. This provides a further
+  description of the contained functionality within the namespace. For example
```
lukewagner, line 40:
```suggestion
* A package name, for example `bar` in `foo:bar`. A package name groups together a set
  of interfaces and worlds that would otherwise be named with a common prefix. For example
```

design/mvp/WIT.md
```diff
@@ -28,6 +27,38 @@ package suitable for distribution.
...
+  `wasi:clocks` has the package name `clocks`.
+
+* An optional version, specified as a major and minor number. Currently full
+  semver isn't allowed and instead only `1.0` is allowed, for example. Both
```
lukewagner, line 44:
```suggestion
* An optional version, specified as a major and minor number. Because WIT packages define
  *interfaces*, not *implementations*, there is not a patch number, as in full semver. Both
```

design/mvp/WIT.md
```diff
@@ -62,8 +95,10 @@ An `interface` can contain [`use`][use] statements, [type][types] definitions,
 and [function][functions] definitions. For example:
 
 ```wit
-interface wasi-fs {
-  use pkg.types.{errno}
+package wasi:filesystem
+
+interface filesystem {
+  use types.{errno}
```
lukewagner, line 101:
```suggestion
package wasi:filesystem

interface types {
  use wasi:clocks.wall-clock.{datetime}
```
The goal here being to have this example slightly more-closely align with the real `wasi:filesystem` which wants to define the interface ID `wasi:filesystem/types`.  If this change makes sense, you'd probably stick a `datetime` inside the `stat` record and remove `errno` from the `result` of `stat-file`.

design/mvp/WIT.md
```diff
@@ -137,217 +161,248 @@ Worlds can contain any number of imports and exports, and can be either a
...
   }
 }
 ```
 
-The name of the `import` or `export` is the name of the corresponding item in
-the final component. This can be different from the [`interface`][interfaces]
-name but must be a [valid identifier][identifiers].
+The kebab name of the `import` or `export` is the name of the corresponding item
+in the final component. This can be different from the [`interface`][interfaces]
+name but must be a [valid identifier][identifiers]. Note that kebab names cannot
```
lukewagner, line 203:
```suggestion
in the final component. Note that kebab names cannot
```

design/mvp/WIT.md
```diff
@@ -137,217 +161,248 @@ Worlds can contain any number of imports and exports, and can be either a
...
-```
 
-If no `default` world is specified in the WIT document and no named world is
-explicitly chosen then bindings cannot be generated.
+world command {
+  // generates an import of the ID `wit:demo/my-interface`
+  import my-interface
+
+  // generates an import of the ID `wasi:random/random`
+  import wasi:random/random
```
lukewagner, line 221:
```suggestion
  // generates an import of the ID `wasi:filesystem/types`
  import wasi:filesystem/types
```

lukewagner: (To avoid the confusion of "why are there two `random`s?".)

design/mvp/WIT.md
```diff
@@ -379,59 +436,25 @@ would generate this component:
...
   ))
 )
 ```
 
 Here it can be seen that despite the `world` only listing `host` as an import
-the component additionally imports a `shared` instance. This is due to the fact
-that the `use { ... } from shared` implicitly requires that `shared` is imported
-into the component as well.
+the component additionally imports a `wit:demo/shared` interface. This is due
+to the fact that the `use { ... } from shared` implicitly requires that
```
lukewagner, line 453:
```suggestion
to the fact that the `use shared.{ ... }` implicitly requires that
```

design/mvp/WIT.md
```diff
@@ -1114,22 +1188,25 @@ would correspond to:
...
 (component
-  (type (export "host") (component
+  (type $demo (component
     (type $console (instance
       (export "log" (func (param "arg" string)))
     ))
-    (export "console" (instance (type $console)))
+    (export (interface "wit:demo/console") (instance (type $console)))
   ))
+  (export (interface "wit:demo/wit") (type $demo))
```
lukewagner, line 1197: I'm a bit confused where the final `/wit` comes at the end of `wit:demo/wit`.  If we want to preserve file grouping, I'd expect the ID to be `wit:demo/host` or even just the kebab-name `host` (highlighting the fact that `host` is not an *interface*).  If we don't want to preserve file grouping, then maybe export the whole package with the kebab-name of the package (again, because it's not an *interface*).

lukewagner (commented):
Looks good!  Some small suggestions and one interesting question at the end about the encoding of Wit into Wat.

design/mvp/WIT.md
alexcrichton: Good question! This is here though not for interfaces but instead for the package itself. I copy-pasted this output from `wasm-tools component wit` which is why it's here, but my thinking is that the package name should be recorded somewhere, probably around here, but IDs require a `/foo` and when naming the package there's nothing to put there, so I arbitrarily put in `/wit`.

This doesn't actually reflect file structure though, it's only there to satisfy the validation requirements of an interface name being present. All contents of the package are in the component type itself.

Basically this ideally would be simply `"wit:demo"`, but that doesn't validate.

lukewagner: Ah, hah, gotcha.  One observation is that, since the namespace, package name and version are already present inside each exported interface and world ID, there's no additional info provided by this.  In fact, I wonder if we even need this enclosing component-type at all: could the outermost component simply contain an *unnamed* `type`-definition for each interface-or-world which is a component-type that imports `use`d dependencies and exports the interface-or-world, establishing the ID via `(export (interface ...) ...)`?

alexcrichton: Validation-wise, yes, that should work. Integration-wise though that would be more difficult I think though because index spaces are generally considered to be implementation details of wasms where the exports/imports define how the embedder interacts with it. 

One other issue though is that if a package is empty with no other interfaces or worlds I don't think the ID would be recorded anywhere?

design/mvp/WIT.md
```diff
@@ -137,217 +161,248 @@ Worlds can contain any number of imports and exports, and can be either a
...
 
 world your-world {
-  import out-of-line: self.out-of-line
+  import out-of-line
   // ... is roughly equivalent to ...
-  import out-of-line2: interface {
+  import out-of-line: interface {
     the-function: func()
   }
 }
```
guybedford, line 198: I get this framing, and it's a nice approach, but I wasn't aware we were going to entirely separate kebab names as no longer being able to reference other dependencies at all.

Again for the componentization use case, if I want to target a dependency that is another component wasm file, it would be nice to be able to have tooling that could do eg:

```
world parent-world {
  import dep
}
```

and then be able to define `dep` to a local component when componentizing:

```
jco componentize x.js --world parent-world --dep dep=./child.wasm
```

It would be nice to leave the door open to interesting use cases like the above.

guybedford (commented):
Really great work. This all seems in line with previous discussions. I have two major concerns:

1. Making package declarations required adds possible redundancy to componentization use cases.
2. Disallowing non-package imports as being resolved by tooling bypasses some interesting usage scenarios.

Both seem relatively surface-level restrictions to me though, and I've tried to go into as much detail as I can on these.

design/mvp/WIT.md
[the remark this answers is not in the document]
lukewagner: Instead of making namespaces optional, which I think would be a half-way measure, I think these cases should just not use `(interface <id>)` and produce plain kebab-names instead.

design/mvp/WIT.md
```diff
@@ -28,6 +27,38 @@ package suitable for distribution.
 [IDL]: https://en.wikipedia.org/wiki/Interface_description_language
 [components]: https://github.com/webassembly/component-model
 
+## Package identifiers
+
+All WIT packages are assigned an "ID". IDs look like `foo:bar@1.0` and have
+three components:
```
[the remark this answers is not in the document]
lukewagner: I agree that it's useful to have `package` names be optional, in which case all the interfaces being defined by the collection of Wit documents passed to the tool don't have any associated interface ID: it's all kebab names.

design/mvp/WIT.md
lukewagner: I suppose another slight variation is for each interface/world to be a separate type definition that is exported with the ID (which would then be always be duplicated).  I guess mainly I was realizing that it doesn't seem like the component-types representing each individual interface/world need to be wrapped up into a single component type.

It's true that if the package was empty, there wouldn't be any record of the package name, but it seems like that's fine, since there's nothing to refer to.

alexcrichton: I did actually initially go for the containerless format like you're mentioning. The names actually already duplicate the package name as each interface is exported with its full ID rather than just its name. The problem is that without a container you're generating an actual component with actual imports since interfaces may need to import interfaces they depend on. That would change things radically since this is supposed to be "just" type-level things.

design/mvp/WIT.md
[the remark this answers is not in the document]
alexcrichton: I agree it would be useful to be able to omit `package` altogether, but that raises enough questions for me that I'm hoping we can get by for now by not implementing it. For example what @lukewagner is proposing means that worlds have drastically different meanings whether or not the enclosing package has an ID or not, and I'm not sure I'm comfortable with that. I would expect some form of opt-in to "yes please interpret everything differently here" more than leaving out `package ...`.

For the JS componentization use case though what you'll be doing is inferring a `world` from a concrete component, and in such a situation there may not have actually ever been a `world` written down in WIT anywhere corresponding to what you're inferring. For example you're inferring a "minimized" version which will be missing functions that aren't otherwise required (e.g. you import `wasi:filesystem/filesystem`'s `stat` function but not `readdir` or something like that). The WIT here, however, entirely synthesized since all you're given was a component. Currently this synthesizes a dummy `package root:root` with a `world root { ... }` in the `Resolve` data structure, since otherwise there's no names to fill in.

Put another way I think this would be good to get to at some point but for now I'm hoping that writing `package foo:foo` at the top or something like that is "good enough".

design/mvp/WIT.md
```diff
@@ -28,6 +27,38 @@ package suitable for distribution.
...
+Package identifiers are specified at the top of a WIT file via a `package`
+declaration:
+
+```wit
+package wasi:clocks
+```
+
+WIT packages can be defined in a collection of files and at least one of them
+must specify a `package` identifier. Multiple files can specify a `package` and
+they must all agree on what the identifier is.
```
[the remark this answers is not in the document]
alexcrichton: I'm tempted to start conservatively and require them in the text, but do you have a use case in mind where this would be helpful to have?

design/mvp/WIT.md
guybedford: > If we go via the route described above where this form automatically uses kebab-names in some situations, I think that would satisfy the use case here?

Can you clarify what you mean by this?

design/mvp/WIT.md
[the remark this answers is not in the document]
guybedford: If we supported the kebab name variant `package foo` that would be fine, but in particular it would be nice to retain the ability to have WIT files that are not part of packages when just using their interfaces and inlining their types, without inlining their names. Ie at the top-level WIT authoring level where you don't even think in terms of a package name yet.

design/mvp/WIT.md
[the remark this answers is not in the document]
guybedford: @alexcrichton in the above case I would have expected that `wasi:filesystem/filesystem` would still be a direct import of the WASI externname host. It would only be the kebab name types (and implementations) which would be inlined into the component surely?

The fact that we are inferring a `world` from a component is exactly the use case though - componentization is creating components from worlds. Handling that in dependencies is exactly recreating worlds from **these** components. I don't think this use case should be surprising?

I don't fully follow how kebab names break the model in terms of them just being package-local resolutions. I'm fine with some kind of `root:root` pattern to make this stuff work internally, and I understand if you need to move ahead here, my main point is just asking are just:

1. Will these use cases be possible, so we are not cutting them out entirely?
2. Are we not making unnecessary restrictions now that will hinder these use cases unnecessarily now or in future?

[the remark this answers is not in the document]
guybedford: After discussing this further with @alexcrichton, he clarified there may still be ways of supporting kebab name external references in future. But for now that is a secondary use case, since inline types solve the majority of the problem. So effectively:

1. Yes these use cases are possible, but would require further design work in future motivated by real need
2. There are no restrictions being made now that inhibit the externalization of type definitions for kebab names in future

design/mvp/WIT.md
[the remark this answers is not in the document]
guybedford: After further discussion, the `local:pkg` convention for now will be fine. In future out-of-band configurations eg for kebab names may be possible, depending on how type externalization for them is managed.

guybedford (approved):
Thanks @alexcrichton for working through my questions today. To summarize -

The kebab names are still being supported fine with inline types which mostly handles the major use cases. Any additional use cases that may be supported in future for kebab names or instance imports can be designed on their own merits.

Finally, with regards to having good conventions for unknown package namespaces in the "optional" scenarios, I've suggested the use of `local:` in the examples as a possible name here for now. I'm open to other alternatives, but that would mitigate my concern of `wit:` being used as an example here then being taken up more widely (or some other ambiguous namespacing). Happy to bikeshed further of course!

design/mvp/WIT.md
lukewagner: Good point, I was misreading the current output and you're right.  Thinking about it some more, given the following Wit:
```wit
package wasi:http
interface types {
  resource request { ... }
  resource response { ... }
}
interface handler {
  use types.{request, response}
  handle: func(request) -> response
}
world proxy {
  import handler
  export handler
}
```
I think perhaps we'd want to get the following Wat encoding:
```wat
(component
  (component (export "types")
    (type (export (interface "wasi:http/types")) (instance
      (export "request" (type (sub resource)))
      (export "response" (type (sub resource)))
    ))
  )
  (component (export "handler")
    (import (interface "wasi:http/types") (instance $types
      (export "request" (type (sub resource)))
      (export "response" (type (sub resource)))
    ))
    (alias export $types "request" (type $request))
    (alias export $types "response" (type $response))
    (type (export (interface "wasi:http/handler")) (instance
      (export "handle" (func (param "request" (own $request)) (result (own $response))))
    ))
  )
  (component (export "proxy")
    (type (export (interface "wasi:http/proxy")) (component
      (import (interface "wasi:http/types") (instance $types
        (export "request" (type (sub resource)))
        (export "response" (type (sub resource)))
      ))
      (alias export $types "request" (type $request))
      (alias export $types "response" (type $response))
      (import (interface "wasi:http/handler") (instance
        (export "handle" (func (param "request" (own $request)) (result (own $response))))
      ))
      (export (interface "wasi:http/handler") (instance
        (export "handle" (func (param "request" (own $request)) (result (own $response))))
      ))
    ))
  ))
)
```
Here, each interface and world is individually wrapped in a component that is able to express `use` dependencies via imports.  For worlds, there's currently never a need to have imports in the wrapper, but I'm not 100% sure this would never make sense in the future and it seems nice to have interfaces and worlds be symmetric, distinguished by whether the wrapper component exports an `instance` type (for an interface) or a `component` type (for a world).

A nice thing about this encoding is that it makes it clear that each interface/world definition is a standalone unit and thus packages can be arbitrarily split and merged.  It also makes clear that each interface's `use`s can be resolved independently and/or in multiple ways, depending on context, which is less clear with the current encoding where they alias each other directly.

Another nice thing is that, given an interface ID like `wasi:http/handler`, you can think of the `wasi:http` as mapping to the above component package and then the trailing `handler` projects out the `handler` subcomponent which is also how nested component *implementation* packages would work.  This also naturally generalizes to deeper nesting (i.e., `foo:bar/baz/quux`) if we ever wanted that.  Also, if we ever wanted to allow `foo:bar` to refer to a specific interface or world, we could achieve that by having the outermost component contain `(type (export (interface "foo:bar")) ...)`, which we could do while still allowing there to be interfaces/worlds named `foo:bar/baz`.  (E.g., `wasi:http` could map to what's currently called `wasi:http/proxy`, if we wanted.)

WDYT?

alexcrichton: I suspect something along these lines could work, but there's two issues given the current translation as I understand it:

1. An empty wit package would be translated as `(component)` which doesn't include the package's name and it's also ambiguous with an acutal empty component itself. While it's sort of an edge case this is all so low-level I'd ideally prefer to avoid any accidental ambiguities with things like this.
2. Do exports have to be of actual concrete components? An operation supported today is "extract the WIT from this concrete component" and that transparently works with actual components and wasm-encoded WIT packages. The detection is currently "does this component export a component type" since no real component will realistically do that for a long time. This detection, however, gets more complicated as the structure of the component looks "real" until you get to the exported instance/component types. Additionally I personally thought it was a nice feature to encode the entire WIT package in the type section, where this is sort of half actual runtime things half type things.

lukewagner: Yeah, both great questions.  I'm not sure these are the absolute best answers, but I'll put them out there to see what you think:
1. In a full workflow where this package gets published, the registry to which it was published would establish the name.  We've also talked about a registry-defined package metadata custom section that would contain information like this too.  I suppose why it seems fine to have a different answer here is that packages are not "semantic" things with IDs in the same way that interfaces and worlds are.
2. I could see s/component def/component type/ working too, almost equivalently, so I've flip-flopped in my head on this one.  For rendering `.wit` from `.wasm`, I think the "signal" would be any `(export (interface ...) (type ...))` and so the Wit-rendering tool would recurse over the component tree and splat out each of these it found into the generated `.wit` document.  As a nice side effect, you could then bundle `.wit` with *any* component `.wasm`.  The other small neat thing about using component defs as the "wrapper" is that last little thing I mentioned about being able to have the outermost component serve as the wrapper (possibly, at some point in the future).

design/mvp/WIT.md
```diff
@@ -28,6 +27,44 @@ package suitable for distribution.
...
+  disambiguate between registries, top-level organizations, etc. For example
+  WASI interfaces use the `wasi` namespace.
+
+* A package name, for example `clocks` in `wasi:clocks`. A package name groups
+  together a set of interfaces and worlds that would otherwise be named with a
+  common prefix.
+
+* An optional version, specified as a major and minor number. Because WIT
+  packages define *interfaces*, not *implementations*, there is not a patch
+  number, as in full semver. Both version numbers must be unsigned integers.
```
lukewagner, line 45: I did realize a potential use case for including an optional patch version: when you're iterating on a pre-1.0 interface, and minor version bumps can be breaking, then you might want to use patch version changes to signal non-breaking interface changes.  Concretely, I could imagine that for Preview 2, we'd want to publish `0.` versions of WASI interfaces (say, `0.2`), and then if we make non-breaking changes post-Preview 2, we could publish them as `0.2.x` versions (saving `0.3` for Preview 3).

rylev: We could also allow a pre-release version number and potentially a build number. I think this would more closely match the actual semantics (i.e., it's not a patch; it's a pre-release)

tschneidereit: In the past we had discussed using patch versions for non-functional changes. Perhaps the only, but still important, change of this kind would be documentation changes. I think those warrant the inclusion of patch version numbers

design/mvp/WIT.md
```diff
@@ -137,217 +168,247 @@ Worlds can contain any number of imports and exports, and can be either a
...
     the-function: func()
   }
 }
 ```
 
-The name of the `import` or `export` is the name of the corresponding item in
-the final component. This can be different from the [`interface`][interfaces]
-name but must be a [valid identifier][identifiers].
+The kebab name of the `import` or `export` is the name of the corresponding item
+in the final component. Note that kebab names cannot overlap and must be unique,
```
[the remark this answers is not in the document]
alexcrichton: Good point! I've clarified that this use case is allowed.

design/mvp/WIT.md
```diff
@@ -28,6 +27,44 @@ package suitable for distribution.
...
+
+or
+
+```wit
+package wasi:clocks@1.2
+```
+
+WIT packages can be defined in a collection of files and at least one of them
+must specify a `package` identifier. Multiple files can specify a `package` and
+they must all agree on what the identifier is.
```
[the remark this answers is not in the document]
alexcrichton: @guybedford do you have thoughts on this?

design/mvp/WIT.md
```diff
@@ -40,6 +77,8 @@ belong to an interface.
 An example of an interface is:
 
 ```wit
+package local:demo
```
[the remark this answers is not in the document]
alexcrichton: @guybedford you might have thoughts on this? (I don't mind either way myself)

design/mvp/WIT.md
alexcrichton: Ok sounds reasonable to me, I went ahead and added full semver in including pre-release/build metadata.

design/mvp/WIT.md
alexcrichton: Given how low-level this tooling is though I'd ideally prefer to not require or assume a registry in place. Basically otherwise everything will need to carry the name around manually separate from the wasm itself, whereas otherwise it's stored inline in the wasm with the current proposal.

I'm still not sure I understand the "neat" part here though myself. Everything feels far enough in the future that nothing is really solid enough that we would be able to predict that an alternate design would indeed fit neatly into a design one way or another. I'm hesitant to change designs now radically for a proposed future design which isn't fully fleshed out.

lukewagner: Regarding the registry aspect: sorry, I didn't mean to imply that the package name lives in the registry, but rather a custom section in the component binary that could also hold other package-metadata types of fields.

That being said, in the short term, maybe it's better to just make the minimal change to the existing encoding to unblock other stuff.  Because we'll eventually be publishing these Wit packages to a registry, the encoding scheme will need to be effectively standardized to allow tools to interoperate and then it'll be important to think long-term about some of these packaging questions, but we're not there yet.

lukewagner (approved):
lgtm, thanks!  Explainer.md and the text format still need to be updated, but I'm happy to do that as a quick follow-on once you merge this PR.

design/mvp/WIT.md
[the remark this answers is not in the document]
guybedford: My concern was exactly that asking users to choose an arbitrary namespace at an early stage risks exactly the naming problem namespaces are supposed to avoid in the first place.

design/mvp/WIT.md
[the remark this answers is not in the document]
guybedford: I'd be open to changing the constraint to a single file, but see no problem with the redundancy either.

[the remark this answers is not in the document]
guybedford: Also the concept of `main` doesn't quite apply to WIT currently since all WIT files are equal. Unless we introduce preferential lookup rules for default worlds and interfaces.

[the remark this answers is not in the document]
rylev: There are other tools that have this convention. For example, terraform has the convention of main files that (I believe) have no significance other than to say to the reader "look here first".

design/mvp/WIT.md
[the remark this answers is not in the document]
rylev: So does `local` actually have a meaning here or is this a convention you're trying to establish or is this purely descriptive? Either way, it would be nice to explicitly state that in this document.

guybedford:
With regards to the ESM integration, it would be great to handle this in this PR as well since those behaviours do need to be worked out as part of this. My suggestion would be a note along the lines of the following bullet points:

- When importing components in the ESM integration, web-compatible components are those that do not use namespaces but instead use kebab names in the outer import and export positions, or optionally URLs in the import position.
- When converting components into web compatible components, namespaced exports may be provided by convention as interfaces nested under the namespace and package interface names. In addition, if there are no conflicts against other exports, the interface and package names may be further flattened without nesting in this convention.

alexcrichton:
Ok I'm going to go ahead and merge this, if at least to have the current spec reflect what's implemented now in tooling. I talked with @lukewagner and he'll update `Explainer.md` shortly afterwards and I think any further minor updates can happen in subsequent PRs. Thanks all for the discussion here!

lukewagner:
ESM-integration updates aren't "minor", but I do think this and other language integrations questions can be done as follow-ups without changing what's in the AST and binary, so it's a good idea to merge now and iterate forward.  I'll be working to update some other parts of Explainer.md shortly.

PR #205, opened 2023-06-16
Update Explainer.md to match Wit interface id changes in Binary.md

lukewagner:
This PR updates Explainer.md to match the Binary.md/Wit.md changes in #198.

Working on this, I saw a few cases where the grammar for a "version" in Binary.md/Wit.md is a slight subset of what's on semver.org (particularly for `pre-release identifier` and `build identifier`).  Rather than restating it all verbatim, I thought maybe we could just literally reuse the production (it seems precise enough and I don't think there's anything we need to slice out), so this PR also does that in Binary.md/Wit.md.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -102,8 +102,9 @@ Notes:
   for aliases (below).
 * Validation of `core:instantiatearg` initially only allows the `instance`
   sort, but would be extended to accept other sorts as core wasm is extended.
-* Validation of `instantiate` requires that `name` is present in an
-  `externname` of `c` (with a matching type).
+* Validation of `instantiate` requires that each `<name>` or `<iid>` in an
+  imported `externname` in `c` matches a `string` in a `with` argument and that
+  the argument's type matches the import's type.
 * When validating `instantiate`, after each individual type-import is supplied
   via `with`, the actual type supplied is immediately substituted for all uses
   of the import, so that subsequent imports and all exports are now specialized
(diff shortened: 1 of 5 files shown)


alexcrichton (approved):
👍 Thanks for writing this up!

design/mvp/Binary.md
```diff
@@ -327,33 +328,21 @@ flags are set.
...
-verbuild ::=                                                           => ϵ
-           | '+' (a:<label> '.') b:<label>                             => +(a.*)b
+import      ::= en:<externname> ed:<externdesc>                      => (import en ed)
+export      ::= en:<externname> si:<sortidx> ed?:<externdesc>?       => (export en si ed?)
+externname  ::= 0x00 n:<name>                                        => n
+              | 0x01 iid:<iid>                                       => (interface iid)
+iid         ::= len:<u32> c:<iid-chars>                              => c (if len = |c|)
+iid-chars   ::= ns:<label> ':' pkg:<label> '/' n:<label> v:<version> => ns:pkg/nv
+version     ::=                                                      => ϵ
+              | '@' v:<valid semver>                                 => @v
```
guybedford, line 338: One benefit of having explicit semver syntax is that we can make the minor / patch versions optional to allow variants like `pkg:name@1.2` or `pkg:name@1` for basic range support. Of course we could always add this back if we decide to go in that direction.

guybedford (approved):

design/mvp/Binary.md
lukewagner: Sounds right to me.  Thus far, I haven't been able to imagine what a version range on an `interface` import/export would mean, but perhaps we'll see a motivating example in the future.  (For an implementation dependency, it makes tons of sense, of course.)

PR #206, opened 2023-06-16
Add Wit.md section for resource items

lukewagner:
This PR rescues a hunk from #203 that adds a Wit.md section for `resource` definitions (including methods).

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -1002,6 +1002,59 @@ union-cases ::= ty
               | ty ',' union-cases?
 ```
 
+### Item: `resource`
+
+A `resource` statement defines a new abstract type for a *resource*, which is
+an entity with a lifetime that can only be passed around indirectly via [handle
+values](#handles). Resource types are used in interfaces to describe things
+that can't or shouldn't be copied by value.
+
+For example, the following Wit defines a resource type and a function that
+takes and returns a handle to a `blob`:
+```wit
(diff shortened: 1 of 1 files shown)


design/mvp/WIT.md
```diff
@@ -1002,6 +1002,59 @@ union-cases ::= ty
...
+A `resource` statement defines a new abstract type for a *resource*, which is
+an entity with a lifetime that can only be passed around indirectly via [handle
+values](#handles). Resource types are used in interfaces to describe things
+that can't or shouldn't be copied by value.
+
+For example, the following Wit defines a resource type and a function that
+takes and returns a handle to a `blob`:
+```wit
+resource blob
+transform: func(blob) -> blob
```
sunfishcode, line 1016: Would it make sense to also introduce a function that uses a `borrow` here, just to introduce the concept, for someone reading this top-down, before launching into more complex examples with `borrow`s below? This would also be a good place to mention the relationship between owned handles and resource lifetimes.

alexcrichton (approved):
👍 Makes me realize we need to adjust the parsing in `wit-parser` a bit, which is good though because I think the syntax here is good

design/mvp/WIT.md
lukewagner: Yes, good point!

sunfishcode (approved):
Looks good!

PR #207, opened 2023-06-17
Fix interface grammar in top-level use

cardoso:
Thought this might be missing at least the ? , but also version (unless it's meant to ignore patch etc)


cardoso:
Oops didn't see this one: https://github.com/WebAssembly/component-model/pull/205

I can update this if that's merged first.

cardoso:
Updated with the change in https://github.com/WebAssembly/component-model/pull/205

lukewagner:
Thanks!

PR #209, opened 2023-06-20
tweak own/borrow handles rules, refactor Canonical ABI logic

lukewagner:
This PR has 4 commits, extracted from the future work on child handles.  I'd recommend looking at the commits separately because the first commit is bigger, but doesn't change observable behavior (it's just refactoring, based on the better understanding of working on the generalized case), while the other 3 are small and make the following 3 tweaks:
* The first change is to only increment `lend_count` when borrowing an `own` handle (not when borrowing a `borrow`).  This is almost unobservable (because of synchronous calls) but it is (in the case of parent->child->parent reentrance) and enables a little optimization (as noted in this commit).
* The second change fixes a bug in the reentrance check when calling a destructor (the logic was over-trapping when dropping a resource your own component implemented and under-trapping when a resource didn't have a dtor (which should be an encapsulated detail)).
* The third change makes `resource.drop` more like the other two `resource.*` built-ins by taking a resource type by `typeidx`.  This makes `drop` a bit simpler and more regular to use for toolchains and I don't expect it is actually useful for optimization, given the branches are needed in any case.  It also generalizes better in the future when there are more complicated handle types.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -281,9 +281,9 @@ Notes:
 ```
 canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift f opts type-index-space[ft])
            | 0x01 0x00 f:<funcidx> opts:<opts>                   => (canon lower f opts (core func))
-           | 0x02 t:<typeidx>                                    => (canon resource.new t (core func))
-           | 0x03 t:<valtype>                                    => (canon resource.drop t (core func))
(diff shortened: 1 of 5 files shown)


alexcrichton (approved):
All seems reasonable to me 👍 

I might have further comments as I get closer to implementing the runtime semantics in Wasmtime, but no need to block any of this on that. Additionally this is technically a breaking change to the binary format, but given how almost no one is using resources at this time I feel like it's probably ok to slide this in without bumping the major version in the component header.

lukewagner:
Great, that sounds right to me.

PR #214, opened 2023-07-11
Fix lifting an own handle from a borrow handle

alexcrichton:
This fixes what I believe is a mistake in the `CanonicalABI.md` explainer currently where a lifting an owned resource handle from the table can succeed if the entry in the table was actually a borrowed handle. This inserts a check to ensure that the handle is an `own` handle before completing the lifting operation.


lukewagner (approved):
D'oh, thanks, I lost that trap in the recent refactoring.  Could you make the same change in design/mvp/canonical-abi/definitions.py as well?

PR #216, opened 2023-07-12
Handle class was renamed to HandleElem but not all references were updated

patrickhuber:
I think commit 05cef2e changed the Handle class to HandleElem but not all references were updated. 

I noticed that even with this change, the CI still passed. Is there enough code coverage to ensure this would have been caught?

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -219,7 +219,7 @@ definitions via the `cx` parameter of type `CallContext`:
 class CallContext:
   opts: CanonicalOptions
   inst: ComponentInstance
-  lenders: [Handle]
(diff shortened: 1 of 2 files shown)


lukewagner:
Thanks!  At the moment, mostly b/c I'm a Python n00b, we're not running Python in the type-checking mode, so all the type annotations are purely documentative atm.

lukewagner (approved):
Could you also make the same change in CanonicalABI.md?

patrickhuber:
> Could you also make the same change in CanonicalABI.md?

Yes, one moment

patrickhuber:
Done

lukewagner:
Thanks!

PR #218, opened 2023-07-17
In the Canonical ABI, disallow empty types.

sunfishcode:
In C++, all types must have a non-zero size, so empty structs have size 1. In C, structs with no members are non-standard, though widely supported, with size 0, making them ABI-incompatible with C++. In Go, the spec says "Two distinct zero-size variables may have the same address in memory", and as a user, my understanding is the fact that it says "may" means one isn't guaranteed it's always one way or the other.

To avoid these complexities, modify the Canonical ABI to require that all types have a size of at least one.


alexcrichton:
I do sympathize with the complexity here but I feel at least that this isn't something that we can escape from. An alternative to saying everything has a minimum size of 1 is to make the C/C++ bindings generator fancier. This "fancier" is likely a nontrivial chunk of work, however, which I'm presuming is the main motivation for proposing this change (correct me if I'm wrong though).

Going through with this change though languages like Rust will want the opposite, to represent zero-size types as actually zero-size rather than with 1 byte. So in some sense this'll shift the work from a C/C++ generator to a Rust generator.

A possible entirely different alternative to dealing with empty types, however, is to perhaps forbid them. The component model could make empty records, flags, and tuples errors along the same lines of empty variants, unions, and enums are already errors. I'm not sure if there's a concrete use case for empty things other than "it's a nice orthogonal thing to have" and it could offer the ability to defer this complexity of bindings generation to later perhaps.

sunfishcode:
That sounds good to me. I've now changed this PR to disallow empty types. If we do this, I'll follow up by submitting PRs to the WASI proposals that currently have empty types.

alexcrichton:
Could this also update `Binary.md` with the extra validation required for these types?

alexcrichton (approved):
Oh I should also say I'm ok with this at-large too

sunfishcode:
Ok, I've now change the `*`s to `+`s in the AST and binary format.

design/mvp/canonical-abi/definitions.py
```diff
@@ -249,6 +249,7 @@ def size_record(fields):
   for f in fields:
     s = align_to(s, alignment(f.t))
     s += size(f.t)
+  trap_if(s == 0)
```
lukewagner, line 252:
```suggestion
  assert(s > 0)
```

PR #222, opened 2023-07-26
Allow imports to name implementations

lukewagner:
This PR extends the binary/text grammar of import names to allow imports to name not just interfaces, but also *implementations*.  There are a number of ways to refer to an implementation (absolute URL, relative URL, precisely-versioned hierarchical name, version-ranged hierarchical name), so to keep things explicit (and avoid ad hoc string analysis) and support good bindings generation in the relevant cases, these cases are enumerated explicitly.  (In the future, other cases could be added to `importname` as they present themselves.)  Putting this semantic intent directly in the component binary allows it to be reliably interpreted by a variety of tools and runtimes, allowing better tooling interoperability than if a separate file or custom section was used.

This PR doesn't add the new import name cases to Wit yet because there are some interesting related workflow questions that I expect influence how this looks in Wit, suggesting this be a follow-up, once we agree on the target grammar/information.  Also, the expectation is that, for the most part, folks won't need to write this in Wit, these implementation imports will be added late by the toolchain based on build-config files or import maps.

Unfortunately, there doesn't seem to be an "obvious" production we can simply reuse for describing version ranges, so, with a lot of help and ideas from @lann, a basic syntax is proposed built on semver that hopefully looks intuitive enough, but feedback welcome on this of course.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -1891,6 +1891,7 @@ and will be added over the coming months to complete the MVP proposal:
 [`wizer`]: https://github.com/bytecodealliance/wizer
 [`warg`]: https://warg.io
 [SemVerRange]: https://semver.npmjs.com/
+[OCI Registry]: https://github.com/opencontainers/distribution-spec
 
 [Scoping and Layering]: https://docs.google.com/presentation/d/1PSC3Q5oFsJEaYyV5lNJvVgh-SNxhySWUqZ6puyojMi8
 [Future and Stream Types]: https://docs.google.com/presentation/d/1MNVOZ8hdofO3tI0szg_i-Yoy0N2QPU2C--LzVuoGSlE
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -1395,10 +1395,9 @@ The `valid semver` production is as defined by the [Semantic Versioning 2.0]
 spec and is meant to be interpreted according to that specification. The
 `verrange` production embeds a minimal subset of the syntax for version ranges
 found in common package managers like `npm` and `cargo` and is meant to be
-interpreted with the same [semantics][SemVerRange]. (Mostly this is
-interpretation is the obvious lexicographic ordering, but note the particular
-behavior of pre-release tags.)
-
+interpreted with the same [semantics][SemVerRange]. (Mostly this
+interpretation is the usual SemVer-spec-defined ordering, but note the
+particular behavior of pre-release tags.)
 
 For the 3 cases where an `importname` contains a kebab-`name`, that `name` is
 required to be unique within the component's imports so that it can be used as
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -200,7 +200,7 @@ sort           ::= core <core:sort>
                  | type
                  | component
                  | instance
-inlineexport   ::= (export <externname> <sortidx>)
+inlineexport   ::= (export <exportname> <sortidx>)
 string         ::= <core:name>
 name           ::= <label>
                  | [constructor]<label>
@@ -513,8 +513,8 @@ instancedecl  ::= core-prefix(<core:type>)
                 | <type>
                 | <alias>
                 | <exportdecl>
-importdecl    ::= (import <externname> bind-id(<externdesc>))
-exportdecl    ::= (export <externname> bind-id(<externdesc>))
+importdecl    ::= (import <importname> bind-id(<externdesc>))
+exportdecl    ::= (export <exportname> bind-id(<externdesc>))
 externdesc    ::= (<sort> (type <u32>) )
                 | core-prefix(<core:moduletype>)
                 | <functype>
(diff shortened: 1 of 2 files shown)


lann:
After doing some more digging into semver constraints/ranges/comparators I want to point out a subtlety that wasn't obvious to me around matching against pre-release versions:

By a naive application of semver's precedence rules, the range `>=1.0.0 <2.0.0` would match pre-release version `2.0.0-alpha.1`, which seems surprising at best. Cargo/NPM avoid this with a subtle rule:
> If a version has a prerelease tag (for example, 1.2.3-alpha.3) then it will only be allowed to satisfy req if at least one comparator with the same major.minor.patch also has a prerelease tag.
- This means, for example (checking with https://semver.npmjs.com/), for the pre-release version `5.0.0-dev`:
  - these ranges **do not** match: `>=4.0.0 <6.0.0`, `>=4.0.0-alpha <6.0.0-zeta`
  - these ranges **do** match: `>=5.0.0-alpha <6.0.0`, `>=4.0.0 <5.0.0-zeta`
- Another somewhat-surprising consequence of this rule is that `*` never matches any prerelease version

I'm not sure what action to take here beyond being aware of this in tooling implementations. Maybe it would be worthwhile to add a note along the lines of "interpretation of `verrange` is not specified here, but you should probably do what cargo/npm do".

lukewagner:
@lann Good to know!  I see I really didn't even say anything about how to interpret the version ranges, so I added a paragraph on semver [here](https://github.com/WebAssembly/component-model/pull/222/commits/09935344b437c8b3c94c4c9ff07b8a531571ede0).

design/mvp/Explainer.md
```diff
@@ -1394,6 +1391,15 @@ can be interpreted by developer tooling as "registries":
 * a programmatically-created tree data structure (such as the `importObject`
   parameter of [`WebAssembly.instantiate()`])
 
+The `valid semver` production is as defined by the [Semantic Versioning 2.0]
+spec and is meant to be interpreted according to that specification. The
+`verrange` production embeds a minimal subset of the syntax for version ranges
+found in common package managers like `npm` and `cargo` and is meant to be
+interpreted with the same [semantics][SemVerRange]. (Mostly this is
+interpretation is the obvious lexicographic ordering, but note the particular
```
lann, line 1399: Semver _isn't_ compared lexicographically (except for non-numeric pre-release identifiers).

design/mvp/Explainer.md
```diff
@@ -1394,6 +1391,15 @@ can be interpreted by developer tooling as "registries":
 * a programmatically-created tree data structure (such as the `importObject`
   parameter of [`WebAssembly.instantiate()`])
 
+The `valid semver` production is as defined by the [Semantic Versioning 2.0]
+spec and is meant to be interpreted according to that specification. The
+`verrange` production embeds a minimal subset of the syntax for version ranges
+found in common package managers like `npm` and `cargo` and is meant to be
+interpreted with the same [semantics][SemVerRange]. (Mostly this is
```
lann, line 1398:
```suggestion
interpreted with the same [semantics][SemVerRange]. (Mostly this
```

design/mvp/Explainer.md
lukewagner: Hah, right.  [Fixed](https://github.com/WebAssembly/component-model/pull/222/commits/664a9e60b0ee57e8a9843fe76712f7ee740b967b)

alexcrichton (approved):

lukewagner:
Looks like no disagreement on the PR, so almost ready to merge.  However, one more important use case came up recently in the context of registries (and, concretely, sharing component contents via OCI Registries) where one wants to name an implementation purely via its content hash without any additional URL or registry name (and where forcing a URL or registry name would just add a needless hoop to jump through).  Thus, [this](https://github.com/WebAssembly/component-model/pull/222/commits/7944c7b024bf61850fe7f18773b71c0e1e16399e) commit adds a 5th implementation import case that has `integrity` but no URL or registry name.  This is a relatively small addition, so I'll wait for comments until the end of the week to merge.

design/mvp/Explainer.md
```diff
@@ -1295,21 +1296,179 @@ of core linear memory.
...
+When a *particular* implementation is indicated (via URL or registry id),
+`importname` allows the component to additionally specify a cryptographic hash
+of the expected binary representation of the wasm implementation, reusing the
+[`integrity-metadata`] production defined by the W3C Subresource Integrity
+specification. When this hash is present, a component can express its intention
+to reuse another component or core module with the same degree of specificity
+as if the component or core module was nested directly, thereby allowing
+components to factor out common dependencies without compromising runtime
+behavior. When an `integrity` hash is present without a registry name, the
+host must locate the contents using the hash (e.g., using an [OCI Registry]).
```
lann, line 1385: Missing reference?

lukewagner: D'oh, [fixed](https://github.com/WebAssembly/component-model/pull/222/commits/4f286cfe0b27869cdc7cf13ef3783f968eab62be)

lann (approved):

design/mvp/Explainer.md
```diff
@@ -1295,21 +1296,179 @@ of core linear memory.
...
+export     ::= (export <id>? <exportname> <sortidx> <externdesc>?)
+exportname ::= <name>
+             | (interface "<regid>")
+importname ::= <exportname>
+             | <name> (url <string> <integrity>?)
+             | <name> (relative-url <string> <integrity>?)
+             | <name> <integrity>
+             | (locked-dep "<regid>" <integrity>?)
+             | (unlocked-dep "<regidset>")
+regname    ::= <namespace>+<label><projection>*
```
alexcrichton, line 1317: One thing I missed about this earlier is that this impacts syntax in WIT as well since it affects the `(interface ..)` production too, right? Would you be up for updating the `WIT.md` bits with this change?

lukewagner: Oops, yes, thanks.

Issue #229, opened 2023-08-15
Document preview2 vs future features


lann:
There are a number of features that are currently specified in this repo which will not make it into the preview2 implementations. @lukewagner's suggestion is to cut a `preview2` branch from this repo and remove the specifications of those features from that branch.

Starting a list of features that are expected to _not_ be in preview2 implementations:

- [ ] value imports
- [ ] interface subtyping
- [ ] component start functions (?)

lukewagner:
Yep, sounds good to me.  I'll get to this in a bit if noone beats me to it.

alexcrichton:
One thing I'd suggest is to perhaps do the opposite and have a `next` branch instead of a `preview2` branch because the state of preview2 is probably what many newcomers are going to be interested in viewing for a good amount of time and having the right docs by default I think is important (as opposed to a README disclaimer saying "go here for non-future-facing docs")

Otherwise though for what to not be in preview2 on the topic of subtyping specifically I'd propose that we should remove all value-level subtyping, but prefer instance-level subtyping. In other words communication with value types (or interface types) must all exactly match: struct fields in the same order, exact same primitives, exact same variant orderings, etc. I don't believe we have a strong use case for having subtyping at this time at this layer given all the work done on WIT and making it easy to have dependencies, meaning that everyone is likely using the same source of truth. Iteration of APIs over time I think is the next major motivation for subtyping but I think we've still got a good deal of work to do on that so I personally feel it's best for the "next" time-frame.

For instance-level subtyping I think that's important to have which is along the lines of "you asked for 2 functions and I gave you 3 and that's ok". In other words it's ok to drop entire functions or entire types at the instance (or component level) as appropriate. That's relied on and used in tooling today and AFAIK all works well.

ricochet:
If we know an issue is something we'd bucket in the `next` milestone, then we should add it to the [Preview 3](https://github.com/WebAssembly/component-model/milestone/1) milestone. This will help plan the next release regardless of the chosen branching strategy.

For changes that are nice-to-have but not immediately required for key implementations, I recommend we judiciously label `enhancement`.

+1 for the `next` branch approach. If we go this route, we should start with rc tags for 0.2.0 until we're ready for the rubber stamp of 0.2.0.

alexcrichton:
This was discussed at today's component model sync meeting, and the conclusion there was to take a relatively lightweight approach to this for now which can optionally evolve into more process later on. The action items are:

[...]

PR #233, opened 2023-08-17
Propose `enum` be a subtype of `flags`.

sunfishcode:
I was exploring some hypothetical use cases where it'd be nice to migrate from an `enum`, "pick one of these", to a `flags`, "pick any number of these".

For example given these types:

```
enum berry-enum {
   blueberry,
   strawberry,
   raspberry,
}

flags berry-flags {
   blueberry,
   strawberry,
   raspberry,
}
```

The proposal is `berry-enum <: berry-flags`.

I know that the subtyping rules aren't currently a focus, and that they may change when the time does come to focus on them, but I wanted to add this idea to the list so that it's considered.

### DIFF
--- a/design/mvp/Subtyping.md
+++ b/design/mvp/Subtyping.md
@@ -17,6 +17,7 @@ But roughly speaking:
 | `option`                  | `T <: (option T)` |
 | `expected`                | `T <: (expected T _)` |
 | `union`                   | `T <: (union ... T ...)` |
+| `enum`                    | an `enum` type is a subtype of a `flags` type if each case has a corresponding flag |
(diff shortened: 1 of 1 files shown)


lukewagner:
That's an interesting idea, let's tag it as an enhancement and consider it when we're defining/implementing value-level subtyping properly.  As #229 suggests, in a preview2 timeframe, we're starting without any value-level subtyping, so probably we'll want to consider in detail the forms of subtyping we do want to allow as a post-preview2 batch.

lukewagner:
I'll close for now, since subtyping is generally removed from the C-M repo (#255), but we can revisit when we properly treat subtyping in the future.  (Feel free to file a new issue to document the idea that we can tag appropriately.)

PR #237, opened 2023-08-22
Remove 'union' types

### LINKED ISSUE #236: WIT: consider dropping `union`s
Unions are uniquely challenging to implement in bindings because their cases do not have kebab-case names. See for example the wit-bindgen Rust generator's awkward solution to this: https://github.com/bytecodealliance/wasmtime/blob/367bdc8c66df62d6854966a90564b42692a3f023/crates/wit-bindgen/src/rust.rs#L317

Rather than expecting each bindings language to come up with a generic naming scheme to handle `union`s, I think we should put the onus on WIT authors to select reasonable case names on an equivalent `variant`.

A quick [Github search](https://github.com/search?q=path%3A%2F.wit%24%2F+%2F%5Cbunion%5Cb%2F&type=code) (and review of WASI) suggests that `union`s are not currently in use outside of tests and samples.

lann:
Closes #236


lukewagner (approved):
Thanks!  I'll wait a few days for visibility and then merge if it's quiet.

PR #241, opened 2023-09-06
Fix wrong item number in Goals.md

sachaos:
I found a wrong item number and corrected it.
The markdown display is not affected. :)


lukewagner:
Wow, good catch (that's been there a while :), thanks!

PR #243, opened 2023-09-06
Fix small typo in UseCases.md

sachaos:
I found another small typo and corrected it. :)
I apologize for sending multiple small PRs. :bow:

### DIFF
--- a/design/high-level/UseCases.md
+++ b/design/high-level/UseCases.md
@@ -197,7 +197,7 @@ use cases that require additional features:
    the client. The component developer can switch languages, toolchains or memory
    representations in the future without breaking existing clients.
 
-The above use cases roughly correspond to the use cases of an [RPC] framework,
+The above use cases roughly correspond to the use cases of a [RPC] framework,
 which have similar goals of crossing language boundaries. The major difference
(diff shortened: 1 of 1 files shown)


design/high-level/UseCases.md
```diff
@@ -197,7 +197,7 @@ use cases that require additional features:
    the client. The component developer can switch languages, toolchains or memory
    representations in the future without breaking existing clients.
 
-The above use cases roughly correspond to the use cases of an [RPC] framework,
```
jeffparsons, line 200: I think this is actually fine as-is. I don't know what the big style guides have to say about it, but most writing I've seen uses the pronunciation of the initialism/acronym rather than its expansion to decide between "a" or "an".

So in this case, the starting sound of "RPC" would be "arr" (what's a pirate's favourite way to communicate between processes?), making it "an RPC" / "an arr pee see".

jeffparsons: Note, though, that I'm just an interested observer — not a maintainer of this project. So take my position with a grain of salt!

sachaos: @jeffparsons Oh, Thank you! You are right! I'll close this PR. :)

PR #244, opened 2023-09-08
Point users towards the documentation

merlijn-sebrechts:
The repository and the explainer are the second and forth result on Google when searching for "wasm component model". Since a lot of users actually want a high-level explanation instead of the design spec, we should point users to the docs instead.

~greetings from Componentize The World, btw!


lukewagner:
Great idea!

PR #245, opened 2023-09-11
add use-path grammar

cardoso:
This seems to be missing from #174 . I just took a rough look at wit-parser and this seems to be what it's doing currently. 

cc @Mossaka @alexcrichton

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -927,11 +927,8 @@ include my-world-2
 ```
 
 ```ebnf
-include-item ::= 'include' use-path
-               | 'include' use-path 'with' '{' include-names-list '}'
-
-use-path ::= id
-           | id ':' id '/' id
+include-item ::= 'include' interface
+               | 'include' interface 'with' '{' include-names-list '}'
 
 include-names-list ::= include-names-item
                      | include-names-list ',' include-names-item
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -880,18 +880,18 @@ into the scope of the current file and/or rename interfaces locally for
 convenience:
 
 ```ebnf
-toplevel-use-item ::= 'use' interface ('as' id)?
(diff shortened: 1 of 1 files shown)


Mossaka (approved):
Thanks! Lgtm

alexcrichton (commented):
Thanks! Could `use-path` reuse the preexisting `interface` production perhaps? That may require renaming `interface`, however, since `include` refers to worlds instead of interfaces.

cardoso:
@alexcrichton can it take the version as well? The parser accepts it, but I didn't see ui tests for it. If the parser is right, I can improve this and add more tests there.

cardoso:
@alexcrichton can it take the version as well? The parser accepts it, but I didn't see ui tests for it. If the parser is right, I can improve this and add more tests there.

cardoso:
Okay, so it seems they're the same:

```wit
interface ::= id
            | id ':' id '/' id ('@' valid-semver)?
```

 If this is it, then I'll open the PR for the parser once this is merged.

cardoso:
It does seem confusing though since it should refer to worlds, but I'll leave it to you.

design/mvp/WIT.md
```diff
@@ -927,8 +927,8 @@ include my-world-2
 ```
 
 ```ebnf
-include-item ::= 'include' use-path
-               | 'include' use-path 'with' '{' include-names-list '}'
+include-item ::= 'include' interface
+               | 'include' interface 'with' '{' include-names-list '}'
```
cardoso, line 931: Would it make more sense to rename interface to use-path instead?

alexcrichton: I think so yeah, but either way works too

cardoso:
Thanks @alexcrichton . I think it reads better now (and uses the same naming as wit-parser).

alexcrichton (approved):

lukewagner:
Thanks!

PR #248, opened 2023-09-19
Change encoding of WIT definitions into Component Model types

lukewagner:
This PR tweaks the encoding of WIT definitions into component-level types as described in WIT.md.  The primary motivation here is to avoid having a special kind of "WIT package" and rather to allow WIT definitions to just be normal (type) definitions inside normal components that can be exported and named like anything else.  This allows more-unified registry tooling and also it allows a bit more flexibility for packages to contain both implementation and interface definitions.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -21,7 +21,7 @@ additionally from other packages through IDs.
 
 This document will go through the purpose of the syntactic constructs of a WIT
 document, a pseudo-formal [grammar specification][lexical-structure], and
-additionally a specification of the [binary format][binary-format] of a WIT
+additionally a specification of the [package format][package-format] of a WIT
 package suitable for distribution.
 
 [IDL]: https://en.wikipedia.org/wiki/Interface_description_language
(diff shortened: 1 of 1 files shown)


design/mvp/WIT.md
```diff
@@ -1370,140 +1370,167 @@ record bar2 {
...
       (export "run" (func))
     ))
-    (export (interface "local:demo/the-world") (component (type $the-world)))
+    (export (interface "local:demo/the-world") (type (eq $the-world')))
   ))
-  (export (interface "local:demo/wit") (type $demo))
+  (export "the-world" (type $the-world))
 )
 ```
+In the current version of WIT, `world`s cannot contain `use` and thus the outer
```
alexcrichton, line 1524: Oh technically this isn't quite right, you can do:

```wit
interface foo {
  type t = u8
}

world bar {
  use foo.{t}
  import x: func() -> t
}
```

alexcrichton (approved):
All looks good to me, thanks for writing this up!

In terms of rolling this out an idea I'd have for a plan would be basically the same as https://github.com/WebAssembly/component-model/pull/249 where both encoding formats are implemented for a short period of time. First with the old one as the default, then the new one as the default, then only the support for the new one.

design/mvp/WIT.md
lukewagner: Ah, thanks for pointing that out.  So does that turn into an import of `foo` (that goes right before the import of `x`)?

lukewagner: Cool, gotcha.  Would it make sense to have the import of `foo` show up in the outer wrapping component instead?  In the short term, this would allow the rendered WIT to display a `use` instead of an `import`.  In the longer term, when one world can import another world, it would be necessary to be able to tell the difference between a `use` (which has the same default-sharing behavior of `use` in `interface`s) and an explicit `import` (which would always stay put).

alexcrichton: Oh right now the world I mentioned "elaborates" to 

```wit
world bar {
  use foo.{t}
  import foo
  import x: func() -> t
}
```

so there's not much of a distinction between `use` and `import`, it's more that `use` is an `alias` of sorts. In that sense I don't think supporting this `use`-in-`world` requires any special treatment since it can show up as an import in the exported component type? So no need for it to be in the outer wrapping component, only in the actual component itself.

lukewagner: Yeah, thinking about it again, that seems reasonable.  The use case I was thinking about where you don't want that `import foo` (only the `use`) involves not just high-order components (worlds-importing-worlds), but also partially-bound components, which would likely need a different type syntax anyhow and so there's no need to change anything now.

lukewagner: [Fixed](https://github.com/WebAssembly/component-model/pull/248/commits/7842b8ff034d86c6bea444a4e6b91a04eddfa3ba)

design/mvp/WIT.md
```diff
@@ -1370,140 +1370,166 @@ record bar2 {
...
+      ))
+    ))
+  ))
+  (type (export "namespace") (component
+    (import (interface "local:demo/types") (instance $types
+      (export "file" (type (sub resource)))
+    ))
+    (alias export $types "file" (type $file))
+    (type (export (interface "local:demo/namespace")) (instance
+      (export "open" (func (param "name" string) (result (own $file))))
```
elliottt, line 1439: This version doesn't parse currently, but I think the following should be equivalent. Additionally, this example doesn't validate, as the handling of instances exported via `(type (eq ...))` ignores some special case handling that otherwise allows instance exports.
```lisp
(component
  (type $types (component
    (type $types' (instance
      (export $file "file" (type (sub resource)))
      (export "[method]file.read" (func
        (param "self" (borrow $file)) (param "off" u32) (param "n" u32)
        (result (list u8))
      ))
      (export "[method]file.write" (func
        (param "self" (borrow $file))
        (param "bytes" (list u8))
      ))
    ))
    (export (interface "local:demo/types") (type (eq $types')))
  ))
  (export "types" (type $types))
  (type $namespace (component
    (import (interface "local:demo/types") (instance $types
      (export "file" (type (sub resource)))
    ))
    (alias export $types "file" (type $file))
    (type $interface' (instance
      (export "open" (func (param "name" string) (result (own $file))))
    ))
    (export (interface "local:demo/namespace") (type (eq $interface')))
  ))
  (export "namespace" (type $namespace))
)
```

alexcrichton:
@elliottt, @lukewagner, and I talked a bit about this new encoding scheme and have reached the conclusion that a proposed tweak would be to change export-of-instance-type to become export-of-instance. The rationale is not about the design per se but instead relates to https://github.com/bytecodealliance/wasm-tools/issues/1253 where currently wasm-tools doesn't validate export-of-instance-type correctly when the type has resources inside of it. Moving to export-of-instance would sidestep the questions raised there and ideally buy more runway to figuring out how to correctly implement this in wams-tools.

---

As an orthogonal point, though, @elliottt and I also realized that with this encoding scheme then an empty WIT package with no interfaces or worlds can't be encoded. Basically the package name and namespace is lost and there's nowhere to put it. Shall we perhaps consider empty packages invalid from a parser level in that case?

lukewagner:
Yes, that changes sounds good to me; it also has the nice property of being more symmetric with the imports encoding `use`, which, quite coincidentally, makes the encoding of the idea in #262 pretty clean.  [Updated](https://github.com/WebAssembly/component-model/pull/248/commits/23e2cc50712a40f419e907ff121f5743be91b132)

design/mvp/WIT.md
lukewagner: [Fixed](https://github.com/WebAssembly/component-model/pull/248/commits/23e2cc50712a40f419e907ff121f5743be91b132) in response to Alex's comment below.

lukewagner:
Oh, and to your second point, that's a good point, and one I wrestled with too.  Although it's a bit subtle, I think what we can say is that the point of the interface names in the encoding is *not* to name the package, but, rather, to name the definitions in a way that is later validated to be consistent with the *externally-assigned* name of the package at the point of publication.  So, e.g., we could union two packages into a single component binary; the result wouldn't be publishable as a package (b/c the package-name-consistency validation would fail), but it would be useful as a single-file collection of "all the type info", which I think is useful for some of the future non-publishing workflows we talked about earlier.

If we wanted, we could define a "package metadata" custom section that would include author, email, description, etc, and this might be a natural place for the package name to canonically reside.

design/mvp/WIT.md
```diff
@@ -1494,17 +1494,16 @@ is encoded as:
...
     ))
-    (export (interface "local:demo/foo") (type (eq $foo')))
   ))
   (export "foo" (type $foo))
 )
 ```
 
 Worlds are encoded similarly to interfaces, but replace the inner exported
-instance-type with an inner exported *component*-type. For example, this WIT:
+instance-type with an inner exported *component*. For example, this WIT:
```
jeffparsons, line 1506: Should this say "instance with an inner exported..."?

lukewagner: Yes, thanks, good catch!  [Fixed](https://github.com/WebAssembly/component-model/pull/248/commits/15a9af9128a1b9936cfc90bd83a554d53e1ffe1c)

PR #249, opened 2023-09-19
Use semicolons as delimiters in WIT text format

### LINKED ISSUE #142: wit feature request: add explicit item delimiters
I'd like to suggest adding explicit delimiters between interface/world items. For example, this could be semicolons:
```wit
interface example {
  a: func();
  b: func() -> ();
  c: func() -> t;
  type t = u32;
}
```

Reasoning:
I was experimenting with textmate grammars for WIT (tmLanguage is used by vscode) and discovered that it's difficult to add full-featured syntax highlighting because of tm limitations. In order to add "scope" to some part of the document, tm has to know exactly where it starts and ends. This works for interfaces (they start with `{` and end with `}`), however it causes problems with interface children. For example, functions can span multiple lines (so we can't use `\n` as end), they can have return type missing or specified without parenthesis (so we can't use ')').
Proto format [uses](https://github.com/zxh0/vscode-proto3/blob/f808059028c870ffd0b38babdfafcd49752dee1e/syntaxes/proto3.tmLanguage.json#L258) semicolons for this.

There are workarounds: for instance it's possible to ignore `meta` and other conventional scopes and highlight keywords only instead. However, this will make writing wit files uncomfortable :( 

VS Code supports "semantic tokens" which provides better highlighting for lsp-enabled languages. However, it's difficult to say if this is enough without actually implementing both tm and lsp

alexcrichton:
This commit commit is an implementation of #142 where semicolons are now required as delimiters between items in the WIT text format. All items in the WIT format are now delimited with either curly braces (`{}`) or semicolons except for the `package` statement where it subjectively felt a bit weird to require a semicolon. I've updated the various examples in `WIT.md` as an example of the new syntax.

My plan on implementing this would be along the lines of:

* Implement the semicolon syntax in `wit-parser`
* Add a parser mode which requires semicolons. This means that the same `wit-parser` crate can either or either not require semicolons.
* Update all tests in the `wasm-tools` repository to require semicolons.
* Publish `wit-parser` and `wasm-tools`, integrating the semicolon-supporting-mode into all existing tools.
* Wait for Wasmtime to get published with this support. At this point everything in the ecosystem should have a point where semicolons are optionally supported.
* Remove the parser mode which doesn't require semicolons, meaning semicolons are now required.
* Push this update through the tooling, fixing any issues that arise.

The hope is to create a period of time where both syntax forms are accepted. This provides a transitionary means from one syntax to the other while proposals are updated. This transitionary period is finite in length, however.

Closes #142

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -77,7 +77,7 @@ An example of an interface is:
 package local:demo
 
 interface host {
-  log: func(msg: string)
+  log: func(msg: string);
 }
 ```
 
@@ -101,7 +101,7 @@ and [function][functions] definitions. For example:
 package wasi:filesystem
 
 interface types {
-  use wasi:clocks.wall-clock.{datetime}
+  use wasi:clocks.wall-clock.{datetime};
 
   record stat {
     ino: u64,
@@ -110,7 +110,7 @@ interface types {
     // ...
   }
 
-  stat-file: func(path: string) -> result<stat>
+  stat-file: func(path: string) -> result<stat>;
 }
 ```
 
@@ -139,10 +139,10 @@ package local:demo
 
 world my-world {
   import host: interface {
-    log: func(param: string)
+    log: func(param: string);
   }
 
-  export run: func()
+  export run: func();
 }
 ```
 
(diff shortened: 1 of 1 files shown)


lukewagner (approved):
Thanks for writing this up!  Agreed this is probably the right answer when we think long-term.  lgtm, but I'd be interested to wait for a bit to collect other feedback.

alexcrichton:
I've got a PR for an implementation of this when it's ready at https://github.com/bytecodealliance/wasm-tools/pull/1212 which can also serve as a means to evaluate the aesthetics of the change via reading all the changes to tests. (aesthetics improved IMO)

PR #250, opened 2023-09-20
Add some language identifier to enable syntax highlighting

sachaos:
I've added some language identifiers to code block for enabling syntax highlighting. :)

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -16,7 +16,7 @@ and validation will be present in the [formal specification](../../spec/).
 ## Component Definitions
 
 (See [Component Definitions](Explainer.md#component-definitions) in the explainer.)
-```
+```ebnf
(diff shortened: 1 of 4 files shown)


ricochet:
Thank you for adding this! 

We have a little bit of a chicken and egg problem where WIT is defined by this repo but the grammar is specified in [bytecodealliance/vscode-wit](https://github.com/bytecodealliance/vscode-wit). The linguist tool used by GitHub for code blocks rendering in md pulls the grammar from that repo. I think this can be OK so long as it is clear that there will be a lag between incoming changes to WIT syntax and when rendering for the new syntax is supported.

I noticed that this change does not add language identifiers for all code blocks, e.g. there are many more blocks in `WIT.md`.

lukewagner:
lgtm, thanks!

Issue #253, opened 2023-09-25
Inferring import kind from string syntax


alexcrichton:
I've been reviewing https://github.com/bytecodealliance/wasm-tools/pull/1146 which is the start of the implementation of implementation imports for the component model and it's raising questions about internal details which I wanted to raise to the design level. Before this PR only two forms of imports were supported for components:

```wasm
(import "foo" (func ...))
(import (interface "foo:bar/baz") (func ...))
```

With the [recently specified](https://github.com/WebAssembly/component-model/pull/222) implementation imports the above PR is adding support for new forms of imports:

```wasm
;; from before
(import "foo" (func ...))
(import (interface "foo:bar/baz") (func ...))

;; new
(import "foo" (integrity "xx") (func ...))
(import "foo" (url "xx") (func ...))
(import "foo" (relative-url "xx") (func ...))
(import (locked-dep "foo:bar/baz") (func ...))
(import (unlocked-dep "foo:bar/baz") (func ...))
```

Throughout these refactorings, and previously when `(interface ...)` imports were added, the internal data structures of much of the tooling around the component model ignores this metadata and instead thinks of imports as a map of "string to thing". This is additionally done for instantiation where instantiation arguments are provided as a list of "string to thing". Each import form then has a canonical string associated with it that is used internally. This canonical string is what disallows overlap between imports, but it additionally loses context like `(url ...)` and `(integrity ...)` which I believe is ok for the current use cases of the tooling (e.g. the `url` or `integrity` doesn't affect validation)

```wasm
(import "foo" (func ...))                         ;; name = "foo"
(import (interface "foo:bar/baz") (func ...))     ;; name = "foo:bar/baz"
(import "foo" (integrity "xx") (func ...))        ;; name = "foo"
(import "foo" (url "xx") (func ...))              ;; name = "foo"
(import "foo" (relative-url "xx") (func ...))     ;; name = "foo"
(import (locked-dep "foo:bar/baz") (func ...))    ;; name = "foo:bar/baz"
(import (unlocked-dep "foo:bar/baz") (func ...))  ;; name = "foo:bar/baz"
```

[...]

PR #255, opened 2023-09-26
Clarify what's in Preview 2

lukewagner:
This PR implements the idea in #229 of qualifying the parts of the current proposal that are not included in the upcoming 'Preview 2' stability milestone.  Subtyping wasn't really written, so I simply removed it and added notes suggesting it's possible addition instead.  I'm not wed to the particular emojis so alternatives welcome.


README.md
```diff
@@ -12,10 +12,9 @@ reference interpreter and test suite.
 
 ## Milestones
 
-The Component Model is being incrementally developed and stabilized as part of
-the overall [WASI preview process]. The current contents of this repo reflect
-the "Preview 2" milestone. The "Preview 3" milestone will be primarily
-concerned with the addition of [async support].
+The Component Model is currently being incrementally developed and stabilized
+as part of [WASI Preview 2]. The subsequent "Preview 3" milestone will be
+primarily concerned with the addition of [async support].
```
vados-cosmonic, line 17: What do you think about making this a list with checkmarks/or strikeouts? I'm envisioning something like:

```markdown
## Milestones

Current and future work is done towards milestones colloquially called "Preview" versions:

- [x] Preview 1
  - [x] WASI interface design & implementation
- [x] Preview 2
  - [x] Component Model design & implementation
- [ ] Preview 3
  - [ ] Async support
```
(sorry if a bit to the above is wrong/mis-stated!)

This might be a decent easy-to-digest way to give an overview of the milestones the project has been through and where it's going.

Could even use emojis like :ship: versus :building_construction: / :construction:  to show things that are in-progress versus shipped?

design/mvp/Explainer.md
```diff
@@ -1313,7 +1325,8 @@ importname ::= <exportname>
              | <name> <integrity>
              | (locked-dep "<regid>" <integrity>?)
              | (unlocked-dep "<regidset>")
-regname    ::= <namespace>+<label><projection>*
+regname    ::= <namespace><label><projection>?
```
alexcrichton, line 1328: I think for this one `<projection>` is required?

design/mvp/Binary.md
```diff
@@ -207,7 +209,7 @@ defvaltype    ::= pvt:<primvaltype>                       => pvt
                 | 0x69 i:<typeidx>                        => (own i)
                 | 0x68 i:<typeidx>                        => (borrow i)
 labelvaltype  ::= l:<label> t:<valtype>                   => l t
-case          ::= l:<label> t?:<valtype>? r?:<u32>?       => (case l t? (refines case-label[r])?)
+case          ::= l:<label> t?:<valtype>? ignored?:<u32>? => (case l t?)
```
alexcrichton, line 212: For this if we want to hae an ignored payload I think it would be best to require this to be 0x00, as otherwise any u32 is valid here today and we wouldn't be able to hijack the bytes later

alexcrichton (approved):

design/mvp/Explainer.md
lukewagner: For the `(interface <regid>)` case, yes, but not for the `(*-dep <regid*>)` cases.  [Fixed](https://github.com/WebAssembly/component-model/pull/255/commits/e6b5977a5f50e79123b8c6bcf93cba81a461d120) by splitting `regname` into separate `id*` and `pkg*` cases.  PTAL

README.md
lukewagner: I think what we're doing here is a finer-granularity: we're circumscribing the description of the individual "features" that make up a single milestone, with the idea that the un-emoji'd stuff is the "current" milestone (currently: Preview 2) and the emoji'd stuff is the "next" milestone (currently: Preview 3).  But once we hit that next milestone, all that shifts forward (we drop the emojis and merge the relevant rules that can now be simplified) and start afresh.  Independently, the existing "TODO" section could probably improved/renamed in-place to document the milestones kindof like you're describing.

design/mvp/Explainer.md
alexcrichton: Ah ok. That's one where the implementation hasn't finished in https://github.com/bytecodealliance/wasm-tools/pull/1146 as it led to https://github.com/WebAssembly/component-model/issues/253. I've been wary of dropping `/foo` as it has interactions with WIT that I dont' think have been fully bottomed out, but if this change is only affecting the `{locked,unlocked}-dep` forms of imports that I think is ok since those don't interact with WIT at all right now.

alexcrichton: As in, the updated version looks good to me 👍

Issue #247, opened 2023-09-19
Special FP handling in component interfaces?


penzn:
Came across what appears a mandate to sanitize all `NaN` values loaded from memory to a specific value:

https://github.com/WebAssembly/component-model/blob/673d5c43c3cc0f4aeb8996a5c0931af623f16808/design/mvp/CanonicalABI.md?plain=1#L466-L488
https://github.com/WebAssembly/component-model/blob/673d5c43c3cc0f4aeb8996a5c0931af623f16808/design/mvp/Explainer.md?plain=1#L554-L555
A [presentation](https://sunfishcode.github.io/RelaxedMathMode.pdf) on in SIMD subgroup admitted that this actually has a substantial overhead on x86 (see slide 7 in the PDF). Can we get a bit more clarity on this, is this proposed to be implemented in browsers when they support Component Model? Is it going to be required in WASI?

penzn:
It depends on the output, I suppose. For a component returning a large array of floats the whole buffer needs to be processed element by element, even if there are no NaNs in it. Most programming environments don't care about exact NaN bits, notably JS and host languages that would compile to wasm, because as far as arithmetic ops are concerned non-signaling NaN values are equivalent (yes, there is NaN-boxing, but that won't be applicable to interface boundaries).

lukewagner:
> notably JS

JS ends up caring a fair amount since several JS engines internally use the [NaN-boxing](https://wingolog.org/archives/2011/05/18/value-representation-in-javascript-implementations) value representation, which causes them to have to canonicalize *all* incoming float values.  That being said, that could happen as part of the JS-specific bindings, instead of as part of the Canonical ABI, so that's not a hard requirement or anything.  But it is a notable example where non-canonical NaNs are a problem.

lukewagner:
Thinking about this some more, it does seem like the cost for a `list<f32>` (and perhaps, one day, `list<f16>`) could likely end up being significant, whereas the practical benefit is fairly limited to just languages doing NaN-boxing, which are used to canonicalizing at external boundaries anyways.  So maybe we should remove this mandatory canonicalization.

@sunfishcode had a good idea though: instead of *forcing* canonicalization, we could instead specify that, at component-to-component boundaries, canonicalization *might* (non-deterministically) happen and thus interfaces must not rely on specific NaN payloads to be faithfully transmitted across component boundaries, thereby *allowing* language implementations to canonicalize if they need to (which JS would).  This seems like a pretty good compromise.

[...]

PR #260, opened 2023-09-27
Don't force deterministic NaN canonicalization

lukewagner:
After revisiting this design choice in #247, this PR changes the Canonical ABI to not force canonicalization of NaNs, so that hosts can just use the raw `f32`/`f64` bit-patterns.  Because some languages and host implementation techniques (especially JS, both as host and as guest) will inevitably end up doing some amount of NaN-canonicalization, this PR explicitly includes this allowed non-determinism in the Canonical ABI lifting and lowering routines.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -436,8 +436,8 @@ def load(cx, ptr, t):
     case S16()          : return load_int(cx, ptr, 2, signed=True)
     case S32()          : return load_int(cx, ptr, 4, signed=True)
     case S64()          : return load_int(cx, ptr, 8, signed=True)
-    case Float32()      : return canonicalize32(reinterpret_i32_as_float(load_int(cx, ptr, 4)))
-    case Float64()      : return canonicalize64(reinterpret_i64_as_float(load_int(cx, ptr, 8)))
+    case Float32()      : return maybe_scramble_nan32(reinterpret_i32_as_float(load_int(cx, ptr, 4)))
+    case Float64()      : return maybe_scramble_nan64(reinterpret_i64_as_float(load_int(cx, ptr, 8)))
     case Char()         : return convert_i32_to_char(cx, load_int(cx, ptr, 4))
     case String()       : return load_string(cx, ptr)
     case List(t)        : return load_list(cx, ptr, t)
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):

design/mvp/CanonicalABI.md
```diff
@@ -463,28 +463,51 @@ def convert_int_to_bool(i):
...
 
-For reasons [given](Explainer.md#type-definitions) in the explainer, floats are
-loaded from memory and then "canonicalized", mapping all Not-a-Number bit
-patterns to a single canonical `nan` value.
+Lifting and lowering float values may (from the component's perspective)
+non-deterministically modify the sign and payload bits of Not-A-Number (NaN)
+values, reflecting the practical reality that different languages, protocols
+and CPUs have different effects on NaNs. Although this non-determinism is
+expressed in the Python code below as generating a "random" NaN bit-pattern,
+native implementations do not need to literally generate a random bit-pattern;
```
penzn, line 471: > native implementations do not need to literally generate a random bit-pattern;
they may canonicalize to an arbitrary fixed NaN value

Can this just side simply be omitted, why is there a need to do anything with NaNs at all outside of 'deterministic mode'?

penzn (commented):
My bad, I let this slide a bit, and then was out sick when the change got merged.

I am not sure why define any additional semantics outside deterministic mode.

design/mvp/CanonicalABI.md
sunfishcode: If we don't say anything, that would be an effective guarantee that NaN bits are preserved across interfaces. We want to be able to transmit NaN values through protocols which don't preserve the bits.

penzn: I am not sure it is clearer if we say that NaNs are "canonicalized to an arbitrary value", but more importantly, won't saying nothing be just equivalent to core semantics?

sunfishcode: Core semantics always preserve NaN bits in call arguments and return values. We want to be able to route component-interface calls through protocols and source languages which don't preserve NaN bits, and that motivates saying something more here.

penzn: Can you elaborate on how they won't preserve NaN bits (in non "deterministic" mode), one NaN value would be passed by the caller, but a different NaN value received by the callee? What would be the protocol where that would happen?

sunfishcode: JSON is a protocol that doesn't have `NaN` as explicit syntax, but there are common conventions to encode NaN as the string "NaN" or as `null`, such as the [JSON mapping of protobufs](https://protobuf.dev/programming-guides/proto3/#json), n which case, NaN bits would not be preserved.

JavaScript is a source language in which floating-point values which don't preserve NaN bits.

penzn: "Not preserving NaN bits" is different from "canonicalize to unknown value". I think it would just suffice to say that NaN bits are not preserved across interface calls/returns, which would also cover the deterministic mode as well.

sunfishcode: I don't see how saying "NaN bits are not preserved" practically differs from "canonicalize to unknown value". They seem to me to have the same effect.

lukewagner: "Not preserving NaN bits" is definitely the intention (and what the words say in CanonicalABI.md).  In the formal spec, this would be modeled with an evaluation rule that non-deterministically turns a NaN into any other valid NaN, and the closest way I could think of to emulate this in Python is using random, which seems to capture the spirit of "anything could pop out on the other side".

PR #263, opened 2023-10-16
Move structured import/export name information into the import/export string

lukewagner:
This PR fleshes out the idea in #253 to move the structured information of imports/exports into the import/export name string, maintaining the same validation structure, but now inside a single quoted strong.  The motivation for this change is to have a simple canonical string for all import/export names that contains all the structured information, without requiring host and toolchain implementations to define their own ad hoc string-serialization conventions.  The rationale here is basically the same as the earlier design choice to put method/constructor/etc annotations inside the name strings, with the net effect of making the way externally-meaningful name information is represented in components somewhat more regular.

Some specific details:
* This PR is intended to be backwards-compatible at the binary format level, assuming that the 5 new "implementation import" cases added by #222 haven't been implemented or used yet.  This binary compatibility is achieved by allowing a now-"junk" byte that was previously used to discriminate `(interface ...)` imports.
* The PR moves the existing kebab-name/label grammar to the "Imports and Exports" section so that the whole group can be defined in one block and also use a lexical grammar that's more explicit about whitespace and metacharacters.  This ended up surfacing a spec-bug in which labels weren't being quoted.
* As there are now a fair number of "<___name>"s in the grammar, it seemed odd to have `<name>` refer to the kebab-name case.  Moreover, `<name>` wasn't really a "kebab-name" anyway, since it also contains method/constructor/static annotations (with embedded square brackets and periods).  Thus, this PR proposes to rename this production to `<plainname>`, attempting to distinguish plain names from all the other kinds of names that carry more semantic information, but happy to hear other thoughts on how to call this.
* Embedded URLs and registry names are deliminted by angle-brackets (not parens, as discussed in #253), mostly because (1) parens are allowed in URLs, (2) double-quotes require extra escaping in WAT, which seems like it'll be annoying, (3) other than double-quotes, angle-brackets are the [RFC-recommended](https://www.rfc-editor.org/rfc/rfc3986#appendix-C) way to delimit URLs.  The other implementation imports follow the same scheme as URLs just for symmetry.

Copying from this PR, an example showing the new import/export names is:
```wasm
(component
  (import "custom-hook" (func (param string) (result string)))
  (import "wasi:http/handler" (instance
    (export "request" (type $request (sub resource)))
    (export "response" (type $response (sub resource)))
    (export "handle" (func (param (own $request)) (result (own $response))))
  ))
  (import "url=<https://mycdn.com/my-component.wasm>" (component ...))
  (import "relative-url=<./other-component.wasm>,integrity=<sha256-X9ArH3k...>" (component ...))
  (import "locked-dep=<my-registry:sqlite@1.2.3>,integrity=<sha256-H8BRh8j...>" (component ...))
  (import "unlocked-dep=<my-registry:imagemagick@{>=1.0.0}>" (instance ...))
  (import "integrity=<sha256-Y3BsI4l...>" (component ...))
  ... impl
  (export "wasi:http/handler" (instance $http_handler_impl))
  (export "get-JSON" (func $get_json_impl (result string)))
)
```

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -310,10 +310,8 @@ in the explainer.)
 ```ebnf
 import      ::= en:<importname'> ed:<externdesc>                     => (import in ed)
 export      ::= en:<exportname'> si:<sortidx> ed?:<externdesc>?      => (export en si ed?)
-importname' ::= <junk> len:<u32> in:<importname>                     => in  (if len = |in|)
-exportname' ::= <junk> len:<u32> en:<exportname>                     => en  (if len = |en|)
-junk        ::= 0x00
-              | 0x01
+importname' ::= 0x00 len:<u32> in:<importname>                       => in  (if len = |in|)
+exportname' ::= 0x00 len:<u32> en:<exportname>                       => en  (if len = |en|)
 ```
 
 Notes:
@@ -343,8 +341,6 @@ Notes:
 * `<valid semver>` is as defined by [https://semver.org](https://semver.org/)
 * `<integrity-metadata>` is as defined by the
   [SRI](https://www.w3.org/TR/SRI/#dfn-integrity-metadata) spec.
-* The `junk` byte is leftover from a previous iteration and will be removed at
-  the next breaking binary format change.
 
 ## Name Section
 
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -73,6 +73,7 @@ instance            ::= ie:<instanceexpr>                                  => (i
 instanceexpr        ::= 0x00 c:<componentidx> arg*:vec(<instantiatearg>)   => (instantiate c arg*)
                       | 0x01 e*:vec(<inlineexport>)                        => e*
 instantiatearg      ::= n:<string>  si:<sortidx>                           => (with n si)
+string              ::= s:<core:name>                                      => s
 sortidx             ::= sort:<sort> idx:<u32>                              => (sort idx)
 sort                ::= 0x00 cs:<core:sort>                                => core cs
                       | 0x01                                               => func
@@ -81,16 +82,6 @@ sort                ::= 0x00 cs:<core:sort>                                => co
                       | 0x04                                               => component
                       | 0x05                                               => instance
 inlineexport        ::= n:<exportname> si:<sortidx>                        => (export n si)
-string              ::= s:<core:name>                                      => s
-name                ::= len:<u32> n:<name-chars>                           => n (if len = |n|)
-name-chars          ::= l:<label>                                          => l
-                      | '[constructor]' r:<label>                          => [constructor]r
-                      | '[method]' r:<label> '.' m:<label>                 => [method]r.m
-                      | '[static]' r:<label> '.' s:<label>                 => [static]r.s
-label               ::= w:<word>                                           => w
-                      | l:<label> '-' w:<word>                             => l-w
-word                ::= w:[0x61-0x7a] x*:[0x30-0x39,0x61-0x7a]*            => char(w)char(x)*
-                      | W:[0x41-0x5a] X*:[0x30-0x39,0x41-0x5a]*            => char(W)char(X)*
 ```
 Notes:
 * Reused Core binary rules: [`core:name`], (variable-length encoded) [`core:u32`]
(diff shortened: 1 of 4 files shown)


design/mvp/Binary.md
```diff
@@ -327,20 +308,12 @@ flags are set.
...
-              | 0x06 p:<pkgidset'>                                   => (unlocked-dep p)
-pkgid'      ::= len:<u32> p:<pkgid>                                  => "p" (if len = |p|)
-pkgidset'   ::= len:<u32> p:<pkgidset>                               => "p" (if len = |p|)
-integrity'  ::= len:<u32> im:<integrity-metadata>                    => (integrity "im") (if len = |im|)
+import      ::= en:<importname'> ed:<externdesc>                     => (import in ed)
+export      ::= en:<exportname'> si:<sortidx> ed?:<externdesc>?      => (export en si ed?)
+importname' ::= <junk> len:<u32> in:<importname>                     => in  (if len = |in|)
+exportname' ::= <junk> len:<u32> en:<exportname>                     => en  (if len = |en|)
+junk        ::= 0x00
+              | 0x01
```
alexcrichton, line 316: One possibility for this is to only require 0x00 here. AFAIK `wasm-tools` and `wasmparser` are the only affected component of this change, so while there I'll implement "discard 0x00 or 0x01 and ignore what you just read" it may not be necessary to codify such behavior in the spec

alexcrichton: The behavior in wasm-tools is actually going to be a bit more subtle as well, we'll need to, have a few phases such as:

1. Add support for this new encoding, but don't emit it by default. Instead default to the prior 0x00 / 0x01 encoding.
2. Wait for tools to percolate a bit throughout the ecosystem
3. Once percolated emit 0x00 by default.

Otherwise we'll run a risk of a tool producing 0x00-encoding with `wasi:foo/bar` for example which would be rejected by prior versions of the validator.

alexcrichton (approved):
All looks good to me 👍

design/mvp/Binary.md
lukewagner: Ah, that's a good point.  So if we wanted a maximally-backwards-compatible transition, we'd need to keep the byte discriminants paired with the string cases (`0x00` for `<plainname>`, etc), which we could do, but if we're able to do a 3-step transition like you're talking about, then we might as well just do what you suggest and just fix the `0x00` byte (which could well be useful in the future as a "flags" byte that enables other subfields, so not even necessarily a thing we need to remove).  So yeah, I'll switch to just a fixed `0x00` if that still sounds good to you.

lukewagner: [Done](https://github.com/WebAssembly/component-model/pull/263/commits/5efe6f8935897d4044346481dff234649bb096ee)

alexcrichton:
To confirm, these two binary rules:

```ebnf
importdecl    ::= in:<importname> ed:<externdesc>         => (import in ed)
exportdecl    ::= en:<exportname> ed:<externdesc>         => (export en ed)
```

should change to refer to `importname'` and `exportname'`, right?

alexcrichton:
One other piece of feedback from https://github.com/bytecodealliance/wasm-tools/pull/1262, an implementation of this (and a few other PRs to the component model). Currently URLs are defined as "stuff that doesn't contain `<` or `>`", but `integrity-metadata` is defined with the w3c definition. It might make sense to do the same thing in both situations? Either require both are w3c-defined valid or both are something-nonempty-someone-else-parses-this. 

I'll admit though I know nothing about `integrity-metadata` in w3c, if it's a simple format that's easy to parse then may as well "just" do that too.

lukewagner:
@alexcrichton That's a good question and I was wondering about the same options for the same reason.  Looking at the EBNF of `integrity-metadata` ([here](https://www.w3.org/TR/SRI/#the-integrity-attribute)) and recursing through all the definitions, it does seem like `integrity-metadata` is much much simpler than a URL, but a good deal more complicated than, say, an `interfacename`, so it's in a sortof grey area where I could go either way.

But another reason in favor of having the Component Model fully parse `integrity-metadata` is that, unlike URLs, which are pretty universal (ignoring the many awful edge cases), there are, I believe, a couple of different hash text formats out there.  E.g., OCI uses [something different](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests).  Thus, it seems quite possible that different parts of the ecosystem might end up adopting slightly-different hash formats (accidentally or intentionally) if the common validation rules baked into the core-est tooling don't fix one.

alexcrichton:
Ok makes sense to me! I'll dig in and implement that

peterhuene (approved):
This looks really great! I'm reviewing the implementation side of this currently as well.

lukewagner:
Awesome, thanks to you and Alex for the feedback.  I'll merge now since there don't seem to be any open issues and we can file follow-up issues if needed as usual.

Issue #261, opened 2023-10-04
Request for spec documentation: all new index spaces and their contents listed in one place


fitzgen:
One thing I keep stumbling over whenever I dig into the component model is what all the new index spaces are and what is in them, in one easy-to-scan place.

For example, I was trying to determine whether component types and defined types are in the same index space. That is really hard to determine right now when reading the explainer.

lukewagner:
I suppose the main point of reference is the definition of `sort` in [the explainer](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#instance-definitions) or [the binary format](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Binary.md#instance-definitions).  In particular: each `sort` gets an index space (indexed by, e.g., a `sortidx`).  Thus, since both component types and defined types both have sort `type`, they'd go in the same index space.  For new documentation, are you imagining some sort of "Index Spaces" section that lists and describes the index spaces, like [this one](https://webassembly.github.io/spec/core/syntax/modules.html#indices) in the core spec?

fitzgen:
Yeah something like that. This has been the sort (ha) of question I want to answer for myself most often when reading the explainer and I don't feel like I come away sure of the answer.

lukewagner:
Opened #264

PR #264, opened 2023-10-18
Add subsection describing Index Spaces

lukewagner:
As requested in #261.  @fitzgen let me know if there's anything else you think should be in here.


fitzgen (approved):
This is really helpful!

I think the one thing that might improve this a little is differentiating core modules and instances from the rest of the core index spaces, since they aren't really part of core Wasm and are instead part of the component-model, even though they deal with core things.

lukewagner:
Ooh yes, good point; will do in a sec.

PR #270, opened 2023-11-08
Use 'package name' more consistently in docs

lukewagner:
This PR updates the terminology to match #263 (no change in behavior).

Prior to #263, there was a sortof haphazard mix of "name" and "id" used in the context of imports/exports referring to interfaces and implementations, which created some ambiguity because there was also this totally independent pre-existing use of the term "identifier" in both WAT and WIT.  #263 tried to remove this inconsistency and simply call everything an "X name" if it was used inside an `importname` or `exportname` (so: "plain name", "interface name", etc), which was more obvious to do once everything was inside the "name" string.  This PR ties up some loose ends that folks in SIG-Registry pointed out today.  This isn't the only possible regular terminology, though; happy to discuss alternatives.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -1364,11 +1364,11 @@ interfacename ::= <namespace> <label> <projection> <version>?
 namespace     ::= <label> ':'
 projection    ::= '/' <label>
 version       ::= '@' <valid semver>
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
👍

peterhuene (approved):

PR #279, opened 2023-11-30
Perform NaN randomization as part of the interpretation of float bits

sunfishcode:
This removes the `THE_HOST_WANTS_TO` option, which appeared to give hosts permission to interpret NaN bits how they "want to". The change here is to say that all core-wasm NaN bitpatterns are interpreted as the same component-model NaN value.

It's my understanding that the `random_nan_bits` function isn't meant to be the precise algorithm that nondeterministic-profile implementations must use, so this doesn't require hosts to do any new randomization work.

This change also fixes what appears to be a bug:
`lift_flat_variant`/`lower_flat_variant` were calling `reinterpret_i32_as_float`/`reinterpret_float_as_i32` without performing NaN scrambling. By making NaN scrambling be part of interpretation, we ensure that it's performed anywhere interpretation is performed.


lukewagner:
Good catch on the missing `maybe_scramble_nan*` in the flat variant cases, we should fix that.  However, I don't think this PR causes "all core-wasm NaN bitpatterns are interpreted as the same component-model NaN value" since it still produces semantically distinct NaN values upon lifting and lowering.  Moreover, although your change is technically equivalent (saying you can non-deterministically choose to non-deterministically scramble NaN payloads is the same as saying that you can non-deterministically scramble NaN values), I think it's useful to make it clear to the reader that the host is allowed by the spec to simply propagate the NaN bitpattern; this just avoids confusion, so I'd rather keep `THE_HOST_WANTS_TO`.

sunfishcode:
> However, I don't think this PR causes "all core-wasm NaN bitpatterns are interpreted as the same component-model NaN value" since it still produces semantically distinct NaN values upon lifting and lowering.

I've now changed the PR to canonicalize on lifting, and scramble on lowering. This better illustrates how "all core-wasm NaN bitpatterns are interpreted as the same component-model NaN value".

>  I think it's useful to make it clear to the reader that the host is allowed by the spec to simply propagate the NaN bitpattern; this just avoids confusion, so I'd rather keep `THE_HOST_WANTS_TO`.

It's my understanding that your goal here is to avoid confusion among hosts about "do I need to randomize?"

My goal here is to avoid the confusion among hosts about "is it ok for me to interpret the bits, if I «want to»?" If they do that, they break virtualization in a subtle way.

Also, if we can say there's only one component-model NaN value, that would make it clear why [wave](https://github.com/lann/wave#wave-web-assembly-value-encoding) doesn't need nan payload syntax.

I've also now added comments to this PR to call out how hosts can avoid skip canonicalization and randomization overhead entirely in practice.

design/mvp/CanonicalABI.md
```diff
@@ -480,34 +480,41 @@ the [deterministic profile], NaNs are canonicalized to a particular NaN
...
   if math.isnan(f):
-    if DETERMINISTIC_PROFILE:
-      f = reinterpret_i64_as_float(CANONICAL_FLOAT64_NAN)
-    elif THE_HOST_WANTS_TO:
-      f = reinterpret_i64_as_float(random_nan_bits(64, 11))
+    f = core_f64_reinterpret_i64(CANONICAL_FLOAT64_NAN)
     assert(math.isnan(f))
   return f
 
 def reinterpret_i32_as_float(i):
```
lukewagner, line 501: "reinterpret" suggests (to me) that "the bits don't change", so could you maybe rename this function to `convert_i32_to_f32`/`convert_i64_to_f64`?

design/mvp/CanonicalABI.md
```diff
@@ -480,34 +480,41 @@ the [deterministic profile], NaNs are canonicalized to a particular NaN
...
+
+def core_f64_reinterpret_i64(i):
   return struct.unpack('!d', struct.pack('!Q', i))[0] # f64.reinterpret_i64
 
+# Generate an integer value representing the bitpattern of a NaN with
+# random bits.
+#
+# The specific random function here isn't part of the specification, and is
+# only an example of randomization which may occur.
 def random_nan_bits(total_bits, exponent_bits):
```
lukewagner, line 518: Since this function isn't used by the code above anymore, could you move it down lower to near where it's used?

design/mvp/CanonicalABI.md
```diff
@@ -480,34 +480,41 @@ the [deterministic profile], NaNs are canonicalized to a particular NaN
...
 ```python
 DETERMINISTIC_PROFILE = False # or True
-THE_HOST_WANTS_TO = True # or False
 CANONICAL_FLOAT32_NAN = 0x7fc00000
 CANONICAL_FLOAT64_NAN = 0x7ff8000000000000
 
-def maybe_scramble_nan32(f):
+# NaN values are always canonicalized when lifted, to illustrate that even
+# though Python floating-point values preserve NaN payloads, Component Model
+# floating-point values do not.
```
lukewagner, line 488: To be consistent with the general style of this document, could you move the comments into prose descriptions above or below (but outside of) this code block?

lukewagner (approved):
I really like this idea of saying that there is only one "semantic" NaN value -- it's just that lowering scrambles and oh hey, it's technically valid not to canonicalize if you don't want to.  Nice and thanks for the consideration!

Just stylistic nits:

design/mvp/CanonicalABI.md
sunfishcode: To me, `convert` sounds like numeric conversion. How about `encode_`/`decode_`?

design/mvp/CanonicalABI.md
sunfishcode: done

design/mvp/CanonicalABI.md
sunfishcode: done

design/mvp/CanonicalABI.md
```diff
@@ -703,8 +694,8 @@ def store(cx, v, t, ptr):
     case S16()          : store_int(cx, v, ptr, 2, signed=True)
     case S32()          : store_int(cx, v, ptr, 4, signed=True)
     case S64()          : store_int(cx, v, ptr, 8, signed=True)
-    case Float32()      : store_int(cx, reinterpret_float_as_i32(maybe_scramble_nan32(v)), ptr, 4)
-    case Float64()      : store_int(cx, reinterpret_float_as_i64(maybe_scramble_nan64(v)), ptr, 8)
+    case Float32()      : store_int(cx, encode_float_as_i32(v), ptr, 4)
+    case Float64()      : store_int(cx, encodencode_float_as_i64(v), ptr, 8)
```
lukewagner, line 698:
```suggestion
    case Float64()      : store_int(cx, encode_float_as_i64(v), ptr, 8)
```

design/mvp/CanonicalABI.md
lukewagner: Oh right, good point.  I like it, thanks!

lukewagner:
Thanks again for thinking through this!

Issue #274, opened 2023-11-25
Rename `world` to `component`


vwkd:
In Wit, a "world" corresponds to a component. This took me too long to grasp, because "world" is a strange word for it. "World" has a notion of vastness, chaos, life. Intuitively, it doesn't relate at all to component.

Consider renaming `world` to `component`.

IIUC, the name "world" originated by looking at built-in components the host offers to guest components as different "worlds". While somehow sensible for that limited notion, this comes at the cost of clarity for non-host components, which likely will be the vast majority in a component universe.

Example:

```wit
package myorg:http@0.0.1

// ...

component proxy {
    export incoming-handler
    import outgoing-handler
}
```

rylev:
Chiming in to provide perspective that isn't from the spec authors.

I believe this proposal only makes sense when considered from the guest's perspective as the guest is itself a component. However, this doesn't work from the host's perspective since the host's world is a superset of any given guest's imports and a subset of the guest component's exports. In other words, from the host's perspective a world is all of the imports it is capable of supplying and all the exports it is capable of calling. Any given guest component will likely only have a subset of these imports and potentially a superset of the exports. 

A world is "all of the things I understand" and from this perspective I think world is sensible choice. 

That being said, "world" is far from standard terminology so its usage will present a substantial educational challenge so I personally am sympathetic to the desire to find terminology that is more immediately obvious.

yordis:
I'd like to share my perspective as someone for whom English is a second language.

My first language is Spanish, and despite having spoken English for over a decade, I still speak with an accent and occasionally struggle with pronunciation.

I rather see "Planet" 😄 Anything that is much easier to say and `type`**** I do not know how many times I typed `wordl` ... in my career.

aschrijver:
Realm?

vwkd:
Appreciate the constructive feedback and highlighting the host perspective which the initial post had mentioned only shortly.

It seems to me the host itself can be seen as a component that the guest component plugs into. This built-in host component does have one special exception: not all its exports need to be fulfilled (the guest component can import only some of them).

It seems easier to me to look at the host boundary as a special component, than to look at every component as a special host boundary.

That said, "X world" could still be a useful synonym to refer to different built-in host components, particularly when a host supports multiple.

rossberg:
Chiming in as well, I'm also not super-happy with the terminology choices made by wit:

* Why use different terminology than the underlying component model and Wasm itself? Doesn't that just create unnecessary friction?

[...]

PR #283, opened 2023-12-04
feat(wit): enrich the definition of a world

Mossaka:
The concept of a "World" in WIT describes something larger than what a `component` type defined in the Wasm Component Model needs. Thus it's no longer precise to say a world is a `component`, but it describes the interaction capabilities of components targeting it. 

Related to #274

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -132,10 +132,18 @@ interfaces are well-formed and acyclic.
 [worlds]: #wit-worlds
 
 WIT packages can contain `world` definitions at the top-level in addition to
-[`interface`][interfaces] definitions. A world is a complete description of
-both imports and exports of a component. A world can be thought of as an
-equivalent of a `component` type in the component model. For example this
-world:
+[`interface`][interfaces] definitions.
+
+Conceptually, a world encapsulates the full spectrum of interaction capabilities for a component.
+There are two perspectives looking at a world:
+
+- The guest's perspective, a world that it targets describes all
+of the imports it is capable of calling and all the exports it is required to provide. A guest `component` may provide more exports than what the world requires.
+- The host's perspective, a world that it targets describes all of the imports it is required to provide and all the exports it is capable of calling. The host may provide more imports than what the world requires.
+
+A world can be thought of as an equivalent of a `component` type in the component model
+
+For example this world:
 
 ```wit
 package local:demo;


design/mvp/WIT.md
```diff
@@ -132,10 +132,18 @@ interfaces are well-formed and acyclic.
...
+[`interface`][interfaces] definitions.
+
+Conceptually, a world encapsulates the full spectrum of interaction capabilities for a component.
+There are two perspectives looking at a world:
+
+- The guest's perspective, a world that it targets describes all
+of the imports it is capable of calling and all the exports it is required to provide. A guest `component` may provide more exports than what the world requires.
+- The host's perspective, a world that it targets describes all of the imports it is required to provide and all the exports it is capable of calling. The host may provide more imports than what the world requires.
+
+A world can be thought of as an equivalent of a `component` type in the component model
```
jsturtevant, line 144: At first glance, this feels mis-leading based on what was added above.  Maybe something like

```suggestion
A world can be thought of as an equivalent of a `component` type in the component model although the actual component may contain more imports/exports than described in the world.
```

and update the component below this to have an extra import with a comment saying it is not in the world directly but the component still satisfies the world?

design/mvp/WIT.md
```diff
@@ -132,10 +132,18 @@ interfaces are well-formed and acyclic.
...
 
 WIT packages can contain `world` definitions at the top-level in addition to
-[`interface`][interfaces] definitions. A world is a complete description of
-both imports and exports of a component. A world can be thought of as an
-equivalent of a `component` type in the component model. For example this
-world:
+[`interface`][interfaces] definitions.
+
+Conceptually, a world encapsulates the full spectrum of interaction capabilities for a component.
+There are two perspectives looking at a world:
```
rossberg, line 138: To me, this reads like a rather convoluted way of saying "a world is the interface of a component". Which indeed it is. :)

design/mvp/WIT.md
```diff
@@ -132,10 +132,18 @@ interfaces are well-formed and acyclic.
...
+
+- The guest's perspective, a world that it targets describes all
+of the imports it is capable of calling and all the exports it is required to provide. A guest `component` may provide more exports than what the world requires.
```
rylev, line 141: It would be helpful to note that the guest *cannot* expect more imports than the world specifies.

design/mvp/WIT.md
```diff
@@ -132,10 +132,18 @@ interfaces are well-formed and acyclic.
...
+- The host's perspective, a world that it targets describes all of the imports it is required to provide and all the exports it is capable of calling. The host may provide more imports than what the world requires.
```
rylev, line 142: Same as above, the host cannot call exports the world does not specify.

design/mvp/WIT.md
rylev: > A world can be thought of as an equivalent of a `component` type in the component model

I don't think this is really true from the host's perspective. The world describes a subset a superset of imports and a subset of exports of any given component. It may be that a host never runs a component that mirrors the world exactly.

lukewagner:
I'll close this PR for now, since it's been quiet for a while and now we have https://component-model.bytecodealliance.org/design/worlds.html.  But feel free to reopen if there's any interest in addressing the feedback above and merging.

Issue #282, opened 2023-12-01
Reserve handle 0


badeend:
to be an always-invalid "NULL" handle?

Two benefits that come to mind for client languages:
- easy to spot usage of uninitialized fields. Trying to use an uninitialized 0 field as a handle would then cause an error, instead of silently using the first resource.
- enables optional handles to be stored with zero overhead. Similar to Rust's option null pointer optimization (`Option<NonZeroU32>`)

lukewagner:
That's a really great idea.  I feel like we even discussed this at some point in the past and then I forgot to include this in CanonicalABI.md.  In particular, in the Preview 3 timeframe, I'm expecting there to be some canonical built-ins that want to optionally return a handle which would benefit from being able to simply return an `i32`.

`-1` is another option for a sentinel value worth considering.  I just noticed that CanonicalABI.md is missing an integer overflow trap btw... but assuming I add that, we could just decrease the max.  `0` seems like the better sentinel value for catching errors, as you pointed out.  However, `0` means either burning an extra element per resource table or adding an extra subtraction on the access path.  Making `0` a sentinel would *technically* be a breaking change (since handle indices are currently deterministically specified), but it seems like code "shouldn't" depend on the precise indices so maybe it's not in practice.

@alexcrichton (and anyone else) WDYT?

pchickey:
Until very recently I made the wasmtime-wasi implementation refuse to use indices 0,1,2 for pseudo-resources, and then resources, in order to catch programs that were accidentally assuming stdio lived there. This did catch some bugs in the early days of pseudo-resources, but I'm not sure if it would since we introduced proper resources in Wasmtime. We got rid of that special case recently on the grounds that we probably shook out all of those bugs. I would be fine with this change.

badeend:
> -1 is another option for a sentinel value worth considering. I just noticed that CanonicalABI.md is missing an integer overflow trap btw... but assuming I add that, we could just decrease the max. 0 seems like the better sentinel value for catching errors, as you pointed out.

`0` would definitely be my preferences because of the pervasiveness of 0 initialization/coercion in the programming language design universe. (In)famous example being JavaScript: `Number(null) === 0`.

> However, 0 means either burning an extra element per resource table (...)

[...]

PR #284, opened 2023-12-05
Reserve handle index 0 and add handle index max in Canonical ABI

lukewagner:
As discussed in #282.  In theory, this shouldn't be breaking since guest code shouldn't be depending on the precise index values they receive, treating them like opaque file-descriptor-like values.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -350,7 +350,7 @@ class HandleTable:
   free: [int]
 
   def __init__(self):
-    self.array = []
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):

alexcrichton:
I've posted https://github.com/bytecodealliance/wasmtime/pull/7661 as the Wasmtime-side implementation change this correponds to

lukewagner:
Great!  I'm thinking I'll wait another few days to see if there's any feedback and otherwise merge mid next week.

PR #288, opened 2023-12-21
Add more documentation about types.

sunfishcode:
Add documentation describing the floating-point types, and briefly describe the difference between core-wasm numeric types and component-level numeric types.

Add more subsection headings to the prose describing the types, breaking up a sizeable quantity of text, and adding anchors that can be linked to. In particular, I've often wanted an anchor for "handle types".

And, fix the number of definition types.


lukewagner (approved):
lgtm, thanks for expanding on this!

(Given folks are still getting back from the holidays, I'll keep this open another week or so for feedback before merging.)

PR #291, opened 2024-01-03
Add canonical `thread.*` builtins

abrown:
This change adds the `thread.spawn` and `thread.hw_concurrency` builtins in line with what was proposed in the shared-everything threads [proposal].

[proposal]: https://github.com/WebAssembly/shared-everything-threads/blob/main/proposals/shared-everything-threads/Overview.md#thread-management-builtins

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1709,8 +1709,6 @@ validation specifies:
 * `$st` is given type `(func (param $f (ref null $ft)) (param $c i32) (result $e
   i32))`.
 
-In the [deterministic profile], validation fails if `thread.spawn` is used.
-
 > Note: ideally, a thread could be spawned with [arbitrary thread parameters].
 > To import polymorphic versions of `$st`, a naming scheme is necessary to
 > differentiate between the imports with varying `$ft`. Since many languages can
@@ -1719,8 +1717,8 @@ In the [deterministic profile], validation fails if `thread.spawn` is used.
 > The inclusion of `$ft` ensures backwards compatibility for when arbitrary
 > parameters are allowed.
 
-Calling `$st` checks that the reference `$f` is not null and satisfies type
-`$ft`. Then, it spawns a thread which:
+Calling `$st` checks that the reference `$f` is not null. Then, it spawns a
+thread which:
   - invokes `$f` with `$c`
   - executes `$f` until completion or trap in a `shared` context as described by
     the [shared-everything threads] proposal.
@@ -1728,8 +1726,8 @@ Calling `$st` checks that the reference `$f` is not null and satisfies type
 In pseudocode, `$st` looks like:
 
 ```python
-def canon_thread_spawn(ft, f, c):
-  trap_if(f is None or ft is not f.type)
+def canon_thread_spawn(f, c):
+  trap_if(f is None)
   if DETERMINISTIC_PROFILE:
     return -1
   if spawn(lambda: f(c)):
@@ -1745,7 +1743,7 @@ For a canonical definition:
 (canon thread.hw_concurrency (core func $f))
 ```
 validation specifies:
-* `$f` is imported with type `(func shared (result i32))`.
+* `$f` is given type `(func shared (result i32))`.
 
 Calling `$f` returns the number of threads the underlying hardware can be
 expected to execute concurrently. This value can be artificially limited by
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1717,8 +1717,8 @@ validation specifies:
 > The inclusion of `$ft` ensures backwards compatibility for when arbitrary
 > parameters are allowed.
 
-Calling `$st` spawns `$n` threads, each of which:
-  - checks that reference `$f` is not null and satisfies type `$ft`
+Calling `$st` checks that the reference `$f` is not null and satisfies type
+`$ft`. Then, it spawns `$n` threads, each of which:
   - invokes `$f` with `$c`
   - executes `$f` until completion or trap in a `shared` context as described by
     the [shared-everything threads] proposal.
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1706,7 +1706,7 @@ For a canonical definition:
 validation specifies:
 * `$ft` must refer to a `shared` function type; initially, only the type `(func
   shared (param $c i32))` is allowed (see explanation below)
-* `$st` is imported with type `(func (param $f (ref null $ft)) (param $n i32)
+* `$st` is given type `(func (param $f (ref null $ft)) (param $n i32)
   (param $c i32))`.
 
 > Note: ideally, a thread could be spawned with [arbitrary thread parameters].
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1729,9 +1729,8 @@ In pseudocode, `$st` looks like:
 
 ```python
 def canon_thread_spawn(ft, f, n, c):
-  assert(not DETERMINISTIC_PROFILE)
   trap_if(f is None or ft is not f.type)
-  if not can_spawn(n):
+  if DETERMINISTIC_PROFILE:
     return -1
   for i in range(n):
     spawn(lambda: f(c))
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1707,7 +1707,7 @@ validation specifies:
 * `$ft` must refer to a `shared` function type; initially, only the type `(func
   shared (param $c i32))` is allowed (see explanation below)
 * `$st` is given type `(func (param $f (ref null $ft)) (param $n i32)
-  (param $c i32))`.
+  (param $c i32) (result $e i32))`.
 
 In the [deterministic profile], validation fails if `thread.spawn` is used.
 
@@ -1731,8 +1731,11 @@ In pseudocode, `$st` looks like:
 def canon_thread_spawn(ft, f, n, c):
   assert(not DETERMINISTIC_PROFILE)
   trap_if(f is None or ft is not f.type)
+  if not can_spawn(n):
+    return -1
   for i in range(n):
     spawn(lambda: f(c))
+  return 0
 ```
 
 ### 🧵 `canon thread.hw_concurrency`
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1706,8 +1706,8 @@ For a canonical definition:
 validation specifies:
 * `$ft` must refer to a `shared` function type; initially, only the type `(func
   shared (param $c i32))` is allowed (see explanation below)
-* `$st` is given type `(func (param $f (ref null $ft)) (param $n i32)
-  (param $c i32) (result $e i32))`.
+* `$st` is given type `(func (param $f (ref null $ft)) (param $c i32) (result $e
+  i32))`.
 
 In the [deterministic profile], validation fails if `thread.spawn` is used.
 
@@ -1720,7 +1720,7 @@ In the [deterministic profile], validation fails if `thread.spawn` is used.
 > parameters are allowed.
 
 Calling `$st` checks that the reference `$f` is not null and satisfies type
-`$ft`. Then, it spawns `$n` threads, each of which:
+`$ft`. Then, it spawns a thread which:
   - invokes `$f` with `$c`
   - executes `$f` until completion or trap in a `shared` context as described by
     the [shared-everything threads] proposal.
@@ -1728,13 +1728,14 @@ Calling `$st` checks that the reference `$f` is not null and satisfies type
 In pseudocode, `$st` looks like:
 
 ```python
-def canon_thread_spawn(ft, f, n, c):
+def canon_thread_spawn(ft, f, c):
   trap_if(f is None or ft is not f.type)
   if DETERMINISTIC_PROFILE:
     return -1
-  for i in range(n):
-    spawn(lambda: f(c))
-  return 0
+  if spawn(lambda: f(c)):
+    return 0
+  else:
+    return -1
 ```
 
 ### 🧵 `canon thread.hw_concurrency`
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1697,6 +1697,53 @@ def canon_resource_rep(inst, rt, i):
 Note that the "locally-defined" requirement above ensures that only the
 component instance defining a resource can access its representation.
 
+### 🧵 `canon thread.spawn`
+
+For a canonical definition:
+```wasm
+(canon thread.spawn (type $ft) (core func $st))
+```
+validation specifies:
+* `$ft` must refer to a `shared` function type; initially, only the type `(func
+  shared (param $c i32))` is allowed (see explanation below)
+* `$st` is imported with type `(func (param $f (ref null $ft)) (param $n i32)
+  (param $c i32))`.
+
+> Note: ideally, a thread could be spawned with [arbitrary thread parameters].
+> To import polymorphic versions of `$st`, a naming scheme is necessary to
+> differentiate between the imports with varying `$ft`. Since many languages can
+> use `$c` to address a memory region containing multiple values (the current
+> [wasi-libc convention]), the initial version restricts `$ft` to `[i32] -> []`.
+> The inclusion of `$ft` ensures backwards compatibility for when arbitrary
+> parameters are allowed.
+
+Calling `$st` spawns `$n` threads, each of which:
+  - checks that reference `$f` is not null and satisfies type `$ft`
+  - invokes `$f` with `$c`
+  - executes `$f` until completion or trap in a `shared` context as described by
+    the [shared-everything threads] proposal.
+
+In pseudocode, `$st` looks like:
+
+```python
+def canon_thread_spawn(ft, f, n, c):
+  trap_if(f is None or ft is not f.type)
+  for i in range(n):
+    spawn(lambda: f(c))
+```
+
+### 🧵 `canon thread.hw_concurrency`
+
+For a canonical definition:
+```wasm
+(canon thread.hw_concurrency (core func $f))
+```
+validation specifies:
+* `$f` is imported with type `(func shared (result i32))`.
+
+Calling `$f` returns the number of threads the underlying hardware can be
+expected to execute concurrently. This value can be artificially limited by
+engine configuration.
 
 
 [Canonical Definitions]: Explainer.md#canonical-definitions
@@ -1730,3 +1777,7 @@ component instance defining a resource can access its representation.
 
 [`import_name`]: https://clang.llvm.org/docs/AttributeReference.html#import-name
 [`export_name`]: https://clang.llvm.org/docs/AttributeReference.html#export-name
+
+[Arbitrary Thread Parameters]: https://github.com/WebAssembly/shared-everything-threads/discussions/3
+[wasi-libc Convention]: https://github.com/WebAssembly/wasi-libc/blob/925ad6d7/libc-top-half/musl/src/thread/pthread_create.c#L318
+[Shared-Everything Threads]: https://github.com/WebAssembly/shared-everything-threads/blob/main/proposals/shared-everything-threads/Overview.md
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1730,7 +1730,14 @@ def canon_thread_spawn(f, c):
   trap_if(f is None)
   if DETERMINISTIC_PROFILE:
     return -1
-  if spawn(lambda: f(c)):
+
+  def thread_start():
+    try:
+      f(c)
+    except CoreWebAssemblyException:
+      trap()
+
+  if spawn(thread_start):
     return 0
   else:
     return -1
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1743,8 +1743,16 @@ validation specifies:
 
 Calling `$f` returns the number of threads the underlying hardware can be
 expected to execute concurrently. This value can be artificially limited by
-engine configuration.
+engine configuration and is not allowed to change over the lifetime of a
+component instance.
 
+```python
+def canon_thread_hw_concurrency():
+  if DETERMINISTIC_PROFILE:
+    return 1
+  else:
+    return NUM_ALLOWED_THREADS
+```
 
 [Canonical Definitions]: Explainer.md#canonical-definitions
 [`canonopt`]: Explainer.md#canonical-definitions
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1710,10 +1710,10 @@ validation specifies:
   i32))`.
 
 > Note: ideally, a thread could be spawned with [arbitrary thread parameters].
-> To import polymorphic versions of `$st`, a naming scheme is necessary to
-> differentiate between the imports with varying `$ft`. Since many languages can
-> use `$c` to address a memory region containing multiple values (the current
-> [wasi-libc convention]), the initial version restricts `$ft` to `[i32] -> []`.
+> Currently, that would require additional work in the toolchain to support so,
+> for simplicity, the current proposal simply fixes a single `i32` parameter type.
+> However, `thread.spawn` could be extended to allow arbitrary thread parameters
+> in the future, once it's concretely beneficial to the toolchain. 
 > The inclusion of `$ft` ensures backwards compatibility for when arbitrary
 > parameters are allowed.
 
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1709,6 +1709,8 @@ validation specifies:
 * `$st` is given type `(func (param $f (ref null $ft)) (param $n i32)
   (param $c i32))`.
 
+In the [deterministic profile], validation fails if `thread.spawn` is used.
+
 > Note: ideally, a thread could be spawned with [arbitrary thread parameters].
 > To import polymorphic versions of `$st`, a naming scheme is necessary to
 > differentiate between the imports with varying `$ft`. Since many languages can
@@ -1727,6 +1729,7 @@ In pseudocode, `$st` looks like:
 
 ```python
 def canon_thread_spawn(ft, f, n, c):
+  assert(not DETERMINISTIC_PROFILE)
   trap_if(f is None or ft is not f.type)
   for i in range(n):
     spawn(lambda: f(c))
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -1270,6 +1270,11 @@ transferring ownership of the newly-created resource to the export's caller.
 
 ##### 🧵 Threads
 
+The [shared-everything-threads] proposal adds component model built-ins for
+thread management. These are specified as built-ins and not core WebAssembly
+instructions because browsers expect this functionality to come from existing
+Web/JS APIs.
+
 The `thread.spawn` built-in has type `[f:(ref null $f) n:i32 c:i32] -> []` and
 spawns a new thread by invoking the shared function `f` `n` times while passing
 `c` to each.
@@ -1958,3 +1963,5 @@ and will be added over the coming months to complete the MVP proposal:
 
 [Scoping and Layering]: https://docs.google.com/presentation/d/1PSC3Q5oFsJEaYyV5lNJvVgh-SNxhySWUqZ6puyojMi8
 [Future and Stream Types]: https://docs.google.com/presentation/d/1MNVOZ8hdofO3tI0szg_i-Yoy0N2QPU2C--LzVuoGSlE
+
+[shared-everything-threads]: https://github.com/WebAssembly/shared-everything-threads
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -32,6 +32,7 @@ emoji symbols listed below; these emojis will be removed once they are
 implemented, considered stable and included in a future milestone:
 * 🪙: value imports/exports and component-level start function
 * 🪺: nested namespaces and packages in import/export names
+* 🧵: thread lifecycle built-ins
 
 (Based on the previous [scoping and layering] proposal to the WebAssembly CG,
 this repo merges and supersedes the [module-linking] and [interface-types]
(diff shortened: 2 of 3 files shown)


design/mvp/Explainer.md
```diff
@@ -32,6 +32,7 @@ emoji symbols listed below; these emojis will be removed once they are
 implemented, considered stable and included in a future milestone:
 * 🪙: value imports/exports and component-level start function
 * 🪺: nested namespaces and packages in import/export names
+* 🧵: thread lifecycle built-ins
```
lukewagner, line 35:
```suggestion
* 🧵: threading built-ins
```
Suggesting this generalization since `thread.hw_concurrency` doesn't seem "lifecycle"-related to me.

design/mvp/Explainer.md
```diff
@@ -1262,6 +1268,15 @@ Here, the `i32` returned by `resource.new`, which is an index into the
 component's handle-table, is immediately returned by `make_R`, thereby
 transferring ownership of the newly-created resource to the export's caller.
 
+##### 🧵 Threads
+
+The `thread.spawn` built-in has type `[f:(ref null $f) n:i32 c:i32] -> []` and
```
lukewagner, line 1273: Perhaps add a short paragraph linking to the shared-everything-proposal and explaining how these built-ins fill in gaps in functionality that is only exposed in browsers through Web/JS APIs?

design/mvp/CanonicalABI.md
```diff
@@ -1697,6 +1697,53 @@ def canon_resource_rep(inst, rt, i):
...
+### 🧵 `canon thread.hw_concurrency`
+
+For a canonical definition:
+```wasm
+(canon thread.hw_concurrency (core func $f))
+```
+validation specifies:
+* `$f` is imported with type `(func shared (result i32))`.
+
+Calling `$f` returns the number of threads the underlying hardware can be
```
lukewagner, line 1744: Could you add a pseudo-code snippet (in the same manner as you did with `canon_thread_spawn` that says `if DETERMINISTIC_PROFILE return 1`?

design/mvp/CanonicalABI.md
```diff
@@ -1697,6 +1697,53 @@ def canon_resource_rep(inst, rt, i):
...
+  - checks that reference `$f` is not null and satisfies type `$ft`
+  - invokes `$f` with `$c`
+  - executes `$f` until completion or trap in a `shared` context as described by
+    the [shared-everything threads] proposal.
+
+In pseudocode, `$st` looks like:
+
+```python
+def canon_thread_spawn(ft, f, n, c):
+  trap_if(f is None or ft is not f.type)
```
lukewagner, line 1730: I guess we should say *something* about what happens when executing in the deterministic profile.  A conservative option is to say that components containing the `thread.spawn` built-in fail validation on a deterministic host (just like core wasm would spec if `thread.spawn` was an instruction).  If that makes sense, could you add a comment saying as much?  Also, there's a `DETERMINISTIC_PROFILE` global boolean (used by float NaN canonicalization elsewhere in this doc), so maybe add a `assert(not DETERMINISTIC_PROFILE)` to the pseudo-code to reinforce this point?

design/mvp/CanonicalABI.md
```diff
@@ -1697,6 +1697,53 @@ def canon_resource_rep(inst, rt, i):
...
+
+For a canonical definition:
+```wasm
+(canon thread.spawn (type $ft) (core func $st))
+```
+validation specifies:
+* `$ft` must refer to a `shared` function type; initially, only the type `(func
+  shared (param $c i32))` is allowed (see explanation below)
+* `$st` is imported with type `(func (param $f (ref null $ft)) (param $n i32)
+  (param $c i32))`.
```
lukewagner, line 1710: As is, `thread.spawn` is infallible.  Is the intention then that OOM causes a trap (like a stack overflow) rather than a dynamic failure (like a `memory.grow` failure)?  If so, I'd add a comment/paragraph, although I think it's worth considering dynamic failure on OOM (which does raises a subtle question of whether failure is atomic when spawning N>1 threads).

lukewagner (commented):
Looks generally good so far; nice work!  Is the only other thing left to add before this PR is complete the binary encoding of `thread.spawn` and `thread.hw_concurrency` in Binary.md?

design/mvp/CanonicalABI.md
tlively: It would be more flexible to have `thread.spawn` fail at runtime. That would allow components to gracefully fall back to running single-threaded if `hw_concurrency` is 1. This is analogous to how we allow atomic operations on non-shared memories and how we allow atomic waits to validate even on non-blocking threads.

design/mvp/CanonicalABI.md
```diff
@@ -1697,6 +1697,53 @@ def canon_resource_rep(inst, rt, i):
...
+> Note: ideally, a thread could be spawned with [arbitrary thread parameters].
+> To import polymorphic versions of `$st`, a naming scheme is necessary to
+> differentiate between the imports with varying `$ft`. Since many languages can
+> use `$c` to address a memory region containing multiple values (the current
+> [wasi-libc convention]), the initial version restricts `$ft` to `[i32] -> []`.
+> The inclusion of `$ft` ensures backwards compatibility for when arbitrary
+> parameters are allowed.
+
+Calling `$st` spawns `$n` threads, each of which:
+  - checks that reference `$f` is not null and satisfies type `$ft`
```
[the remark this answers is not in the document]
abrown: When, though? At instantiation? Or is that too restrictive?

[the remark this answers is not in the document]
abrown: Oh, never mind, I think I see what you mean: "Check for null... then spawn `$n` threads," right?

design/mvp/CanonicalABI.md
abrown: I was leaning toward what @tlively suggested but then considered this: the point in time the spawned function begins executing is non-deterministic and unspecified. It is non-deterministic in the timing of the instructions executed by the main thread and the spawned thread &mdash; in a single-threaded world, which "next instruction" do we execute first? We _could_ fix this by specifying an order: we first execute the spawned function until... when exactly?... then continue running the main thread, etc. But then we're taking on the specification burden for a deterministic profile that may or may not really want threads. The safer choice is to prevent `thread.spawn` by validation for now and if we really need `thread.spawn` to be available then undo that and specify how deterministic threads should work. The other way around seems more problematic: allow `thread.spawn` now and then reject it by validation at a later date &mdash; this would break things.

abrown: The above is written with assuming that engines would attempt to run Wasm threads in a single-threaded way, but on a closer reading that is probably not what @tlively meant: he probably meant `thread.spawn` would just always return `-1` for the deterministic profile, right? Both are valid options I guess...

design/mvp/CanonicalABI.md
abrown: Yeah, I think this is just an oversight of mine. This has been brought up in the past (e.g., [here](https://github.com/WebAssembly/shared-everything-threads/discussions/14)) and I think it does make sense to provide an escape valve for cases where the system can't handle the thread burden. I'll change `$st` to allow it to return a `-1` error like `memory.grow` does.

abrown: The challenge with spawning `n` threads is that we cannot return `-1` at some point in the future &mdash; we have to return it now. So, if some engine were to choose to allocate threads lazily they would have to be able to do the resource calculations at the `thread.spawn` point in order to return `-1`. This does not sound like a problem if the engine is using a thread pool.

design/mvp/CanonicalABI.md
tlively: > he probably meant `thread.spawn` would just always return `-1` for the deterministic profile, right?

Yes, exactly.

lukewagner: Yep, dynamic failure also makes sense; I was mostly noting the infallibility of the operation (which I see you've now fixed).

design/mvp/CanonicalABI.md
```diff
@@ -1697,7 +1697,68 @@ def canon_resource_rep(inst, rt, i):
...
+    the [shared-everything threads] proposal.
+
+In pseudocode, `$st` looks like:
+
+```python
+def canon_thread_spawn(ft, f, n, c):
+  assert(not DETERMINISTIC_PROFILE)
+  trap_if(f is None or ft is not f.type)
+  if not can_spawn(n):
+    return -1
```
lukewagner, line 1735:
```suggestion
  trap_if(f is None or ft is not f.type)
  if DETERMINISTIC_PROFILE:
    return -1
```

design/mvp/CanonicalABI.md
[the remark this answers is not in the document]
lukewagner: Yep!  Thanks

design/mvp/CanonicalABI.md
lukewagner: Yes, I think reporting failure would imply that the engine must eagerly do the resource allocation.  But even worse, to the other half of my question ("is failure atomic"), if we want creation of N threads to report an atomic success-or-failure result, that means we can't start observably running 2 threads, then fail creating the third, and then return failure because now we have 2 running threads (and we can't just hard-abort them mid-wasm-execution either).  Naively, this would require putting some mutex at thread start so that you could create all N threads and only start running them once all N were created, but of course this seems like a potential performance problem.

Thus, it feels like we need to explore other options.  E.g., maybe `thread.spawn` returns the number of successfully-started threads (in the range `[0, num-requested]`).  But is that enough to correctly implement non-leaky graceful handling of failures?

In thinking through this question (and how I'd write code that cleans up per-thread-allocations on failure), I realized a more basic question: if you use `thread.spawn` with `n > 1`, is it sufficient to pass every thread the same closure parameter, or would you often want to pass some extra parameter that is different for each thread?  I've only ever used the `n=1` case through, e.g., `pthread_create`, so I might just be ignorant of the use case for `n>1`, but it seems like you would.  Thinking about this more, I wonder if we really need `n` or if a single-shot `thread.spawn` would be sufficient (esp. if it is inlined), which would then nicely avoid this whole atomicity/locking question.

WDYT?

rossberg: I agree with @lukewagner that we should start with a basic unary thread.spawn, for the reasons he mentions. It is simpler and avoids usability questions, and could potentially be more expressive.

mingqiusun: I like the idea of returning "the number of successfully-started threads (in the range [0, num-requested])". In addition, I think there needs to be a way to return a thread-ID for each running thread.

abrown: Get ready for a long post!

First, what resources are we talking about? My understanding from @lukewagner's comment is that we're talking about memory allocations. We have:
a. a few runtime bits (the constant-size values of `f`, `c`, `n`), known when `thread.spawn` is invoked
b. the WebAssembly resources (e.g., the WebAssembly stack, the WebAssembly thread-locals); these will not only be the same for each of the `n` threads but will be a known size when `thread.spawn` is invoked
c. the actual OS thread resources (if an OS thread is what an implementor chooses), of unknown size

Now, we can allocate these resource classes at different times, but let's lump them together into _eager_ or _lazy_ schemes: I think that in either case we can avoid allocation failures:
- one _eager_ allocation scheme is a __thread pool__: any OS resources, (c), are allocated even before instantiation, while the WebAssembly resources, (a) and (b), are allocated when `thread.spawn` is invoked. (a) and (b) are allocated into the thread pool's queue, presumably on the heap; if we run out of memory or reach some engine limit we can report that fact immediately via a ~thread.spawn~ error. The WebAssembly threads that don't immediately get a pool thread wait in the queue to take the place of threads that finish. No new allocations are needed. This is how thread pools work today: these kinds of resource errors happen at enqueue time, not later.
- but I think this _lazy_ allocation scheme also works: imagine we only store (a) and dynamically allocate (b) and (c). At `thread.spawn` time, if we can spawn all `n` threads then there is no problem; but if we can only successfully spawn `m < n` threads for some reason (e.g., engine limit), we know that we have at least `m * (b + c)` space available--the spawn operations succeeded. Recall that each instance of (b) will be exactly the same size and assume the same for (c): once an active thread is finished, one of the leftover `m` threads can take its place--it uses the same amount of resources!

I suspect that the sweet spot is somewhere between the above _eager_ and _lazy_ schemes above: e.g., use a thread pool to avoid variation in (c) but reuse the finished thread's (b) space. In any case, it seems possible to "over-subscribe" `thread.spawn` with one of these schemes and use resources as they become available because WebAssembly gives us the sizes of (a) and (b). For any type of resource management that is more complex than this, I believe the answer should be provided by libraries within the WebAssembly module; I haven't thought too much about whether returning `[0, num-requested)` is enough for such libraries but it's a start.

-----

Next, @lukewagner brings up per-thread identification: in other words, how does a thread know what part of the problem to work on? This is a key bit of information for many algorithms, especially those that allow `n` threads to start together. In my original draft, I proposed a `thread.id` primitive that would provide exactly this. I would prefer that approach, but I have been trying (perhaps too hard) to be minimal in the changes I'm proposing, so I have been resorting to WebAssembly-level tricks. Without `thread.id`, each thread can atomically increment a `shared` global--each thread gets a new ID and stores it away in a TLS `$tid` slot for later use.

Now, this counter approach isn't very satisfying, especially when one considers multiple invocations of `thread.spawn`, each of which can spawn `n` threads. I suppose there is some way to make this work exclusively in WebAssembly, but I had always hoped for a `thread.id` that could return `(i32, i32)`, where the first value identifies the invocation and the second the particular thread within it. I haven't been particularly focused on the per-thread identification, though, so maybe we can discuss that in a separate PR.

-----

Finally, @rossberg questions the premise of spawning `n` threads at all:

> I agree with @lukewagner that we should start with a basic unary thread.spawn, for the reasons he mentions. It is simpler and avoids usability questions, and could potentially be more expressive.

I agree it would be simpler and hopefully you see that I've addressed some of the usability concerns. Spawning a single thread is not more expressive, though: one can always write `thread.spawn 1` for the unary case but without `n` one cannot communicate to the machine how many threads start at the same time. This bit of information is required in fork-join paradigms (e.g., OpenMP), in non-CPU configuration (e.g., GPU kernels), in parallel language constructs (e.g., [Julia](https://docs.julialang.org/en/v1/base/multi-threading/)), in compiler optimizations (e.g., auto-vectorizations), etc. All of these need `n` in some way: e.g., GPUs use `n` to assume "`n` threads _must_ start at the same time;" OpenMP assumes that "`n` threads _may_ start at the same time" (hand-waving: many ways to configure OpenMP!). In WebAssembly, we should choose the latter semantics, "`n` threads _may_ start at the same time" for portability reasons. But I'm convinced it makes sense to communicate this bit of information to the runtime: it's not just a convenience, it's critical information for runtime and hardware optimization. From my unique perspective at a hardware company, I see more of a need for this kind of thing, not less.

rossberg: For example, a per-thread primitive could be more expressive because e.g. it could return a thread id.

Wasm's current design is clearly not optimised for GPUs on many levels, so I don't think that is a concern that should take high priority all of a sudden. It seems to me that more substantial additions would be necessary to get there, and an n-ary spawn may be one such consideration once we decide to go that down route, but that's gonna be a whole different story. (I also believe a careful consideration of the differences between concurrency and parallelism is needed for that.)

lukewagner: Thanks for all the background on your thinking @abrown.  I'll start with the main remaining question it leaves me with b/c that probably influences the other questions/answers:

I see how the `shared` global counter scheme you described allows you to approximate a `thread.id` (which, incidentally, I agree is a good idea for core wasm to have...).  However, when I imagine spawning N threads, I imagine that I have some input data set and I want to split it up and hand a chunk to each of N threads.  So, before the `thread.spawn`, I want to compute how to chunk it (e.g., using `thread.hw_concurrency`) and compute pointers to each chunk, and then I want to pass these distinct chunk pointers into each of the N distinct threads.  Even using your global-`shared`-`thread.id`-approximation scheme, this still seems unnecessarily hard/expensive (and like it involves some locking to coordinate between the pre-`thread.spawn` code and the spawned threads, noting that, in general, there can be multiple concurrent `thread.spawn`s in progress at any one time).  Fundamentally, it seems like I want to be able to manually pass a distinct scalar value to each spawned thread.  You can of course imagine passing in an N-length array to `thread.spawn` (with separate variants for a linear-memory array vs. a wasm-gc `array`), but this starts to sound like a lot of trouble compared to just a single-thread `thread.spawn`.

abrown: I think the assumption that one would want to split the problem into chunks and hand one to each thread _before_ spawning does not always hold. Because in these situations we usually want to maximize the amount of work done in the parallel section, most of the "where does this thread start working" logic happens in the thread itself based on something that boils down to `thread.id`. Take this [DGEMM](https://github.com/ParRes/Kernels/blob/3ba090863d663fd5e3fda7697bbae3cc24236dc9/OPENMP/DGEMM/dgemm.c#L221-L246) OpenMP example: the global `block`, `order`, and matrix regions (`A`, `B`, `C`) are calculated beforehand but the `*_arr` calculations &mdash; where the thread should do its work &mdash; happen in each thread. Or this [reduce](https://github.com/ParRes/Kernels/blob/3ba090863d663fd5e3fda7697bbae3cc24236dc9/OPENMP/Reduce/reduce.c#L244) example: the part of the vector to process is calculated directly from OpenMP's version of `thread.id`. I tend to believe the most common use case is to pass a base pointer and a size to each thread and use `thread.id` to set up individual pointers in the thread rather than pre-allocating an array of pointers, one for each thread.

But let's say one _did_ want to pre-allocate all the chunks into an N-length array and then spawn N threads, each pointing at its own chunk. In a C world, wouldn't it be enough to pass the base pointer to the chunk array to the `thread.spawn` and have each thread pick out its chunk using the `thread.id` as the index? It seems the same works fine for GC arrays, though I'm not as familiar with GC (just extra array references, not copies, right?). Indexing out the thread context using the thread ID is flexible enough to support even the "pass a distinct scalar to each spawned thread" case.

tlively: That all assumes there is an instruction like `thread.id` in the first place. In the absence of that, it would be most efficient to have the spawning thread pass the thread ID as an argument to each thread it spawns, and they can proceed to compute divisions of the work based on that argument. Passing a different thread ID to each thread would be best done spawning one thread at a time, though.

On the other hand, if there is a `thread.id` instruction, then the producer probably wouldn't control its value (otherwise why not just manage it in userspace entirely?), so it wouldn't be directly usable as an array index like you're suggesting.

It's also notable that the thread spawning syscalls in POSIX and Windows all spawn just one thread at a time, so spawning just one thread is consistent with the lowest-level native primitives.

lukewagner: +1 to @tlively's points.  Overall, n-way `thread.spawn` seems like an optimization over the base-case of a scalar `thread.spawn` so it seems like we should start with the base case, optimize it fully and then use a real-world workload to measure and motivate making an optimized n-way version, to validate that we are in fact solving a problem in exchange for the additional complexity.

abrown: > On the other hand, if there is a `thread.id` instruction, then the producer probably wouldn't control its value (otherwise why not just manage it in userspace entirely?), so it wouldn't be directly usable as an array index like you're suggesting.

No, I think the ID returned would need to be one that _can_ be used to index into such an array as is done in other parallel frameworks. How these IDs would be generated would necessarily be part of the specification of `thread.id`; I think `[0, n)` is a reasonable start.

> It's also notable that the thread spawning syscalls in POSIX and Windows all spawn just one thread at a time, so spawning just one thread is consistent with the lowest-level native primitives.

Yes, but I'm more concerned with enabling optimizations for a programming level higher than that. The ability to execute "parallel for" blocks is hardly uncommon in parallel libraries and frameworks. WebAssembly need not be restricted here to OS-level APIs: we wouldn't have the GC proposal, e.g., if that were the case. I feel that an n-way `thread.spawn` is a good in-between: it presents little additional complexity in the `thread.spawn 1` case but also allows users to provide important information to the runtime and HW.

> it seems like we should start with the base case, optimize it fully and then use a real-world workload to measure and motivate making an optimized n-way version

I'm struggling with this: this could also be done in the reverse order. Since I am more interested in the n-way spawn, another approach is to investigate that more thoroughly (e.g., with interested partners) and downgrade to the single spawn if there truly is no interest. I certainly don't want to argue the case for this indefinitely; it makes it seem like this idea is more controversial than it probably is.

lukewagner: > I'm struggling with this: this could also be done in the reverse order.

As a general rule, though, we try start simple and then measure before adding optimized versions of a thing.  Given that N-way `thread.spawn` opens up a number of semantic/design questions, it seems like a good candidate for starting simple.  After all, it's not like N-way `thread.spawn` is O(1) -- it's still executing a loop internally, so the question is: how much faster is it to do that loop inside the engine than to execute `thread.spawn` in a wasm loop and is that speedup significant in the context of a real workload?  I dunno, maybe it's significant, but it seems like, at least in the short-term before we have optimized implementations, we should start with the N=1 case.

badeend:
I assume that any parallelism introduced by these built-ins are contained to the component instance that spawned them, and that cross-component communication (including client<->host) is still required to be fully sequential in order to comply with the component invariants.

It might be worthwhile to throw in a quick clarification of this under the new Threads section in the explainer.

lukewagner:
@badeend That's correct yes.  A future (more extensive) extension to the C-M would be to add `shared` attributes to component-level functions so that they could be lifted-from or lowered-to `shared` core functions but, until we add that, lowering a (non-shared) component-level function necessarily produces a (non-shared) core function which does end up forcing serialization back to the "main" (entry) thread for cross-component calls.

design/mvp/CanonicalABI.md
abrown: Ok, after some delay, I'm back! Looks like I got out-voted here, so I pushed 821f2c2 to remove the "spawn n" bits.

design/mvp/CanonicalABI.md
```diff
@@ -1697,7 +1697,68 @@ def canon_resource_rep(inst, rt, i):
...
+(canon thread.spawn (type $ft) (core func $st))
+```
+validation specifies:
+* `$ft` must refer to a `shared` function type; initially, only the type `(func
+  shared (param $c i32))` is allowed (see explanation below)
+* `$st` is given type `(func (param $f (ref null $ft)) (param $c i32) (result $e
+  i32))`.
+
+In the [deterministic profile], validation fails if `thread.spawn` is used.
+
```
lukewagner, line 1713:
```suggestion
```
IIRC, we switched to a dynamic failure mode and thus use of `thread.spawn` *validates*, it just fails at runtime.

design/mvp/CanonicalABI.md
```diff
@@ -1697,7 +1697,68 @@ def canon_resource_rep(inst, rt, i):
...
+```
+
+### 🧵 `canon thread.hw_concurrency`
+
+For a canonical definition:
+```wasm
+(canon thread.hw_concurrency (core func $f))
+```
+validation specifies:
+* `$f` is imported with type `(func shared (result i32))`.
```
lukewagner, line 1748:
```suggestion
* `$f` is given type `(func shared (result i32))`.
```

design/mvp/Explainer.md
```diff
@@ -1219,7 +1220,12 @@ canon ::= ...
         | (canon resource.new <typeidx> (core func <id>?))
         | (canon resource.drop <typeidx> (core func <id>?))
         | (canon resource.rep <typeidx> (core func <id>?))
+        | 🧵 (canon thread.spawn <typeidx> (core func <id>?))
+        | 🧵 (canon thread.hw_concurrency (core func <id>?))
```
lukewagner, line 1224:
```suggestion
        | (canon thread.spawn <typeidx> (core func <id>?)) 🧵
        | (canon thread.hw_concurrency (core func <id>?)) 🧵
```

design/mvp/Explainer.md
```diff
@@ -1262,6 +1268,19 @@ Here, the `i32` returned by `resource.new`, which is an index into the
...
 
+##### 🧵 Threads
+
+The [shared-everything-threads] proposal adds component model built-ins for
+thread management. These are specified as built-ins and not core WebAssembly
+instructions because browsers expect this functionality to come from existing
+Web/JS APIs.
+
+The `thread.spawn` built-in has type `[f:(ref null $f) c:i32] -> []` and spawns
+a new thread by invoking the shared function `f` while passing `c` to it.
```
lukewagner, line 1279:
```suggestion
The `thread.spawn` built-in has type `[f:(ref null $f) c:i32] -> [i32]` and spawns
a new thread by invoking the shared function `f` while passing `c` to it, returning
whether a thread was successfully spawned.
```

design/mvp/Explainer.md
```diff
@@ -1262,6 +1268,19 @@ Here, the `i32` returned by `resource.new`, which is an index into the
...
+
+The `resource.hw_concurrency` built-in has type `[i32] -> []` and returns the
+number of threads can be expected to execute concurrently.
```
lukewagner, line 1282:
```suggestion
The `resource.hw_concurrency` built-in has type `[] -> [i32]` and returns the
number of threads that can be expected to execute concurrently.
```

design/mvp/Explainer.md
```diff
@@ -1943,3 +1962,5 @@ and will be added over the coming months to complete the MVP proposal:
 
 [Scoping and Layering]: https://docs.google.com/presentation/d/1PSC3Q5oFsJEaYyV5lNJvVgh-SNxhySWUqZ6puyojMi8
 [Future and Stream Types]: https://docs.google.com/presentation/d/1MNVOZ8hdofO3tI0szg_i-Yoy0N2QPU2C--LzVuoGSlE
+
+[shared-everything-threads]: https://github.com/WebAssembly/shared-everything-threads
```
lukewagner, line 1966: nit: could you move this up to be right under "[gc]" (at the end of all the core wasm proposals)?

design/mvp/CanonicalABI.md
```diff
@@ -1697,7 +1697,68 @@ def canon_resource_rep(inst, rt, i):
...
+> use `$c` to address a memory region containing multiple values (the current
+> [wasi-libc convention]), the initial version restricts `$ft` to `[i32] -> []`.
+> The inclusion of `$ft` ensures backwards compatibility for when arbitrary
+> parameters are allowed.
+
+Calling `$st` checks that the reference `$f` is not null and satisfies type
+`$ft`. Then, it spawns a thread which:
+  - invokes `$f` with `$c`
+  - executes `$f` until completion or trap in a `shared` context as described by
+    the [shared-everything threads] proposal.
```
lukewagner, line 1726: Could you add a note that an uncaught wasm exception in this new thread turns into a trap (and do likewise in the pseudo-code below, wrapping the invocation of `f(c)` in a `try ... except CoreWebAssemblyException: trap()` (see [`canon_lift`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#canon-lift) which does likewise))).

lukewagner (approved):
Ah nice, that closes out a lot of questions; thanks!  LGTM w/ these nits fixed.  I'll wait another week or so for any other feedback before merging.

design/mvp/CanonicalABI.md
```diff
@@ -1697,7 +1697,68 @@ def canon_resource_rep(inst, rt, i):
...
+* `$st` is given type `(func (param $f (ref null $ft)) (param $c i32) (result $e
+  i32))`.
+
+In the [deterministic profile], validation fails if `thread.spawn` is used.
+
+> Note: ideally, a thread could be spawned with [arbitrary thread parameters].
+> To import polymorphic versions of `$st`, a naming scheme is necessary to
+> differentiate between the imports with varying `$ft`. Since many languages can
+> use `$c` to address a memory region containing multiple values (the current
+> [wasi-libc convention]), the initial version restricts `$ft` to `[i32] -> []`.
```
rossberg, line 1718: This explanation is very confusing. Of course, a different $ft will produce a core function $st of corresponding type, but at no point would _that_ be polymorphic. I guess you're hinting at polymorphism in the spawn "builtin" itself, which under certain (though not all!) implementation strategies of the component model could somehow be lowered to a core Wasm import? But that's not what this is saying.

design/mvp/CanonicalABI.md
abrown: Take a look at df0f8b1; I think that's what you're looking for.

design/mvp/CanonicalABI.md
abrown: I see what you mean. How about the following:

> Note: ideally, a thread could be spawned with arbitrary thread parameters. This would require the `thread.spawn` builtin to be polymorphic when lowered as a core WebAssembly import; i.e., in current implementations of the component model, this could mean implementing a naming scheme to distinguish which import matches each used `$ft` type. Since many languages can use `$c`...

design/mvp/CanonicalABI.md
lukewagner: Perfect, thanks!

design/mvp/Explainer.md
```diff
@@ -1262,6 +1268,20 @@ Here, the `i32` returned by `resource.new`, which is an index into the
 component's handle-table, is immediately returned by `make_R`, thereby
 transferring ownership of the newly-created resource to the export's caller.
 
+##### 🧵 Threads
+
+The [shared-everything-threads] proposal adds component model built-ins for
+thread management. These are specified as built-ins and not core WebAssembly
+instructions because browsers expect this functionality to come from existing
+Web/JS APIs.
```
yamt, line 1276: to me, it seems very awkward to make this tie with component-model.
can you explain a bit more?
do you consider it to belong to the canonical abi for some reasons?
what do you mean by "come from existing Web/JS APIs"?
how does it prevent this from being core wasm instructions?

lukewagner: Yes, ideally `thread.spawn` would be a core instruction included in the [shared-everything-threads proposal](https://github.com/webAssembly/shared-everything-threads), and that's what we initially pushed for (from a wasm-outside-the-browser perspective).  However, there was a lot of pushback from browsers since a core wasm `thread.spawn` instruction would add a ton of work in a Web Platform context, so `thread.spawn` was removed (at least for now, maybe in a future proposal it could be added to core wasm).  The next-best-place is to add `thread.spawn` as a Component Model built-in so that it can be used from core wasm as a core function import (without knowing whether the import is implemented by either existing Web Platform primitives or, outside the browser, the `thread.spawn` built-in.

yamt: >  However, there was a lot of pushback from browsers since a core wasm thread.spawn instruction would add a ton of work in a Web Platform context,

i guess component-model is a ton of work for the other set of platforms.

lukewagner: Technically, you can ignore the component model if you only want to run a single core module by having the core wasm runtime just implement `thread.new` directly as a host built-in function with the core function signature defined here.

design/mvp/CanonicalABI.md
lukewagner: I think [this](https://github.com/WebAssembly/component-model/pull/291/commits/336952e7e7452e30579e1567259611534494d287) just-merged suggestion addresses this feedback, so I'll resolve this, but feel free to reopen if not.

lukewagner:
I think all the actionable feedback has been addressed, so I'll plan to merge at the end of the week unless anything new comes up.

PR #292, opened 2024-01-08
Python: fix invalid type annotations and undefined variables

orsinium:
1. Fix invalid document fragment names in ToC.
2. Use the correct syntax in type annotations for lists. It should be `list[int]` instead of `[int]`.
3. Replace `opts` with `cx.opts` in a few functions where `opts` variable is undefined.


lukewagner:
Thanks!

PR #298, opened 2024-01-29
Document and define WIT IDs that are referenced elsewhere

ggoodman:
As a newcomer coming to the Component Model, I was struggling to understand what a ID _meant_ in the context of a WIT file. To help myself and others following in my footsteps, I took a stab at writing down what I thought I understood.

I don't expect this to be correct but hope we can use it as the starting point to get a solid definition in the reference doc.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -65,6 +65,19 @@ in the Component Model's representation of [`interface`s][interfaces] and
 
 [names of imports and exports]: Explainer.md#import-and-export-definitions
 
+## WIT IDs
+
+Leaf elements in WIT files have IDs that are references to their canonical
+definitions. A WIT ID includes the WIT package specifier and path within that
+uniquely defines the element.
+
+For example, the `wasi:clock@0.2.0` world imports the `wall-clock` interface
+which has a `now: func() -> datetime` method. The ID for this method would be
+`wasi:clock@0.2.0/wall-clock/now`.
+
+When different worlds are combined, the canonical IDs of leaf
+nodes will be used for deduplication.
+
 ## WIT Interfaces
 [interfaces]: #wit-interfaces
 


design/mvp/WIT.md
```diff
@@ -65,6 +65,19 @@ in the Component Model's representation of [`interface`s][interfaces] and
...
 
+## WIT IDs
+
+Leaf elements in WIT files have IDs that are references to their canonical
+definitions. A WIT ID includes the WIT package specifier and path within that
+uniquely defines the element.
+
+For example, the `wasi:clock@0.2.0` world imports the `wall-clock` interface
+which has a `now: func() -> datetime` method. The ID for this method would be
+`wasi:clock@0.2.0/wall-clock/now`.
```
lukewagner, line 76: FWIW, the version string is specified to go at the end, so it'd be `wasi:clock/wall-clock@0.2.0`.

lukewagner (commented):
Hi, thanks for taking the time to suggest improvements here!  One thing to note is that there are already several sections that discuss identifiers ("Package identifiers", "WIT identifiers" and "Identifiers"; I wonder if the latter two could be collapsed...).  Thus, we probably don't want to add a new section but rather augment an existing one.

ggoodman:
Hi @lukewagner thanks for taking a look. The perceived gap in the document was understanding how deduplication happens. My understanding was that the deduplication happened on an ID basis. In the [WIT Worlds](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#wit-worlds) section, we start referencing 'IDs' but I couldn't find any clear reference of what these IDs point to and what their components are. I may simply not have read attentively enough. I thought it would be helpful to have a little glossary-like section explaining the concept before it started getting used.

lukewagner:
Yes, that makes sense and again I appreciate the work to improve the docs.  Perhaps you could update the PR to add the new information to one of the existing sections (and add links to those sections where you think they're missing)?

lukewagner:
I think #339 improves the situation for WIT identifiers, so I'll close this PR, but feel free to reopen if there's further changes you'd like to make.

PR #300, opened 2024-02-05
Allow start functions to call imports.

sunfishcode:
Remove the restriction on component start functions calling imports. This allows start functions to run artbitrary user code.


lukewagner:
Based on discussion in #297 (which this was forked from), I think we can merge this, but I'll wait a day or so.

Issue #181, opened 2023-04-11
Fixed length arrays


LLFourn:
Rust has fixed length arrays whose type is written as `[u8;32]` for a 32-byte array. You can achieve this in the current wit specification by doing `tuple<u8,u8 ... u8>` (with 32 `u8`s). Could `list<u8,32>` or something be a first class type to avoid abusing tuples in this way even if it desugars to something similar?

lukewagner:
That does seem like a potentially good idea, and could be defined as another "specialization" (just like how `tuple` itself is a specialization of `record`).  I feel like this was brought up in a previous issue or comment somewhere, but I can't seem to find it at the moment.

sunfishcode:
A similar issue that came up earlier was IPv6 addresses, which would want to use `[u16; 8]`.

lukewagner:
Oh right, that was it, thanks.  I suppose mostly this is just a question of prioritization and whether we think there are enough use cases to motivate adding this in the MVP time frame.

cameron1024:
Just stumbled across this - I'd be very interested in these types for cryptographic hashes, in case there's a need for more motivating examples

f5yacobucci:
Seconding @cameron1024 use case. I've been trying to understand wit, components, and wit-bindgen using the md5 algorithm. Being able to create `md5.buffer: list<u8, 64>` would be desireable.

LLFourn:
This actually happened to be my use-case too. I have ids for certain resources that are referenced by host and guest that are the output of cryptographic hashes so always have a fixed length. It's a bit of a PITA having to check the length and copy the bytes into a fixed length array on both sides.

lukewagner:
It does seem like there are some good use cases for this and it's not a big addition if this is just a specialization of `tuple`, so it does seem like a good idea to add.  That being said, I think maybe we should hold off on adding this before the [Preview 2 milestone](https://github.com/WebAssembly/component-model/#milestones) since there is already a bunch of other loose ends to wrap up in the milestone.  But after that, I'd be happy to propose adding it.

badeend:
For reference: https://github.com/WebAssembly/interface-types/issues/146. I think that issue was raised before the existence of tuples, though.

ydnar:
I’m working on the Go bindings for WASI Preview 2 and the Component Model. The `ipv4-address` and `ipv6-address` types seem to be reasonable justification for fixed-with array types.

```wit
    type ipv4-address = tuple<u8, u8, u8, u8>;
[...]

PR #304, opened 2024-02-19
Add optional fixed length to lists

lukewagner:
This PR adds fixed-length lists as a "gated feature" (i.e., features spec'd and ready to implement, but not yet included in a Preview release), as proposed and discussed in #181.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -12,7 +12,7 @@ rules, but rather merge the minimal need-to-know elements of both, with just
 enough detail to create a prototype. A complete definition of the binary format
 and validation will be present in the [formal specification](../../spec/).
 
-See the [explainer introduction](Explainer.md) for an explanation of 🪙.
+See [Gated Features](Explainer.md#gated-features) for an explanation of 🪙 and 🔧.
 
 
 ## Component Definitions
@@ -184,6 +184,7 @@ defvaltype    ::= pvt:<primvaltype>                       => pvt
                 | 0x72 lt*:vec(<labelvaltype>)            => (record (field lt)*)    (if |lt*| > 0)
                 | 0x71 case*:vec(<case>)                  => (variant case*)
                 | 0x70 t:<valtype>                        => (list t)
+                | 0x67 t:<valtype> len:<u32>              => (list t len) 🔧
                 | 0x6f t*:vec(<valtype>)                  => (tuple t+)    (if |t*| > 0)
                 | 0x6e l*:vec(<label'>)                   => (flags l+)    (if |l*| > 0)
                 | 0x6d l*:vec(<label'>)                   => (enum l*)
@@ -371,6 +372,12 @@ nameassoc ::= idx:<u32> name:<name>
 where `namemap` is the same as for core wasm. A particular `sort` should only
 appear once within a `name` section, for example component instances can only be
 named once.
+
+
+## Binary Format Warts to Fix in a 1.0 Release
+
+* The two `list` type codes should be merged into one with an optional immediate.
+* The `0x00` prefix byte of `importname'` and `exportname'` will be removed or repurposed.
 
 
 [`core:u32`]: https://webassembly.github.io/spec/core/binary/values.html#integers
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -99,11 +99,21 @@ def alignment(t):
     case Float32()          : return 4
     case Float64()          : return 8
     case Char()             : return 4
-    case String() | List(_) : return 4
+    case String()           : return 4
+    case List(t, l)         : return alignment_list(t, l)
     case Record(fields)     : return alignment_record(fields)
     case Variant(cases)     : return alignment_variant(cases)
     case Flags(labels)      : return alignment_flags(labels)
     case Own(_) | Borrow(_) : return 4
+```
+
+List alignment is the same as tuple alignment when the length is fixed and
+otherwise uses the alignment of pointers.
+```python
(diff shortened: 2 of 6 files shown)


badeend:
I'm not too familiar with the ABI, but can a large fixed-size List blow the stack? If so, it might be an idea to fall back to heap allocation after a certain size.

ydnar:
I’m glad to have fixed-width arrays added to WIT.

However from the perspective of an ABI implementer, I struggle with the representation of fixed-length arrays as a variation of `list<T>`, rather than as a new standalone type that shares some similarities to both `tuple` and `list`.

- A fixed-length list shares nothing with list other than name (ABI representation, size, alignment, flattening and lift/lower rules).
- Where previously the size and alignment of `list` were constant (8 and 4, respectively), this adds complexity that every WIT/Component Model implementer will need to address.

Concretely, I’d vote for a new type, let‘s call it `array`, that works exactly as described in this PR, but decouples its logic from `list`, leaving `list` unchanged from the existing spec. An `array` would unequivocally be a value type with non-optional `type` and `len`, leaving `list` as a type with a pointer to allocated storage.

The WIT syntax for `array` could be anything:

- `array<u8, 4>`
- `[4]u8`
- `[u8, 4]`

lukewagner:
I think we shouldn't let the CABI representational concerns determine the WIT- and Component-Model-level names we choose; rather we should just focus on what's the least confusing things to folks reading/writing WIT.

Perhaps there is an argument that having both `list<u8>` and `list<u8,4>` share the same type-name "`list`" is confusing and so having different type names for the variable- and fixed-length cases is less confusing?  I don't think so, but this is admittedly subjective.  There is some precedent we can observe in Go and Rust, though: the syntax of the fixed-length thing (`[T; N]` in Rust, `[N]T` in Go) is a syntactic extension of the dynamic-length thing (`[T]` in Rust, `[]T` in Go).  Thus, there is evidence that having the fixed-length construct syntactically refine the variable-length construct is not confusing.  But I haven't done an exhaustive survey of languages or anything.

Overall, while I'm not opposed to having `array<T,N>`, I think less names is better, as long as we're not introducing confusion.  But I'd be interested to hear more thoughts from folks on this.

oovm:
I don't think list contains any mutability constraints, neither capacity mutability nor internal element mutability.

Without any numbers, it means that this is a list with unknown capacity (unknown at compile time)

That is:

```js
list<T>     = list<T, ?>
list<u8, 4> = tuple<u8, u8, u8, u8>
```

Mutability needs to be passed via abi options and verified

For example `list<u8, 4>` can be:

- dynamic passing: (ptr: i32, len: i32)
- static passing: (ptr: i32)
- inline passing: (i32, i32, i32, i32)
- Pass by reference: (arrayref)

Even though `length` and `size_of(u8)` are known at compile time, they will still be passed in by host if required.

ydnar:
> Therefore I suggest adding an abi option `static-length`, requiring the language to interpret it as an `array`, to be compatible with some languages ​​that do not have the concept of `array`(fixed length list).
> 
> Considering that `array` is a feature widely present in popular languages, it should require static by default, and then add the reverse `dynamic-length` ​

How are `tuple` handled in those languages today?

oovm:
@ydnar 

For tuples, according to my classification, the current working mode is `inline`.

In other words, if `list<u8, 4>` is the syntactic sugar of `tuple<u8, u8, u8, u8>`, `(i32, i32, i32, i32)` will be passed in.

Unless the input parameters are greater than 16 or the return value is greater than 1, `dynamic-length` mode will be used.
​
```wit
package examples: tests;

world imports {
  import tuples;
}

interface tuples {
  tuple0: func(x: tuple<>);
  tuple1: func(x: tuple<u8>);
  tuple2: func(x: tuple<u8, u8>);
  tuple4: func(x: tuple<u8, u8, u8, u8>);
  tuple16: func(x: tuple<u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8>);
  tuple17: func(x: tuple<u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8, u8>);

  tuple-r0: func() -> tuple<>;
  tuple-r1: func() -> tuple<u8>;
  tuple-r2: func() -> tuple<u8, u8>;
}
```


```wat
(module
  (type (;0;) (func))
  (type (;1;) (func (param i32)))
  (type (;2;) (func (param i32 i32)))
  (type (;3;) (func (param i32 i32 i32 i32)))
  (type (;4;) (func (param i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32)))
  (type (;5;) (func (result i32)))
  (type (;6;) (func (param i32 i32 i32 i32) (result i32)))
  (import "examples:tests/tuples" "tuple0" (func (;0;) (type 0)))
  (import "examples:tests/tuples" "tuple1" (func (;1;) (type 1)))
  (import "examples:tests/tuples" "tuple2" (func (;2;) (type 2)))
  (import "examples:tests/tuples" "tuple4" (func (;3;) (type 3)))
  (import "examples:tests/tuples" "tuple16" (func (;4;) (type 4)))
  (import "examples:tests/tuples" "tuple17" (func (;5;) (type 1)))
  (import "examples:tests/tuples" "tuple-r0" (func (;6;) (type 0)))
  (import "examples:tests/tuples" "tuple-r1" (func (;7;) (type 5)))
  (import "examples:tests/tuples" "tuple-r2" (func (;8;) (type 1)))
)

```

cpetig:
I think that with #384 merged this PR can be closed as a duplicate?

lukewagner:
Oh right, thanks!

PR #317, opened 2024-03-13
working with gc types

### LINKED ISSUE #305: Working with GC types without copy
I'm having some trouble switching to wasi preview 2.

For example, the following interface:

```wit
package wasi:random@0.2.0;
interface random {
    get-random-bytes: func(len: u64) -> list<u8>;
}
```

The function signature is `func (u64) -> (list<u8>)`

But its lower type is `core func (i64, i32) -> ()`, which is very difficult to use.

If I want to convert it to `core type (array (mut u8))`, a very long glue code is required.

---
I hope to add a GC mode canon option that can make the lower type similar to `core func (i64) -> (array u8)`.

For complex nested types, getting the specified data requires very complex pointer algebra, whereas if using array it only requires multiple `array.get`.

I think this helps simplify the use of some external interfaces, such as:

```wit
package wasi:filesystem@0.2.0;
interface preopens {
    get-directories: func() -> list<tuple<descriptor, string>>;
}
```

oovm:
Add Canonical ABI options to support passing parameters by reference type

- Close #305

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -22,7 +22,7 @@ See the [explainer introduction](Explainer.md) for an explanation of 🪙.
 component ::= <preamble> s*:<section>*            => (component flatten(s*))
 preamble  ::= <magic> <version> <layer>
 magic     ::= 0x00 0x61 0x73 0x6D
-version   ::= 0x0a 0x00
+version   ::= 0x0d 0x00
 layer     ::= 0x01 0x00
 section   ::=    section_0(<core:custom>)         => ϵ
             | m:section_1(<core:module>)          => [core-prefix(m)]
@@ -36,19 +36,23 @@ section   ::=    section_0(<core:custom>)         => ϵ
             | s: section_9(<start>)               => [s]
             | i*:section_10(vec(<import>))        => i*
             | e*:section_11(vec(<export>))        => e*
+            | v*:section_12(vec(<value>))         => v* 🪙
 ```
 Notes:
 * Reused Core binary rules: [`core:section`], [`core:custom`], [`core:module`]
 * The `core-prefix(t)` meta-function inserts a `core` token after the leftmost
   paren of `t` (e.g., `core-prefix( (module (func)) )` is `(core module (func))`).
 * The `version` given above is pre-standard. As the proposal changes before
-  final standardization, `version` will be bumped from `0xa` upwards to
+  final standardization, `version` will be bumped from `0x0d` upwards to
   coordinate prototypes. When the standard is finalized, `version` will be
   changed one last time to `0x1`. (This mirrors the path taken for the Core
   WebAssembly 1.0 spec.)
 * The `layer` field is meant to distinguish modules from components early in
   the binary format. (Core WebAssembly modules already implicitly have a
-  `layer` field of `0x0` in their 4 byte [`core:version`] field.)
+  `layer` field of `0x0` if the existing 4-byte [`core:version`] field is
+  reinterpreted as two 2-byte fields. This implies that the Core WebAssembly
+  spec needs to make a backwards-compatible spec change to split `core:version`
+  and fix `layer` to forever be `0x0`.)
 
 
 ## Instance Definitions
@@ -72,8 +76,8 @@ core:inlineexport   ::= n:<core:name> si:<core:sortidx>                    => (e
 instance            ::= ie:<instanceexpr>                                  => (instance ie)
 instanceexpr        ::= 0x00 c:<componentidx> arg*:vec(<instantiatearg>)   => (instantiate c arg*)
                       | 0x01 e*:vec(<inlineexport>)                        => e*
-instantiatearg      ::= n:<string>  si:<sortidx>                           => (with n si)
-string              ::= s:<core:name>                                      => s
+instantiatearg      ::= n:<name>  si:<sortidx>                             => (with n si)
+name                ::= n:<core:name>                                      => n
 sortidx             ::= sort:<sort> idx:<u32>                              => (sort idx)
 sort                ::= 0x00 cs:<core:sort>                                => core cs
                       | 0x01                                               => func
@@ -96,7 +100,7 @@ Notes:
 * Validation of `core:instantiatearg` initially only allows the `instance`
   sort, but would be extended to accept other sorts as core wasm is extended.
 * Validation of `instantiate` requires each `<importname>` in `c` to match a
-  `string` in a `with` argument (compared as strings) and for the types to
+  `name` in a `with` argument (compared as strings) and for the types to
   match.
 * When validating `instantiate`, after each individual type-import is supplied
   via `with`, the actual type supplied is immediately substituted for all uses
@@ -111,7 +115,7 @@ Notes:
 (See [Alias Definitions](Explainer.md#alias-definitions) in the explainer.)
 ```ebnf
 alias       ::= s:<sort> t:<aliastarget>                => (alias t (s))
-aliastarget ::= 0x00 i:<instanceidx> n:<string>         => export i n
+aliastarget ::= 0x00 i:<instanceidx> n:<name>           => export i n
               | 0x01 i:<core:instanceidx> n:<core:name> => core export i n
               | 0x02 ct:<u32> idx:<u32>                 => outer ct idx
 ```
@@ -157,9 +161,6 @@ Notes:
   initially-empty type index space.
 * `alias` declarators currently only allow `outer` `type` aliases but
   would add `export` aliases when core wasm adds type exports.
-* Validation of `outer` aliases cannot see beyond the enclosing core type index
-  space. Since core modules and core module types cannot nest in the MVP, this
-  means that the maximum `ct` in an MVP `alias` declarator is `1`.
 
 ```ebnf
 type          ::= dt:<deftype>                            => (type dt)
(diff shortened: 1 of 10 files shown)


design/mvp/Explainer.md
```diff
@@ -1173,6 +1176,34 @@ The Canonical ABI will use `realloc` both to allocate (passing `0` for the
...
+| `u8`                 | `(i32,)`                    | `(i32,)`                         |
+| `u16`                | `(i32,)`                    | `(i32,)`                         |
+| `u32`                | `(i32,)`                    | `(i32,)`                         |
+| `u64`                | `(i32,)`                    | `(i32,)`                         |
+| `resource`           | `(ptr: i32)`                | `(ref extern,)`                  |
+| `list<T>`            | `(ptr: i32, len: i32)`      | `(ref array (mut $t),)`          |
+| `record { a, b }`    | `(a: A, b: B)` (flatten)    | `(ref struct $record,)`          |
+| `tuple<A, B>`(in⩽16) | `(a: A, b: B)` (flatten)    | `(a: A, b: B)`   (flatten)       |
+| `tuple<A, B>`(out>1) | `(ptr: i32)`                | `(ref struct (mut $a)(mut $b),)` |
+| `option<T>`          | `(i32, i32)`                | `(ref null $t,)`                 |
```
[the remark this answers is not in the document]
oovm: I looked at some WAML language compilation procedures and I think (scalar) boxing makes sense, if you want to avoid allocate, just don't add the option.

The only problem might be that you can't set different behaviors for two options in the same call, but I can't think of a practical example of doing this.

design/mvp/Explainer.md
```diff
@@ -1173,6 +1176,34 @@ The Canonical ABI will use `realloc` both to allocate (passing `0` for the
...
+| `char`               | `(i32,)`                    | `(i32,)`                         |
+| `u8`                 | `(i32,)`                    | `(i32,)`                         |
+| `u16`                | `(i32,)`                    | `(i32,)`                         |
+| `u32`                | `(i32,)`                    | `(i32,)`                         |
+| `u64`                | `(i32,)`                    | `(i32,)`                         |
+| `resource`           | `(ptr: i32)`                | `(ref extern,)`                  |
+| `list<T>`            | `(ptr: i32, len: i32)`      | `(ref array (mut $t),)`          |
+| `record { a, b }`    | `(a: A, b: B)` (flatten)    | `(ref struct $record,)`          |
+| `tuple<A, B>`(in⩽16) | `(a: A, b: B)` (flatten)    | `(a: A, b: B)`   (flatten)       |
+| `tuple<A, B>`(out>1) | `(ptr: i32)`                | `(ref struct (mut $a)(mut $b),)` |
```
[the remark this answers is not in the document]
oovm: My concerns about mutability mainly involve covariant arrays. A previous call may break the type safety guarantee of the external language, and then errors will occur in the subsequent code.

Although this possibility exists in pointer-passing linear memory interfaces, reference types with subtype relationships are more likely to cause this error inadvertently.

oovm:
> I realized that the discriminant actually is required for variants that have multiple arms of the same type, like:

> ```wit
> variant either-list {
>     left(list<u8>),
>     right(list<u8>),
> }
> ```

> In this case, both left and right would have the same `(ref eq)` subtype.

@fitzgen It occurred to me that the method compiled to `(ref eq)` still exists, that is, each variant is actually a new type.

If a language went the `(enum, ref null eq)` route, then the binding might look like (rust pseudocode):

```rust
enum Variants {
    Null,                 // (0, 0, 0)
    Bool { value: i32 },  // (1, i32, 0)
    U64 { value: i64 },   // (2, i64) , flatten i64 to i32 + i32
    I64 { value: i64 },   // (3, i64) , flatten i64 to i32 + i32
}

impl Variants {
    fn is_bool(&self) -> bool {
        match self {
            Variants::Bool { value } => true,
            _ => false,
        }
    }
}
```

If a language took the `(ref eq)` route, then the binding might look like (C# pseudocode):

```csharp
abstract class Variants
{
    class None : Variants
    {
        public override bool isBool()
        {
            return false;
        }
    }

    class Bool : Variants
    {
        int value;

        public override bool isBool()
        {
            return true;
        }
    }

    class U64 : Variants
    {
        long value;

        public override bool isBool()
        {
            return false;
        }
    }

    class I64 : Variants
    {
        long value;

        public override bool isBool()
        {
            return false;
        }
    }

    public abstract bool isBool();
}
```

These languages are more likely to implement variant determination through virtual tables

- X is Variants
- X is Variants::Bool
- X is Variants::Bool
- X is Variants::U64
- X is Variants::I64

design/mvp/Explainer.md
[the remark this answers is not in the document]
lukewagner: >  At worse it is equivalent to the other option, at best it avoids copies that the other option cannot.

I think the worst case is when both sides of a lift and lower have mutable GC types as their "stable state" representation (e.g., of a mutable language-level array) and thus there are 2 copies made: one by the lowering, one by the compiler glue code that copies the immutable type into a mutable type.

I'm not sure there's a great single answer here, so I've been thinking that we'd probably end up wanting both.

> My concerns about mutability mainly involve covariant arrays. 

With mutable types, the semantics of lift/lower would be to copy (just like with linear memory today), and thus I think there shouldn't be any covariant array issues (like one has in Java), unless I'm missing a different point you're getting at?

design/mvp/FutureFeatures.md
```diff
@@ -38,6 +38,23 @@ to and from the statically-compiled host implementation language). See
...
+version, which may lead to unexpected semantic damage. This can only be 
+guaranteed by the language compiler for the time being, and it is not 
+cross-language safe.
+
+## Integrate `stringref` with reference types 📡
+
+The current wit `string` type is regarded as `(array (u8))` or `(array (mut u8))`. 
+When the `stringref` proposal is stable, it can be considered to be mapped to 
+`stringref`. In other words, the `reference-type` option can work together with 
+the `string-encoding` option.
```
lukewagner, line 55: Currently, `stringref` appears to have been abandoned in favor of the [string built-ins proposal](https://github.com/WebAssembly/js-string-builtins/blob/main/proposals/js-string-builtins/Overview.md) (for more, see [this post](https://wingolog.org/archives/2023/10/19/requiem-for-a-stringref)).  Thus, I'd suggest removing this paragraph.

Alternatively, though, we could add a new lowering option that lowers `string` into an `externref` (or, in the farther future, [type import](https://github.com/WebAssembly/proposal-type-imports/)) and then we could provide a conservative list of string [canonical built-ins](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#canonical-built-ins) for operating on these string references (being careful to only provide built-ins for operations that were reliably O(1) for all host string implementations).

design/mvp/Explainer.md
```diff
@@ -1173,6 +1180,49 @@ The Canonical ABI will use `realloc` both to allocate (passing `0` for the
...
+📡 `(reference-type ...)` does not take effect by default. When this option 
+is turned on, Canonical ABI will be required to use reference types to pass 
+parameters. This option conflicts with `memory` and `realloc` and cannot 
+exist at the same time.
+
+📡 In the MVP version, the reference type refers to the specific monomorphic 
+type, such as `(ref array (mut u8))`, instead of type-erased 
+`(ref eq)`. 
+
+📡 When `reference-type` is enabled, the parameter type will change as follows:
```
lukewagner, line 1194: I think the right place to put this spec information is in CanonicalABI.md (along-side the current spec information for how non-gc works) and in `canonical-abi/definitions.py` (which has the benefit that you can write tests for it in `run_tests.py`)

design/mvp/Binary.md
```diff
@@ -276,6 +276,7 @@ canonopt ::= 0x00                                                => string-encod
            | 0x03 m:<core:memidx>                                => (memory m)
            | 0x04 f:<core:funcidx>                               => (realloc f)
            | 0x05 f:<core:funcidx>                               => (post-return f)
+           | 0x06 vec(<ref-options>)                             => (reference-type opt*) 📡
```
lukewagner, line 279: Is `ref-options` defined somewhere?  I'd be inclined just to name this new option `gc`, and we can always add new `canonopt`s in the future to refine how `gc` works.

lukewagner (commented):
Thanks for starting this!  I left a few comments and requests above.

design/mvp/FutureFeatures.md
oovm: How to map strings is beyond the scope of this proposal, so I will delete this paragraph.

design/mvp/Binary.md
oovm: If I want to add a refine option that only takes effect when the reference-type is enabled, is it possible to use a extra op code (e.g. `0x07`)?

lukewagner: Yep!  We can keep adding them going forward and specify the rules for which combinations of `canonopt`s are valid.  E.g., you can consider `0x03` (`memory`) and `0x04` (`realloc`) to both be two sub-options of the default "linear memory" option, with neither being valid in combination with the `gc` option.

oovm: Ok, I removed the ref-options content

oovm:
`reference-type` now focuses on the mapping of three heap types: `string`, `array`, and `struct`.

The reference types corresponding to `option`, `result`, `variants` have been moved to https://github.com/WebAssembly/component-model/issues/325

design/mvp/Explainer.md
oovm: I don't know how to rewrite `definitions.py`. The binary needs to reference an index of an already defined type.

design/mvp/Explainer.md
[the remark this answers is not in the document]
rossberg: FWIW, `ref null $t` does not generally work for options, because you can well have `option<option<T>>` – screwing this up is a sad mistake in many languages, let this not be one of them. :)

design/mvp/Explainer.md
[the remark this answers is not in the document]
oovm: @lukewagner Okay, I understand, arrays that cross boundaries need to be copied. So if mutability is required, the data provider should provide a resource type and corresponding setter function?

design/mvp/Explainer.md
lukewagner: Since `definitions.py` is not a complete Python reference implementation but, rather, just a "suggestive" subset that describes just the lifting/lowering/built-in rules, what we do is, for `canon` definitions that take a `typeidx` in the binary format, we have the corresponding `canon_*` Python function take a Python object that directly represents the type (e.g., see how [`canon_lift`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/canonical-abi/definitions.py#L1102) takes a [`FuncType`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/canonical-abi/definitions.py#L71-L84) directly).  That being said, I don't think there are any cases where you'll need to take a *core* wasm gc type as an immediate -- core wasm types are programmatically derived from *component-level* types by, e.g., [`flatten_functype`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/canonical-abi/definitions.py#L835) which is called by `canon_lift` and `canon_lower`.  Thus, I think you just need to add `CoreArrayType` and `CoreStructType` Python classes (analogous to `CoreFuncType`) so they can be created by `flatten_functype` in the appropriate cases when `cx.opts.gc` is true (which you'd also add to [`CanonicalOptions`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/canonical-abi/definitions.py#L298).

kevmoo:
Super excited about the possibilities here. What's the status?

oovm:
- The mapping of `struct <-> record`, `struct <-> tuple`, `list <-> array` has been agreed upon.
- For `enum` and `flags`, no mapping is required.
- For `variant` and `string`, there are objections and they have been removed from the proposal.

---

The unprepared part is mainly the implementation of the python validator (`definitions.py`), which has recently changed.

I have not been able to spend too much time on implementing this part recently due to my employer, so I am very sorry.

oovm:
Closed because the underlying runtime model has changed since 0.2 and needs to be reconsidered based on the 0.3 runtime model and semantics.

Issue #319, opened 2024-03-14
Question:  Specialized value types


GordonSmith:
Looking at:  https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md?plain=1#L638-L653 it shows that a `string` should be despecialized to a `list<char>`, but it is not implemented like that in definitions.py, is that a conflict?

sunfishcode:
The Canonical ABI uses custom representations for `string` and `flags`, rather than using the corresponding generic despecialized types.

I submitted #320 with some documentation which I hope helps here.

PR #320, opened 2024-03-14
Document what "specialized" means.

sunfishcode:
Add a paragraph documenting specialized types and their relationship to despecialized types.

This is intended to help with #319.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -651,7 +651,7 @@ defined by the following mapping:
 Specialized value types have the same semantics as their corresponding
 despecialized types, but have distinct type encodings. This allows them to
 convey a more specific intent; for example, `result` isn't just a variant,
-it's a variant that *means* succcess or failure, so source-code bindings
+it's a variant that *means* success or failure, so source-code bindings
 can expose it via idiomatic source-language error reporting. And this can
 also sometimes allow values to be represented differently; for example
 `string` may use different Unicode encodings, while still representing the
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -648,15 +648,18 @@ defined by the following mapping:
                                 string ↦ (list char)
 ```
 
-Specialized value types have the same semantics as their corresponding
-despecialized types, but have distinct type encodings. This allows them to
+Specialized value types have the same set of semantic values as their 
+corresponding despecialized types, but have distinct type constructors
+(which are not type-equal to the unspecialized type constructors) and
+thus have distinct binary encodings. This allows specialized value types to
 convey a more specific intent; for example, `result` isn't just a variant,
 it's a variant that *means* success or failure, so source-code bindings
-can expose it via idiomatic source-language error reporting. And, this can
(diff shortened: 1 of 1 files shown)


design/mvp/Explainer.md
```diff
@@ -647,6 +647,16 @@ defined by the following mapping:
 (result <valtype>? (error <valtype>)?) ↦ (variant (case "ok" <valtype>?) (case "error" <valtype>?))
                                 string ↦ (list char)
 ```
+
+Specialized value types have the same semantics as their corresponding
+despecialized types, but have distinct type encodings. This allows them to
+convey a more specific intent; for example, `result` isn't just a variant,
+it's a variant that *means* succcess or failure, so source-code bindings
```
lann, line 654:
```suggestion
it's a variant that *means* success or failure, so source-code bindings
```

design/mvp/Explainer.md
```diff
@@ -647,6 +647,16 @@ defined by the following mapping:
...
+can expose it via idiomatic source-language error reporting. And this can
```
lann, line 655: It might be nice to call out canonical ABI string encoding here as well.

lann (approved):

design/mvp/Explainer.md
sunfishcode: I've now added a sentence mentioning that the Canonical ABI has different representations for `string` and `flags`.

design/mvp/Explainer.md
```diff
@@ -647,6 +647,17 @@ defined by the following mapping:
...
+
+Specialized value types have the same semantics as their corresponding
+despecialized types, but have distinct type encodings. This allows them to
+convey a more specific intent; for example, `result` isn't just a variant,
+it's a variant that *means* success or failure, so source-code bindings
+can expose it via idiomatic source-language error reporting. And, this can
+also sometimes allow values to be represented differently; for example,
+`string` may use different Unicode encodings while still representing the
+same set of values. Notably, the Canonical ABI uses different representations
+for `string` and `flags`.
```
lukewagner, line 659:
```suggestion
Specialized value types have the same set of semantic values as their 
corresponding despecialized types, but have distinct type constructors
(which are not type-equal to the unspecialized type constructors) and
thus have distinct binary encodings. This allows specialized value types to
convey a more specific intent; for example, `result` isn't just a variant,
it's a variant that *means* success or failure, so source-code bindings
can expose it via idiomatic source-language error reporting. This can
sometimes allow values to be represented differently in the Canonical ABI.
For example, `string` uses various Unicode encodings while `list<char>`
uses a sequence of 4-byte `char` code points.  Similarly, `flags` uses a
bit-vector while an equivalent record of boolean fields uses a sequence of
boolean-valued bytes.
```

lukewagner (commented):
Thanks, good idea!  I made a few suggested tweaks:

sunfishcode:
Thanks! I made a few additinoal minor wording adjustments; in particular, I clarified that the property that values can have different representations isn't specific to the Canonical ABI, which will matter more when the component model grows new features involving encoded values.

Issue #277, opened 2023-11-29
Rename `float32`/`float64` back to `f32`/`f64`


esoterra:
As of preview2, the NaN-canonicalization rules have been relaxed and so in preview 2 the set of semantic values of `float32` is the same the set for core wasm `f32` (and likewise for `float64`/`f64`). 

Since this was the main motivator for distinguishing between core wasm `fXX` and component model `floatXX` types, we should consider gradually renaming the component model floating point types to match core wasm again.

sunfishcode:
I find it useful to think of the rules not as "the bits might be randomized" but instead as "component-model floating-point types only have a single NaN value", with the randomization just being an artifact of how they're encoded in core wasm.

sunfishcode:
Independently of what the types are named, is it desirable to allow multiple distinct NaN payloads to be observed when passing values out to non-wasm? It would seem that any code depending on that functionality would be unvirtualizable.

lukewagner:
The original motivation for the change was to avoid forcing an extra canonicalization pass when, e.g., passing a `list<f32>` out to a host library.  The allowed randomization means you can't (or at least *shouldn't*) depend on seeing any particular NaN payload, although you also can't depend on seeing exactly one or a particular one either.

sunfishcode:
With #279, the component-level `float32` and `float64` values are different from core-wasm `f32` and `f64, in that they only have a single NaN value. Implementations can omit canonicalization and randomization instructions when doing lifting and lowering together, but this is a permitted optimization rather than a property of the types.

I've also now posted https://github.com/WebAssembly/component-model/pull/288 to add more documentation.

lukewagner:
Yep, with those semantic changes, I think you're right.  Once we merge #288, I think we can close this issue, but I'll wait a while to collect more thoughts if anyone disagrees.

rossberg:
Hm, I understand the motivation, but as is I find the naming discrepancy more confusing than helpful, because there is nothing in the name which actually conveys _how_ `floatN` is different, and moreover, why the naming scheme deviates from that for ints. As is, it just looks like a seemingly random choice and all it does is create a mental impedance mismatch.

Feel free to ignore, but short of a more self-explanatory alternative, I think I'd prefer the short names, since the semantics still is close enough and the difference, where it matters at all, implied by context.

lukewagner:
[...]

PR #324, opened 2024-03-18
Rename `float32`/`float64` to `f32`/`f64`.

sunfishcode:
The `float32` and `float64` types are being
[renamed to `f32` and `f64`]. All the main tools have been updated to accept both old and new names for now, so there's no urgency to change anything, but users who wish to can now start switching to the new `f32`/`f64` names.

[renamed to `f32` and `f64`]: https://github.com/WebAssembly/component-model/issues/277

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -176,8 +176,8 @@ primvaltype   ::= 0x7f                                    => bool
                 | 0x79                                    => u32
                 | 0x78                                    => s64
                 | 0x77                                    => u64
-                | 0x76                                    => float32
-                | 0x75                                    => float64
+                | 0x76                                    => f32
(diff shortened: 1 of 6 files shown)


lukewagner (approved):
Thanks for doing this!  Could you also s/Float32/F32/ and s/Float64/F64/ in CanonicalABI.md and canonical-abi/definitions.py?

In terms of merging into this repo: if the tools are already updated to accept `f32`/`f64`, then it seems like we can merge (after a day or two for final comments); is that right?

sunfishcode:
`Float32`/`Float64` fixed. Yes, I believe everything is ready for this to be merged, pending the day or two for final comments.

lukewagner:
(Oops, I forgot to merge this earlier.)

PR #326, opened 2024-03-21
Fix a broken link to the old Subtyping.md.

sunfishcode:
Update the wording to reflect that subtyping is something that may be defined in the future, rather than being defined today.


lukewagner:
Thanks!

PR #336, opened 2024-04-10
formalize component value definitions

rvolosatovs:
This is not ready for full review yet - I will still write the documentation and keep updating the PR, however opening this draft PR to gather some early feedback on:
- EBNF definitions - I used https://github.com/WebAssembly/component-model/blob/145752a9795f326875c347527040d0e2ebe87c35/design/mvp/Binary.md?plain=1#L363-L367, https://github.com/WebAssembly/component-model/blob/145752a9795f326875c347527040d0e2ebe87c35/design/mvp/Binary.md?plain=1#L265 as main points of reference
- Encoding format itself. I tried to stay as close as possible to core Wasm spec and the canonincal ABI spec.
   - flags are encoded as variable-length `u32` bitmask - this does favor cases where flags with lowest indexes are set, however I feel like this is a reasonable approach to take at least for consistency with all other numerical values
   - enum and variant discriminants are encoded as variable-length `u32`-s, same argument applies as above. I do feel like for most use cases the discriminant should fit in a single byte
   - results and options match canonical ABI despecialization rules (https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#despecialization)
      - `0x00` is used to identify `result::ok`
      - `0x01` for `result:error`
      - `option::none` is `0x00`
      - `option::some` is `0x01`
   - tuples and records are encoded as just the sequences of their field/element values, in order
- `valtype` is specified once in the top-level `(value t:<valtype> v:<val(t)>)`, because, although type inference should be possible in most use cases by looking at the usages of the value index, values may potentially be unused by the component directly (for example, exported). I feel like it's a very little price to pay for simpler implementation and type checking.

`val` encoding reference implementation is available at https://github.com/rvolosatovs/wrpc/blob/4fd7488caa4cff19079f851cd7d0efda8f613a64/crates/transport/src/lib.rs#L2557-L3223


design/mvp/Binary.md
```diff
@@ -346,6 +347,57 @@ Notes:
...
+val(i:<typeidx>)                          ::= v:<defval(t)>                 => v (if (type t) = type-index-space[i])
+defval((record lt+))                      ::= (v:<fieldval(f:lt)>)^n        => v^n (if |lt*| > 0 and n = |lt*|)
+defval((variant case*))                   ::= i:<core:u32> v:caseval(c)     => i v (if i < |case*| and c = case[i])
+defval((list t))                          ::= vec(<val(t)>)
+defval((tuple t+))                        ::= (v:<val(e:t)>)^n              => v^n (if |t*| > 0 and n = |t*|)
+defval((flags l+))                        ::= v:<core:u32>                  => v (if |l*| > 0 and v < 2^|l*|)
+defval((enum l*))                         ::= i:<core:u32>                  => i v (if i < |l*|)
+defval((option t))                        ::= 0x00                          => none
+                                            | 0x01 v:<val(t)>               => (some v)
+defval((result t? (error u)?))            ::= 0x00                          => ok
```
rvolosatovs, line 379: not quite sure if the fact that `t` is not set here should be somehow expressed?

design/mvp/Binary.md
```diff
@@ -346,6 +347,57 @@ Notes:
...
+                                            | 0x00 v:<val(t)>               => (ok v)
+                                            | 0x01                          => error
```
rvolosatovs, line 381: Same as above, I feel like something is missing to denote that `(error u)` is not set for this case

design/mvp/Binary.md
```diff
@@ -346,6 +347,57 @@ Notes:
...
+val(u16)                                  ::= v:<core:u16>                  => v
+val(s32)                                  ::= v:<core:s32>                  => v
+val(u32)                                  ::= v:<core:u32>                  => v
+val(s64)                                  ::= v:<core:s64>                  => v
+val(u64)                                  ::= v:<core:u64>                  => v
+val(f32)                                  ::= v:<core:f32>                  => v
+val(f64)                                  ::= v:<core:f64>                  => v
+val(char)                                 ::= v:<core:u32>                  => v
+val(string)                               ::= v:<core:name>                 => v
+val(i:<typeidx>)                          ::= v:<defval(t)>                 => v (if (type t) = type-index-space[i])
```
lukewagner, line 370: The decode rules for `t:<valtype>` have already mapped a `<typeidx>` into the expanded type expression, so I think you can basically delete this `val(i:<typeidx>)` rule and rename the subsequence rules from `defval(...)` to `val(...)`.

design/mvp/Binary.md
```diff
@@ -346,6 +347,57 @@ Notes:
...
+defval((record lt+))                      ::= (v:<fieldval(f:lt)>)^n        => v^n (if |lt*| > 0 and n = |lt*|)
```
lukewagner, line 371: Hah, yeah, it's a rather peculiar notation that Guy Steele calls [Computer Science Metanotation](https://www.youtube.com/watch?v=7HKbjYqqPPQ).  I think you can write it like this (although I probably messed up the indentation).
```
defval((record (field l t)+))                      ::= v+:<val(t)>+        => { (l: v)+ }
```

design/mvp/Binary.md
```diff
@@ -346,6 +347,57 @@ Notes:
...
+defval((variant case*))                   ::= i:<core:u32> v:caseval(c)     => i v (if i < |case*| and c = case[i])
```
lukewagner, line 372:
```
defval((variant (case l t)+))               ::= i:<core:u32> v:<val(t[i])>     => l[i](v)
```

design/mvp/Binary.md
lukewagner: I'd have 4 separate rules:
```
val((result)) ::= ... => ...
val((result t)) ::= ... => ...
val((result (error e))) ::= ... => ...
val((result t (error e))) ::= ... => ...
```

design/mvp/Explainer.md
```diff
@@ -1633,7 +1635,12 @@ the standard [avoidance problem] that appears in module systems with abstract
 types. In particular, it ensures that a client of a component is able to
 externally define a type compatible with the exports of the component.
 
+### 🪙 Value Definitions
 
+Components may define values in the value index
+```ebnf
+value ::= t:<valtype> v:<val(t)> => (value t v)
+```
```
lukewagner, line 1643: I think this part is in-progress, but just fwiw, I think we also want to be able to optionally bind an identifier, so you'll want `(value <id>? <valtype> <val>)`.  Note also that while Binary.md uses "attribute grammar" syntax (mapping bytes to an AST), here we just want a grammar (no mapping).

So I had been thinking we'd connect this to default parameters, but I guess an even more minimalist first usage of Value Definitions is just defining constant values that are exported, which could actually be useful.  So perhaps show an example component that defines and exports a value.

Oh, and one other really-cool thing is if you could an an `eq` bound to value imports and exports (similar to what we already have in `type` imports and exports) where the right-hand side of the `eq` bound could be a `<validx>`, and that way constant values would become part of the component *type*, which I was actually seeing a use case for in CNCF land (it could be how a component declares what port it wants to bind to without needing a separate config file).

lukewagner (commented):
Great start, this is looking good.

design/mvp/Binary.md
rvolosatovs: Does this mean that we don't need to account for the fact that `t` may be missing? (since it's defined as `t?` in the rule https://github.com/WebAssembly/component-model/blob/444b0f49bffbfb1cb74224b4cecc09d3e8540d96/design/mvp/Binary.md?plain=1#L195)

design/mvp/Binary.md
rvolosatovs: I was not sure how abstract we wanted the WAT representation here, if we use the "struct" syntax, then shouldn't we also account for the commas(or some other delimiter), so maybe something like this:

```
defval((record (field l t)+))             ::= v+:<val(t)>+                  => { (l: v,)^(n-1) l[n-1]: v[n-1] } (if n = |l*|)
```
?

Also, it seems that the `+ implies a particular *order* of field definitions here, is that fine? I'm assuming that users would expect that field definition order would not matter, can we leave this loosely defined and let the parsers figure it out? If we do enforce the field definition order here, is there still a reason to require the label?

design/mvp/Binary.md
```diff
@@ -346,6 +347,59 @@ Notes:
...
+val(f32)                                  ::= v:<core:f32>                  => v
+val(f64)                                  ::= v:<core:f64>                  => v
+val(char)                                 ::= v:<core:u32>                  => v
+val(string)                               ::= v:<core:name>                 => v
+val(i:<typeidx>)                          ::= v:<defval(t)>                 => v (if (type t) = type-index-space[i])
+defval((record (field l t)+))             ::= v+:<val(t)>+                  => { (l: v,)^(n-1) l[n-1]: v[n-1] } (if n = |l*|)
+defval((variant (case l t?)+)             ::= i:<core:u32> v:val(t[i])      => l[i](v)
+defval((list t))                          ::= v:vec(<val(t)>)               => [v]  
+defval((tuple t+))                        ::= v+:<val(t)>+                  => (v+)
+defval((flags l+))                        ::= v:<core:u32>                  => ((l[i] (if and(v, 2^|i|) > 0))^(i:|l*|))
```
rvolosatovs, line 375: not quite sure how to define the WAT piece here

design/mvp/Binary.md
lukewagner: Ah hah, you're right!  I forgot the `valtype` binary decode rules leave the `typeidx` as an index.  In that case, I'd tweak my recommendation to instead be:
```
val(i:<typeidx>) ::= v:<val(type-index-space[i])> => v
val((record ...
val((variant ...
```
since `type-index-space[i]` is a `defvaltype`, which also includes the primitive types.

design/mvp/Binary.md
lukewagner: Oops, you're right, I forgot about the optionality of the case payload.  Ok, instead (and considering your above point about decoding into WAT literals), I'd suggest something like:
```
defval((variant (case l t?)+)) ::= i:<core:u32> v?:<val(t[i])>? => (variant l[i] v?)
```

design/mvp/Binary.md
lukewagner: I think we could say that the literal `(flags "a" "b" "c")` refers to the single flags value with the `a`, `b` and `c` fields set and the rest clear (recalling that, semantically, "flags" is a record with boolean fields for each defined flag label).

design/mvp/Binary.md
```diff
@@ -347,6 +350,61 @@ Notes:
...
+val(f64)                   ::= v:<core:f64>                    => v
+val(char)                  ::= v:<core:u32>                    => v
+val(string)                ::= v:<core:name>                   => v
+val(i:<typeidx>)           ::= v:<val(type-index-space[i])>    => v
+val((record (field l t)+)) ::= v+:<val(t)>+                    => (record v+)
+val((variant (case l t?)+) ::= i:<core:u32> v?:<val(t[i])>?    => (variant l[i] v?)
+val((list t))              ::= v:vec(<val(t)>)                 => (list v)
+val((tuple t+))            ::= v+:<val(t)>+                    => (tuple v+)
+val((flags l+))            ::= v:<core:uN>                     => (flags (l[i] for i in 0..N-1 if v & 2^i > 0)) (if N = |l*|)
+val((enum l+))             ::= i:<core:u32>                    => (enum l[i])
```
rvolosatovs, line 379: unlike records and flags, for example, `enum`s and `variant`s may be empty. These rules do not cover that case, should we add an explicit rule for `val((enum))` and `val((variant))`?
If we do have to do that, I'd prefer to use 0 bytes to encode them (i.e. do not emit any bytes), is there a way to express that?

lukewagner: Looking at Explainer.md just now, `enum` and `variant` must be non-empty.  (If they were empty, it would mean one of these strange bottom values like you have in Haskell... ;-)

design/mvp/Explainer.md
```diff
@@ -1341,22 +1345,110 @@ number of threads that can be expected to execute concurrently.
...
+        | (variant "<label>" <val>?)
+        | (list <val>*)
+        | (tuple <val>+)
+        | (flags "<label>"*)
+        | (enum "<label>")
+        | none | (some <val>)
+        | ok | (ok <val>) | error | (error <val>)
+```
+
+Where `val` definition must match the `valtype` specified as defined in [Binary.md](Binary.md).
```
lukewagner, line 1381: I'm not sure what you mean here; it seems like Explainer.md shouldn't need to refer to Binary.md (only the other way around).

design/mvp/Explainer.md
```diff
@@ -1341,22 +1345,110 @@ number of threads that can be expected to execute concurrently.
...
+            (record
+                (field "a" (option string))
+                (field "b" (tuple (option u8) string))
+            )
+            (list char)
+            $abc
+            string
+        )
+    )
+    (value $complex1 $complex
```
lukewagner, line 1428:
```suggestion
    (value $complex1 (type $complex)
```

design/mvp/Explainer.md
```diff
@@ -1341,22 +1345,110 @@ number of threads that can be expected to execute concurrently.
...
+            (record
+                none
+                (tuple none "empty")
+            )
+            (list)
+            (flags)
+            ""
+        )
+    )
+    (value $complex2 $complex
```
lukewagner, line 1439:
```suggestion
    (value $complex2 (type $complex)
```

design/mvp/Explainer.md
```diff
@@ -1366,6 +1458,28 @@ components. As an example value import:
 As this example suggests, value imports can serve as generalized [environment
 variables], allowing not just `string`, but the full range of `valtype`.
 
+Example of a value export:
```
lukewagner, line 1461:
```suggestion
Values can also be exported:
```

design/mvp/Explainer.md
```diff
@@ -1366,6 +1458,28 @@ components. As an example value import:
...
+```wasm
+(value $default_url string "https://example.com")
+(export "default-url" (value $default_url))
+;; `default-url` value export will infer `(eq $default_url)` value bound in the component type,
+;; include a type ascription to explicitly set type to `string`:
+(export "default-url-string" (value $default_url) (value string))
+```
```
lukewagner, line 1468: Could you end the triple-quote block and put the words in comments into a new paragraph, and then start a new triple-quote block with the alternative `export`?

design/mvp/Explainer.md
```diff
@@ -1927,6 +2041,17 @@ and will be added over the coming months to complete the MVP proposal:
...
 
+[`core:s8`]: https://webassembly.github.io/spec/core/syntax/values.html#integers
+[`core:u8`]: https://webassembly.github.io/spec/core/syntax/values.html#integers
+[`core:s16`]: https://webassembly.github.io/spec/core/syntax/values.html#integers
+[`core:u16`]: https://webassembly.github.io/spec/core/syntax/values.html#integers
+[`core:s32`]: https://webassembly.github.io/spec/core/syntax/values.html#integers
+[`core:u32`]: https://webassembly.github.io/spec/core/syntax/values.html#integers
+[`core:s64`]: https://webassembly.github.io/spec/core/syntax/values.html#integers
+[`core:u64`]: https://webassembly.github.io/spec/core/syntax/values.html#integers
+[`core:f32`]: https://webassembly.github.io/spec/core/syntax/values.html#floating-point
```
lukewagner, line 2052: Based on the above comment, I think all of these go away and you only keep `f64` and `i64` (note: `i64` contains both `u64` and `s64`).

lukewagner (approved):
Nice work!  This looks good to me, but I think @sunfishcode has also thought a lot about how to express component-model values in a similar context where we are not inferring the types from the value, but rather checking that they match a given target type, so I'd appreciate his review too.

design/mvp/Binary.md
rvolosatovs: sounds good to me, opened a PR to update that in `Binary.md` as well #346

design/mvp/Binary.md
```diff
@@ -36,6 +36,7 @@ section   ::=    section_0(<core:custom>)         => ϵ
             | s: section_9(<start>)               => [s]
             | i*:section_10(vec(<import>))        => i*
             | e*:section_11(vec(<export>))        => e*
+            | v*:section_12(vec(<value>))         => v*
```
rvolosatovs, line 39: @lukewagner should there also be a `🪙` here?

design/mvp/Explainer.md
```diff
@@ -564,10 +566,12 @@ externdesc    ::= (<sort> (type <u32>) )
                 | <functype>
                 | <componenttype>
                 | <instancetype>
-                | (value <valtype>) 🪙
+                | (value <valuebound>) 🪙
                 | (type <typebound>)
 typebound     ::= (eq <typeidx>)
                 | (sub resource)
+valuebound    ::= (eq <valueidx>)
+                | <valtype>
```
rvolosatovs, line 574: should these be gated by 🪙 as well?

design/mvp/Binary.md
```diff
@@ -215,12 +216,14 @@ importdecl    ::= in:<importname'> ed:<externdesc>        => (import in ed)
...
                 | 0x01 i:<typeidx>                        => (func (type i))
-                | 0x02 t:<valtype>                        => (value t) 🪙
+                | 0x02 b:<valuebound>                     => (value b) 🪙
                 | 0x03 b:<typebound>                      => (type b)
                 | 0x04 i:<typeidx>                        => (component (type i))
                 | 0x05 i:<typeidx>                        => (instance (type i))
 typebound     ::= 0x00 i:<typeidx>                        => (eq i)
                 | 0x01                                    => (sub resource)
+valuebound    ::= 0x00 i:<valueidx>                       => (eq i)
+                | 0x01 t:<valtype>                        => t
```
rvolosatovs, line 226: Do we need a 🪙 here?

design/mvp/Binary.md
```diff
@@ -347,6 +350,61 @@ Notes:
...
+val(s8)                    ::= v:<core:s8>                     => v
+val(s16)                   ::= v:<core:s16>                    => v
+val(u16)                   ::= v:<core:u16>                    => v
+val(s32)                   ::= v:<core:s32>                    => v
+val(u32)                   ::= v:<core:u32>                    => v
+val(s64)                   ::= v:<core:s64>                    => v
+val(u64)                   ::= v:<core:u64>                    => v
+val(f32)                   ::= v:<core:f32>                    => v
+val(f64)                   ::= v:<core:f64>                    => v
+val(char)                  ::= v:<core:u32>                    => v
```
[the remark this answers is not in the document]
rvolosatovs: Good point, according to `char` definition in https://webassembly.github.io/spec/core/syntax/values.html#names, would this change be sufficient to enforce that `v` represents a valid unicode character?
```suggestion
val(char)                  ::= v:<core:u32>                    => v (if v < 1114112)
```

design/mvp/Explainer.md
```diff
@@ -1341,22 +1345,110 @@ number of threads that can be expected to execute concurrently.
...
+        | <core:u16>
+        | <core:s32>
+        | <core:u32>
+        | <core:s64>
+        | <core:u64>
+        | <core:f32>
+        | <core:f64>
+        | '<core:char>'
+        | "<core:name>"
+        | (record <val>+)
```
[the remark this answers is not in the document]
rvolosatovs: My initial approach did not include `record`/`tuple` etc. keywords (but did surround in parentheses), however in conversation with @lukewagner I found out that (paraphrasing Luke) by convention `(operator <expr> ... <expr>)` s-expressions are used in place of pure `(<expr> ... <expr>)` in WAT - which is why I've added the type.

It seems to me that omission of parentheses here would introduce ambiguity in parsing e.g. `(record (field "a" (list u8)) (field "b" (list u8)))` values - without some kind of delimiter, how would one differentiate `{a:[1], b:[]}` from `{a:[], b:[1]}` ?

design/mvp/Binary.md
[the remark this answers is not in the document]
sunfishcode: We also need to exclude surrogates 0xd800 ≤ c < 0xe000.

design/mvp/Explainer.md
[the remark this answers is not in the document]
sunfishcode: Ah, I hadn't thought through nesting. I see now that `record`s would get tricky to read, and nested `list`s would create ambiguity. In light of that, In theory we could just omit the parens+keyword for `flags`, `variant`, and `enum`, but then it'd be awkwardly inconsistent with `record`/`list`. So your current proposed syntax here sounds good.

design/mvp/Binary.md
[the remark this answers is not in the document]
sunfishcode: It looks correct, though the `core:u16` made me pause for a moment because any time I see mentions of 16-bit values in a Unicode context I tend to assume there's something related to UTF-16 happening, which isn't the case here. I'd find a single `core:u32` rule with `(if v < 0xD800 or 0xE000 <= v <= 0x10FFFF)` easier to read, and it'd more closely correspond to what implementations are likely to do&mdash;decode the `u32` once, and then check the ranges.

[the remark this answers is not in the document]
rvolosatovs: I've already added a bunch of new syntax here and tried to avoid yet another addition - `or`. updated

design/mvp/Binary.md
```diff
@@ -347,6 +350,61 @@ Notes:
...
+val(u8)                    ::= v:<core:u8>                     => v
+val(s8)                    ::= v:<core:s8>                     => v
+val(s16)                   ::= v:<core:s16>                    => v
+val(u16)                   ::= v:<core:u16>                    => v
+val(s32)                   ::= v:<core:s32>                    => v
+val(u32)                   ::= v:<core:u32>                    => v
+val(s64)                   ::= v:<core:s64>                    => v
+val(u64)                   ::= v:<core:u64>                    => v
+val(f32)                   ::= v:<core:f32>                    => v
+val(f64)                   ::= v:<core:f64>                    => v
```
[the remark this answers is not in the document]
rvolosatovs: I've updated the proposal to handle NaNs and encode floats using leb128, just like all other numbers https://github.com/WebAssembly/component-model/pull/336/commits/db2250a7d8d2635a8e4ed356ed7ab7061aee7234 @lukewagner @sunfishcode what do you think about this approach?

design/mvp/Binary.md
lukewagner: Oh right yes; good catch.  In general any addition in this PR should be gated, so anywhere else too.

design/mvp/Explainer.md
lukewagner: Yep

design/mvp/Binary.md
lukewagner: Yep

design/mvp/Binary.md
```diff
@@ -347,6 +350,63 @@ Notes:
...
+val(s16)                   ::= v:<core:s16>                 => v
+val(u16)                   ::= v:<core:u16>                 => v
+val(s32)                   ::= v:<core:s32>                 => v
+val(u32)                   ::= v:<core:u32>                 => v
+val(s64)                   ::= v:<core:s64>                 => v
+val(u64)                   ::= v:<core:u64>                 => v
+val(f32)                   ::= v:<core:f32>                 => v (if v < 0x7F800000 or 0x7FFFFFFF < v <= 0xFF800000)
+                             | 0x7FC00000                   => NaN
+val(f64)                   ::= v:<core:f64>                 => v (if v < 0x7FF0000000000000 or 0x7FFFFFFFFFFFFFFF < v <= 0xFFF0000000000000)
+                             | 0x7FF8000000000000           => NaN
```
sunfishcode, line 372: The logic here looks good. I myself don't know the constraints of the spec language here, so I have two questions:
 - Can we specify multi-byte integers like 0x7FC00000 in the binary encoding, or do we need to split them into bytes?
 - If we have `v:<core:f32>`, can we do comparisons like `v < 0x7F800000` that operate on an integer view of the floating-point representation?

design/mvp/Binary.md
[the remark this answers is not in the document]
rvolosatovs: Adapted the type `eq` explanation to address this

sunfishcode (approved):

design/mvp/Binary.md
```diff
@@ -347,6 +350,64 @@ Notes:
...
+## 🪙 Value Definitions
+
+(See [Value Definitions](Explainer.md#value-definitions) in the explainer.)
+
+```ebnf
+value                      ::= t:<valtype> v:<val(t)>                  => (value t v)
+val(bool)                  ::= 0x00                                    => false
+                             | 0x01                                    => true
+val(u8)                    ::= v:<core:u8>                             => v
+val(s8)                    ::= v:<core:s8>                             => v
```
lukewagner, line 362: If I read the spec correctly for [`uN`](https://webassembly.github.io/spec/core/binary/values.html#integers), then the encoding of 128 as a `core:u8` will be the two-byte sequence 0x81 0x00, since even `u8` uses LEB128.  I wonder if it's worth special-casing `u8` and `s8` to instead use `core:byte` (so exactly 1 byte is decoded in all cases).

design/mvp/Binary.md
```diff
@@ -347,6 +350,64 @@ Notes:
...
+val(f64)                   ::= v:<core:f64>                            => v (if !isnan(v))
+                             | 0x00 0x00 0x00 0x00 0x00 0x00 0xF8 0x7F => NaN
+val(char)                  ::= v:<core:u32>                            => v (if v < 0xD800 or 0xE000 <= v <= 0x10FFFF)
+val(string)                ::= v:<core:name>                           => v
+val(i:<typeidx>)           ::= v:<val(type-index-space[i])>            => v
+val((record (field l t)+)) ::= v+:<val(t)>+                            => (record v+)
+val((variant (case l t?)+) ::= i:<core:u32> v?:<val(t[i])>?            => (variant l[i] v?)
+val((list t))              ::= v:vec(<val(t)>)                         => (list v)
+val((tuple t+))            ::= v+:<val(t)>+                            => (tuple v+)
+val((flags l+))            ::= v:<core:uN>                             => (flags (l[i] for i in 0..N-1 if v & 2^i > 0)) (if N = |l*|)
```
lukewagner, line 380:
```suggestion
val((flags l+))            ::= v:<core:uN>                             => (flags (l[i] for i in 0..N-1 if v & 2^i > 0)) (where N = |l+|)
```

design/mvp/Explainer.md
```diff
@@ -1341,22 +1345,102 @@ number of threads that can be expected to execute concurrently.
...
-Like modules, components can have start functions that are called during
-instantiation. Unlike modules, components can call start functions at multiple
-points during instantiation with each such call having parameters and results.
-Thus, `start` definitions in components look like function calls:
 ```ebnf
-start ::= (start <funcidx> (value <valueidx>)* (result (value <id>?))*)
+value ::= (value <id>? <valtype> <val>)
+val   ::= false | true
+        | <core:i64>
+        | <core:f64> | NaN
```
lukewagner, line 1360: If I'm reading the [text format rules for `f64`](https://webassembly.github.io/spec/core/text/values.html#text-float) correctly, `nan` is already included.  So is `nan:0x____`, for specifying non-canonical NaNs, so we need to disallow that somehow.  Maybe use `<f64canon>` here, defined below as:
```
f64canon ::= <core:f64> without the `nan:0x` case.
```

design/mvp/Explainer.md
```diff
@@ -1341,22 +1345,102 @@ number of threads that can be expected to execute concurrently.
...
+        | (tuple <val>+)
+        | (flags "<label>"*)
+        | (enum "<label>")
+        | none | (some <val>)
+        | ok | (ok <val>) | error | (error <val>)
+```
+
+Where `val` definition must match the `valtype` specified.
+
+For example:
```
lukewagner, line 1375:
```suggestion
The validation rules for `value` require the `val` to match the `valtype`.  For example:
```

design/mvp/Explainer.md
```diff
@@ -1366,6 +1450,37 @@ components. As an example value import:
 As this example suggests, value imports can serve as generalized [environment
 variables], allowing not just `string`, but the full range of `valtype`.
 
+Values can also be exported:
+```wasm
+(value $default_url string "https://example.com")
+(export "default-url" (value $default_url))
```
lukewagner, line 1456:
```suggestion
(component
  (value $default_url string "https://example.com")
  (export "default-url" (value $default_url))
)
```

design/mvp/Explainer.md
rvolosatovs: updated, also updated `Binary.md` to match `nan` casing https://github.com/WebAssembly/component-model/pull/336/commits/302f90a7fc1e670f72b28ea665ae57eb127154ee

design/mvp/Explainer.md
```diff
@@ -1366,6 +1450,37 @@ components. As an example value import:
...
+
+The `eq` bound says that the imported/exported value must be structurally equal
+to some preceding value definition. This allows, for example, an imported abstract value to be re-exported.
+
+Include a type ascription to explicitly set type to `string`:
+
+```
+(value $default_url string "https://example.com")
+(export "default-url" (value $default_url) (value string))
+```
```
````suggestion
Values can also be exported.  For example:
```wasm
(component
  (import "system-port" (value $port u16))
  (value $url string "https://example.com")
  (export "default-url" (value $url))
  (export "default-port" (value $port))
)
```
The inferred type of this component is:
```wasm
(component
  (import "system-port" (value $port u16))
  (value $url string "https://example.com")
  (export "default-url" (value (eq $url)))
  (export "default-port" (value (eq $port)))
)
```
Thus, by default, the precise constant or import being exported is propagated into the component's type and thus its public interface.  In this way, value exports can act as semantic configuration data provided by the component to the host or other client tooling.

Components can also keep the exact value being exported abstract (so that the precise value is not part of the type and public interface) using the "type ascription" feature mentioned [below](#imports-and-exports).
````

Comment (lukewagner) [reply design/mvp/Explainer.md]: And then could we add an example below (where the type ascription feature is introduced) of a value type ascription?

File: design/mvp/Explainer.md (hunk 24 above)
Comment (lukewagner) [reply design/mvp/Explainer.md]: Thanks!  As an ultra-nit, could you fix the indentation so that the `::=` of `f64canon` lines up?

File: design/mvp/Binary.md [hunk 28]
```diff
@@ -347,6 +350,64 @@ Notes:
...
+## 🪙 Value Definitions
+
+(See [Value Definitions](Explainer.md#value-definitions) in the explainer.)
+
+```ebnf
+value                      ::= t:<valtype> v:<val(t)>                  => (value t v)
+val(bool)                  ::= 0x00                                    => false
+                             | 0x01                                    => true
+val(u8)                    ::= v:<core:byte>                           => v
+val(s8)                    ::= v:<core:byte>                           => v
```
lukewagner, line 362:
```suggestion
val(s8)                    ::= v:<core:byte>                           => v if v < 128 else (v - 256)  
```

design/mvp/Explainer.md
rvolosatovs: Move the example to the import and export section and other than fixing the section reference and making it consistent with previous occurence applied the diff verbatim

lukewagner (approved):
Beautiful, thanks!

lukewagner:
This looks good to me, so I'll wait another couple of days to let folks comment and if nothing new comes up, merge.

design/mvp/Binary.md
```diff
@@ -358,7 +358,7 @@ Notes:
 (See [Value Definitions](Explainer.md#value-definitions) in the explainer.)
 
 ```ebnf
-value                      ::= t:<valtype> v:<val(t)>                  => (value t v)
+value                      ::= t:<valtype> len:<uN> v:<val(t)>         => (value t v) (where len = |v+|)
```
lukewagner, line 361:
```suggestion
value                      ::= t:<valtype> len:<uN> v:<val(t)>         => (value t v) (where len = ||v||)
```
Here I'm copying what the core spec does in the [decoding rule for function bodies](https://webassembly.github.io/spec/core/binary/conventions.html#auxiliary-notation), where ||x|| is defined in the conventions section [here](https://webassembly.github.io/spec/core/binary/conventions.html#auxiliary-notation) to be the length of the byte sequence decoded into `x`.  Could you also add a bullet below noting the meaning of `||v||` and linking to https://webassembly.github.io/spec/core/binary/conventions.html#auxiliary-notation?

lukewagner:
In case anyone isn't also following #352, the recent change above prefixes each value definition with a byte length (analogous to function bodies), so that values can be skipped-over during decoding or decoded in parallel, while still eagerly populating the value index space.

alexcrichton:
One part of this that I would be concerned about this is that this non-trivially increases the complexity of the binary-to-text tooling. Today binary-to-text for core modules and components doesn't require any validation to happen. With this new section, however, it's not possible to translate binary-to-text without a validator because you'll need to keep track of the component type section to print the value section. A kernel of this concern can be seen in https://github.com/bytecodealliance/wasm-tools/pull/1541 where the decoding step has to take type information which isn't easily acquired right now.

Unfortunately I don't know of a great way to simplify this which doesn't involve lots of discriminant bytes everywhere which seems like this was trying to avoid, however.

lukewagner:
@alexcrichton Yes, that's a real downside for sure.  What's hard to weigh is whether avoiding that problem is worth adding all the discriminants to the encoding of values, which ultimately would mean that we'll need a separate binary encoding for situations where we don't want the size/decode overhead.  It's hard to say for sure whether anyone will want to store very large values in a component, but, farther out in the future, maybe `list<record { ... }>`-typed value exports end up being nicer than raw text files for representing metadata/config, so... maybe?  But yeah, I feel the tension.

alexcrichton:
One possible alternative, albeit not a pretty one, would be for the text format to reflect the binary format where it'd be something like:

```wasm
(value $foo (type $the_type_index) "\0\0\01\02")
```

where the text format just has the raw bytes and doesn't try to pretty-print the shape

lukewagner:
@alexcrichton Oh right, I forgot that option!  So then the byte-string (using the same string-literal format as data segments) could be considered the "base" AST whereas the `(list 1 2 3)` syntax is optional sugar.

alexcrichton:
Ah yeah I like the idea of having sugar as well! I had forgotten that was an option, and that should make it much easier to hand-write as well.

Orthogonally, though, wouldn't `(list 1 2 3)` be invalid and you'd need something like `(list (i32 1) (i32 2) (i32 3))` or similar? (basically won't the text format need to know how big the integer is supposed to be?)

lukewagner:
Good point!  Because yeah, whether we have discriminants in the text format is independent of whether we have discriminants in the binary format.  I was thinking earlier only about the decoding direction, where you need type info to render values anyhow; but in the encoding direction, having the discriminants lets you parse without type info, addressing the original concern in #352.  It also lets you write invalid test cases more-easily.

So at the moment we only require discriminants where required by the more-fundamental WAT convention that all S-Expression lists start with an atom, hence the discriminant in `(list ...)` and `(record ...)`, but based on this reasoning, I think it makes sense to add them for all the scalars too, hence `(list (i32 1) (i32 2) (i32 3))`, like you're saying.  Does that make sense to you too @rvolosatovs?

design/mvp/Binary.md
```diff
@@ -350,6 +353,65 @@ Notes:
...
+## 🪙 Value Definitions
+
+(See [Value Definitions](Explainer.md#value-definitions) in the explainer.)
+
+```ebnf
+value                      ::= t:<valtype> len:<uN> v:<val(t)>         => (value t v) (where len = ||v||)
+val(bool)                  ::= 0x00                                    => false
+                             | 0x01                                    => true
+val(u8)                    ::= v:<core:byte>                           => (u8 v)
+val(s8)                    ::= v:<core:byte>                           => (s8 v) if v < 128 else (v - 256)
```
lukewagner, line 365:
```suggestion
val(s8)                    ::= v:<core:byte>                           => (s8 v') (where v' = v if v < 128 else (v - 256))
```

lukewagner (approved):
Thanks @rvolosatovs!

design/mvp/Binary.md
```diff
@@ -350,6 +353,65 @@ Notes:
...
+val(u16)                   ::= v:<core:u16>                            => (u16 v)
+val(s32)                   ::= v:<core:s32>                            => (s32 v)
+val(u32)                   ::= v:<core:u32>                            => (u32 v)
+val(s64)                   ::= v:<core:s64>                            => (s64 v)
+val(u64)                   ::= v:<core:u64>                            => (u64 v)
+val(f32)                   ::= v:<core:f32>                            => (f32 v) (if !isnan(v))
+                             | 0x00 0x00 0xC0 0x7F                     => (f32 nan)
+val(f64)                   ::= v:<core:f64>                            => (f64 v) (if !isnan(v))
+                             | 0x00 0x00 0x00 0x00 0x00 0x00 0xF8 0x7F => (f64 nan)
+val(char)                  ::= v:<core:u32>                            => v (if v < 0xD800 or 0xE000 <= v <= 0x10FFFF)
```
alexcrichton, line 376: I think this might need to be `(char v)` in the text format?

alexcrichton: (although text-wise it's probably nicer to have `(char "x")` rather than `(char 0x61)`

design/mvp/Binary.md
```diff
@@ -350,6 +353,65 @@ Notes:
...
+val(string)                ::= v:<core:name>                           => v
+val(i:<typeidx>)           ::= v:<val(type-index-space[i])>            => v
```
alexcrichton, line 378: Could the text format here perhaps be `(val $foo)` or `(val 10)` to better disambiguate integers?

design/mvp/Binary.md
```diff
@@ -350,6 +353,65 @@ Notes:
...
+val((record (field l t)+)) ::= v+:<val(t)>+                            => (record v+)
+val((variant (case l t?)+) ::= i:<core:u32> v?:<val(t[i])>?            => (variant l[i] v?)
+val((list t))              ::= v:vec(<val(t)>)                         => (list v)
+val((tuple t+))            ::= v+:<val(t)>+                            => (tuple v+)
+val((flags l+))            ::= v:<core:uN>                             => (flags (l[i] for i in 0..N-1 if v & 2^i > 0)) (where N = |l+|)
```
alexcrichton, line 383: Could this be fleshed out a bit more? For example is the intention to support a uleb47 for example with 47 flags? That will make decoding relatively complicated relative to what decoders already implement today. Personally I'd recommend going with one of two options:

* The binary encoding is u8, u16, or u32*N to match the in-memory canonical ABI
* The binary encoding is u32*N to match the in-flat-values canonical ABI

Additionally I think that the `(flags ...)` text format will need an integer perhaps for the total size of the flags? Otherwise it won't know what the encoding length is.

design/mvp/Explainer.md
```diff
@@ -1355,22 +1360,101 @@ number of threads that can be expected to execute concurrently.
...
+           | <core:name>
+           | (record <val>+)
+           | (variant "<label>" <val>?)
+           | (list <val>*)
+           | (tuple <val>+)
+           | (flags "<label>"*)
+           | (enum "<label>")
+           | none | (some <val>)
+           | ok | (ok <val>) | error | (error <val>)
+f64canon ::= <core:f64> without the `nan:0x` case.
```
alexcrichton, line 1386: I think this may want to be sync'd with the `Binary.md` section above? For example integers have discriminants and flags/variants/enums don't have strings for payloads but instead integers.

design/mvp/Binary.md
lukewagner: Good point calling out `flags`.  But thinking through what it would take to encode the text format into binary without type info, it seems like we'd need the full `flags` type (to know which flag goes into which bit), and similarly for variants (to know which discriminant value to assign).  So I wonder if the whole goal of being able to encode text to binary without type context is fraught.

design/mvp/Binary.md
rvolosatovs: The intention is actually to use `'x'` syntax (see [`Explainer.md`](https://github.com/WebAssembly/component-model/pull/336/files#diff-1661eacd58a4502758481f41023421baa5eb5c681e46d0ae3de215591f85e71eR1403)), I agree that current rule definition seems to be insufficient to express that, but I'm not sure how to correctly express that whether that actually belongs in the rule or simply as a note below. @lukewagner what do you think?

design/mvp/Explainer.md
rvolosatovs: Will update the numeric types, however the intention for flags, variants and enums is indeed to use labels and not integers

design/mvp/Binary.md
rvolosatovs: > Could this be fleshed out a bit more? For example is the intention to support a uleb47 for example with 47 flags? That will make decoding relatively complicated relative to what decoders already implement today. Personally I'd recommend going with one of two options:
> 
>     * The binary encoding is u8, u16, or u32*N to match the in-memory canonical ABI
> 
>     * The binary encoding is u32*N to match the in-flat-values canonical ABI

The intention for flags is indeed to be representable as an unsized unsigned LEB128, there are quite a few benefits to this approach:
- the decoder does *not*, in fact, *need* to know the flag variant count - any flag value *can* be decoded without the type context available - the decoder most probably *should* use the type info to bound the decode process, especially if the payload is not ultimately trusted
- given the above, adding and removing flag type variants to types can be both forwards and backwards compatible process across versions, which is generally a very useful property

To me it seems that making the flag encoding dependent on the variant count would make the binary format more complex. 


> Additionally I think that the `(flags ...)` text format will need an integer perhaps for the total size of the flags? Otherwise it won't know what the encoding length is.

The only way that a flag value definition could be encoded without type context available externally would be either fully duplicating the flag type definition or using the binary form, i.e. `(flags 0b00101)` for setting flag `C` and `E` in `flags { A, B, C, D, E }`

How about we simply introduce an alternative binary value representation?
"smart", user-facing, decoders could print:
```wasm
(flags "C" "E")
```

while this form would also be possible:
```wasm
(flags 0b00101)
```

Similarly, for variants, for example:
```wasm
(variant 0xDE "mypayload")
(variant 0x02)
(variant 0x80 (record (some (u8 42)) (list 't' 'e' 's' 't') (flag 0b00010101) ok ok error (error "with-string-payload") (ok (u16 123))))
```

Such format can be fully encoded without external type context.

I do think that the "label" format should be supported as a sugar for this "raw" representation, but not all decoders/encoders necessarily have to support it

design/mvp/Explainer.md
rvolosatovs: See also https://github.com/WebAssembly/component-model/pull/336#discussion_r1599858948 for potentially an alternative numeric representation with label syntax being the sugar for it

design/mvp/Binary.md
alexcrichton: To some degree this is running up against limitations of the text format I think because `'x'` isn't a valid or defined token in the text format in the same way Rust defines `'x'` and `"x"`. That could be added here, however, but as-is I don't believe it can be assumed to exist already.

design/mvp/Binary.md
alexcrichton: Representation-wise what I don't understand is that this rule uses `core:uN` where `N` is the size of the flags set. If the intention is to use a `u47` for example that has strict requirements about how large the representation can be given an overlong representation. Basically I'm asking for more precise details about how this is supposed to work because while I think it makes sense to use variably-sized integers to pack the bits here I still think there's some detail to flesh out.

For the text format the problem with `(flags "C" "E")`, which while I agree that's the best desired form, is that you need type/validation context when encoding text-to-binary. Currently the text-to-binary transform carries only a very small amount of context and it might not be trivial to implement carrying the type information through to be able to understand how to turn `"C"` into a flag.

lukewagner: Agreed, but I think the only way to get `(flags ... "C" "E")` to encode into binary without having to separately refer to the type index space is to basically inline all the type info of the `flags` type into that `...`, which sortof defeats the purpose of not needing types.

I guess there's always the spartan option of just nixing the whole optional expression form for values and *only* supporting the string literal with the encoded binary format (just like `data` segmenst).  Ultimately WIT will do its own literal syntax (probably embedding WAVE) and mostly noone will read that WAT, so this seems fine.

WDYallT?

alexcrichton: Personally I think that's ok yeah, given the relative rarity of working with WAT I don't think it's worth putting too too much worry into it.

Again though I still would like some more clarification on `uN` here. I don't know what that means for 47 flags, for example. Is the intention that a literal 47-bit unsigned LEB is used? Is the intention that a 128-bit LEB is always used which prevents >128 flags? Is an over-long encoding of a 47-bit leb with 10 bytes allowed? (example questions)

lukewagner: If I'm interpreting the existing definition of `uN` correctly in the [core spec](https://webassembly.github.io/spec/core/binary/values.html#integers), to decode a `u47`, you decode at most 7 bytes (`= ceil(47 / 7)`, with each byte containing the next-least-significant 7 bits of the `u47` and with the high bit indicating whether to decode another byte, and rejecting any bits set in the final byte that go beyond the 47th bit.  So to answer your qustions:
* A 128-bit LEB is not always used, and so there is no theoretical upper bound (although an impl could set a reasonably-high one, like 1000).  Thus, the "decode" primitive for decoding a `uN` might be given the signature `fn decode(n: u32) -> Vec<u8>`.
* The current spec gives you at most `ceil(n / 7)` bytes (i.e., the smallest number of bytes needed to encode the biggest value), so 10 bytes would not be allowed (nor 9 nor 8).

alexcrichton: If the intention is to use a literal `u47` then I'd loop back to what I said originally about how that's going to be relatively complicated given what existing decoders already have. Decoders already have u32/u64 decoders but the routines are generally heavily optimized and specialized to a bit width and are not amenable to "pass in 47 instead". 

I suppose I can rephrase what I'm saying here in that I would personally push back against the usage of u47. I'd prefer something like u32*N where N is calculated depending on the number of flags. There'd be validation for the final u32 read that bits aren't set, but that would make it much easier to reuse existing decoder methods.

rvolosatovs: FWIW, in Go, flags containing up to 64 variants can be decoded/encoded using just the standard library https://pkg.go.dev/encoding/binary#PutUvarint (this just uses LEB128)
Similarly, many crates exist in Rust for decoding at least the `u64`  https://docs.rs/leb128/latest/leb128/read/index.html LEB128

So the issue only arises when a flag variant 65 or higher is used (in other words bit 65 or higher is set)

Given the abundancy of prior art, it should not be too difficult to extend the LEB128 encode/decode algorithms to work on e.g. a https://docs.rs/num/latest/num/struct.BigUint.html to be able to represent flags utilizing more than 64 variants.

I do think that flags with more than 64 variants should be quite rare and so the overwhelming majority of the flags used by real-world interfaces should be trivial to implement

rvolosatovs: Just to make sure we're all aligned on this, the intention is for the encoding to be byte-aligned, so even if, for example, 13 bits are required to encode a flag value, 16 bits will be used (with the spare bits zeroed out) - in case that's the source of the concern

rvolosatovs: Here's an example (naive) Go bindgen flag decoder https://github.com/wrpc/wrpc/blob/d3d722035cd7387ebc6fcb94b5cefdf1926c34fe/crates/wit-bindgen-go/src/interface.rs#L707-L738
Where the "discriminant read" is just the LEB128 decoder https://github.com/wrpc/wrpc/blob/d3d722035cd7387ebc6fcb94b5cefdf1926c34fe/crates/wit-bindgen-go/src/interface.rs#L2401-L2438

This implementation is limited to flags containing up to 64 variants, as you can see, but like I said above it's just a matter of extending the algorithm to maintain the state in a "bit uint" of some sorts as opposed to a builtin sized uint type. decoding of "small flags" can be optimized

alexcrichton: At a high level I fear I'm being overly pedantic here, but concretely I do at the same time pretty strongly believe that a change in the spec is needed here. Your examples and code are actually all incorrect with respect to the spec as-is, and my point is that I'd like to make your examples and code correct with respect to the spec.

The weird parts and edge cases of LEB encodings mostly with overlong encodings. Let's consider a flag with 9 variants. As-is the spec as-written right now will read that as a `u9`. The LEB encoding of a `u9` mandates that at most two bytes can be used. It's not valid to do `0x80 0x80 0x80 0x80 0x80 0x1` as an encoding of 1. The maximal value of a `u9`, `0x1f`, is encoded as two bytes so more than two bytes cannot be used for the encoding here.

You're using `u64`-based decoders, however. A 64-bit value can take up to 10 bytes to be encoded, meaning that my above example wich should be invalid for `u9` is actually valid for a `u64`. This means that your decoding process can incorrectly classify an invalid binary as actually valid due to the acceptance of overlong encodings.

My suggestion is to not use `uN`, but instead use integral amounts of u32 or u64 to encode the flags (picking one, probably u64, I don't think it matters too too much). That means that your implementations are in fact correct, the spec says to decode a u64 and then validate the bits for the flags, which is what your implementations are doing.

lukewagner: Ah, I see your point; this does indeed seem (evidently) like a likely source of accidental incompatibility.  If we chunk in terms of `u64`, then approximately 99.9% of use cases will still have the same number of bytes as if we did `uN`, but now everyone gets to reuse their existing decode-a-`u64` path.

So I think we're decoding `(u:<u64>)^m` where `m = ceil(n / 64)` and `(u_1 | ... | u_m) < 2**n`, yes?

Do you think you could make that change @rvolosatovs?

alexcrichton: Yeah sounds reasonable to me! I'm happy with a sequence of uleb64-values

rossberg: Drive-by comment: u64^N seems like a hack when we already have LEB. Not sure I see the problem. A generic uN decoder with max length check is a trivial loop that is still simpler than a logic for splitting up into multiple u64s (e.g. it's 5 lines of code in the Wasm reference interpreter). I doubt it has more potential for error — @lukewagner's side condition above is likewise incorrect ;). Implementations of uN missing the length check can easily be prevented with a couple of negative tests in the test suite.

rvolosatovs: > At a high level I fear I'm being overly pedantic here, but concretely I do at the same time pretty strongly believe that a change in the spec is needed here. Your examples and code are actually all incorrect with respect to the spec as-is, and my point is that I'd like to make your examples and code correct with respect to the spec.
> 
> The weird parts and edge cases of LEB encodings mostly with overlong encodings. Let's consider a flag with 9 variants. As-is the spec as-written right now will read that as a `u9`. The LEB encoding of a `u9` mandates that at most two bytes can be used. It's not valid to do `0x80 0x80 0x80 0x80 0x80 0x1` as an encoding of 1. The maximal value of a `u9`, `0x1f`, is encoded as two bytes so more than two bytes cannot be used for the encoding here.
> 
> You're using `u64`-based decoders, however. A 64-bit value can take up to 10 bytes to be encoded, meaning that my above example wich should be invalid for `u9` is actually valid for a `u64`. This means that your decoding process can incorrectly classify an invalid binary as actually valid due to the acceptance of overlong encodings.
> 
> My suggestion is to not use `uN`, but instead use integral amounts of u32 or u64 to encode the flags (picking one, probably u64, I don't think it matters too too much). That means that your implementations are in fact correct, the spec says to decode a u64 and then validate the bits for the flags, which is what your implementations are doing.

I'm confused about this comment.

(assuming that the example values are given as little endian)

Decoding of `u9` from `0x80 0x80 ....` would stop after reading the first two bytes and fail with an overflow, since the encoded value could be at most 2 bytes long and the encoding had the most significant bit set in both bytes read.

Decoding of `u64` would, indeed, succeed and would produce: `0b_1_0000000_0000000_0000000_0000000_0000000` in BE

Here's an example decoder accounting for the bit size in Rust with test cases for the examples and a few more: https://github.com/wrpc/wasm-tokio/commit/62c89fa114088611df459794ec77f5e8d3834399

I have validated this using the OCaml decoder in https://github.com/WebAssembly/spec/blob/228549757b301c698c5588ec0d58b201b5777c92/interpreter/binary/decode.ml#L83-L88

```
─( 04:25:28 )─< command 0 >─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Wasm.Decode.uN 2 (Wasm.Decode.stream "test" (String.of_seq (List.to_seq (List.map Char.chr [0x01]))));;
- : int64 = 1L
─( 04:25:28 )─< command 1 >─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Wasm.Decode.uN 2 (Wasm.Decode.stream "test" (String.of_seq (List.to_seq (List.map Char.chr [0x02]))));;
- : int64 = 2L
─( 04:25:34 )─< command 2 >─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Wasm.Decode.uN 2 (Wasm.Decode.stream "test" (String.of_seq (List.to_seq (List.map Char.chr [0x03]))));;
- : int64 = 3L
─( 04:25:37 )─< command 3 >─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Wasm.Decode.uN 2 (Wasm.Decode.stream "test" (String.of_seq (List.to_seq (List.map Char.chr [0x04]))));;
Exception: Error (_, "integer too large").
─( 04:25:40 )─< command 4 >─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Wasm.Decode.uN 64 (Wasm.Decode.stream "test" (String.of_seq (List.to_seq (List.map Char.chr [0x80; 0x80; 0x80; 0x80; 0x80; 0x01]))));;
- : int64 = 34359738368L
─( 04:25:44 )─< command 5 >─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Wasm.Decode.uN 9 (Wasm.Decode.stream "test" (String.of_seq (List.to_seq (List.map Char.chr [0x80; 0x80; 0x80; 0x80; 0x80; 0x01]))));;
Exception: Error (_, "integer representation too long").
```

(it's my first time using OCaml, don't know if there's a more reasonable way to handle this, but I simply removed `interpreter/binary/decode.mli` file to ensure all functions are public and ran this in `dune utop`)

@alexcrichton does this address your concerns?

lukewagner: @primoly That's a good point: LEB128 is going to be a pretty mediocre compression format for bit-vectors, and so perhaps for simplicity we just go with the statically-sized (based on the type) list of bytes that get OR'd together.

sunfishcode: Additionally, the other reason the Wasm binary format uses LEB128 in many places is extensibility. Whenever we need some field to be bigger, we can just extend the range. But here, we don't need that kind of extensibility, because we're not trying to support subtyping at this level.

rvolosatovs: Happy to make the switch, as I mentioned in the PR description, LEB128 was really just used here for consistency with the rest of the spec, treating flags as numeric types

Are we ok with LEB128 for enum and variant discriminants, however, or should I change these to just LE unsigned integers?

lukewagner: It's hard to have a strong opinion regarding discriminants given that tangible differences only appear when you have >= 128 cases, which will be quite rare.  If we were doing LEB128-everywhere, that seemed like the consistent choice, but now that we're using fixed-width integers in a few cases, maybe it's better just to use the fixed-width [`discriminant_type`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#alignment) for discriminants too?

rossberg: > LEB128 is going to be a pretty mediocre compression format for bit-vectors

That depends on the kind of bit vector. Often flags are biased towards false as the common case (or can be designed such), in which case LEB still is beneficial on average. In cases where it isn't, it at least isn't much worse.

lukewagner: I expect the differences between these variations we're considering won't matter much in practice, given that most flags will fit in a byte or two.  Also, like core wasm, we're not trying to eek out every bit, just design something "reasonably compact" (on top of which generic compression could be applied if you really want).  Thus, for flags, I think we can just optimize for what's least likely to have subtle bugs in implementations, which suggests the static tuple of bytes.

rvolosatovs: Switched to using plain fixed-size little-endian uints for discriminants https://github.com/WebAssembly/component-model/pull/336/commits/3cd3dc844a5e96fc769bbf13a6b76d8af8b6c824

This encoding does allow for a 3-byte encoding of a 24-bit integer depending on the number of cases, which may not be too common, but should be fairly trivial to encode/decode

rvolosatovs:
For reference, in its' current shape binary encoding proposal is implemented at https://github.com/rvolosatovs/wrpc/blob/84960bb739fd2a26c1d534c67954bbace0f2f005/crates/runtime-wasmtime/src/lib.rs#L482-L1090 in terms of Wasmtime types and values.

rvolosatovs:
Updated the PR according to latest feedback gathered on and off-line. PTAL

alexcrichton (approved):

design/mvp/Binary.md
```diff
@@ -350,6 +353,66 @@ Notes:
...
+val(f64)                   ::= v:<core:f64>                            => v (if !isnan(v))
+                             | 0x00 0x00 0x00 0x00 0x00 0x00 0xF8 0x7F => nan
+val(char)                  ::= b*:<core:byte>*                         => c (where b* = core:utf8(c))
+val(string)                ::= v:<core:name>                           => v
+val(i:<typeidx>)           ::= v:<val(type-index-space[i])>            => v
+val((record (field l t)+)) ::= v+:<val(t)>+                            => (record v+)
+val((variant (case l t?)+) ::= i:<core:u32> v?:<val(t[i])>?            => (variant l[i] v?)
+val((list t))              ::= v:vec(<val(t)>)                         => (list v)
+val((tuple t+))            ::= v+:<val(t)>+                            => (tuple v+)
+val((flags l+))            ::= (v:<core:byte>)^N                       => (flags (l[i] for i in 0..N-1 if v & 2^i > 0)) (where N = ceil(|l+| / 8))
```
lukewagner, line 383: I think this rule needs to be updated to have the range `0..|l+|-1` and also probe the appropriate index into `v*` and the appropriate bit modulo 8, etc.  Overall, it'll be important to specify the "endian-ness" of these bytes, and I think what we want is little-endian (i.e, the first flag goes in the LSB of v[0]).

design/mvp/Binary.md
```diff
@@ -350,6 +353,66 @@ Notes:
...
+    - [`core:name`]
+    - [`core:s8`]
+    - [`core:s16`]
+    - [`core:s32`]
+    - [`core:s64`]
+    - [`core:u8`]
+    - [`core:u16`]
+    - [`core:u32`]
+    - [`core:u64`]
+    - [`core:uN`]
```
lukewagner, line 408:
```suggestion
```

design/mvp/Binary.md
```diff
@@ -379,7 +442,18 @@ appear once within a `name` section, for example component instances can only be
...
 
+[`core:s8`]: https://webassembly.github.io/spec/core/binary/values.html#integers
+[`core:u8`]: https://webassembly.github.io/spec/core/binary/values.html#integers
+[`core:s16`]: https://webassembly.github.io/spec/core/binary/values.html#integers
+[`core:u16`]: https://webassembly.github.io/spec/core/binary/values.html#integers
+[`core:s32`]: https://webassembly.github.io/spec/core/binary/values.html#integers
 [`core:u32`]: https://webassembly.github.io/spec/core/binary/values.html#integers
+[`core:s64`]: https://webassembly.github.io/spec/core/binary/values.html#integers
+[`core:u64`]: https://webassembly.github.io/spec/core/binary/values.html#integers
+[`core:uN`]: https://webassembly.github.io/spec/core/binary/values.html#integers
```
lukewagner, line 453:
```suggestion
```

design/mvp/Explainer.md
```diff
@@ -1355,22 +1360,116 @@ number of threads that can be expected to execute concurrently.
...
-instantiation. Unlike modules, components can call start functions at multiple
-points during instantiation with each such call having parameters and results.
-Thus, `start` definitions in components look like function calls:
 ```ebnf
-start ::= (start <funcidx> (value <valueidx>)* (result (value <id>?))*)
+value    ::= (value <id>? <valtype> <val>)
+val      ::= false | true
+           | <core:u8>
+           | <core:u16>
+           | <core:u32>
```
lukewagner, line 1376:
```suggestion
```

design/mvp/Explainer.md
```diff
@@ -1355,22 +1360,116 @@ number of threads that can be expected to execute concurrently.
...
+           | <core:u64>
+           | <core:s8>
+           | <core:s16>
+           | <core:s32>
```
lukewagner, line 1380:
```suggestion
```

design/mvp/Explainer.md
```diff
@@ -1355,22 +1360,116 @@ number of threads that can be expected to execute concurrently.
...
+           | <core:s64>
+           | <f32canon>
+           | nan
```
lukewagner, line 1383:
```suggestion
```

design/mvp/Explainer.md
```diff
@@ -1355,22 +1360,116 @@ number of threads that can be expected to execute concurrently.
...
+           | (record <val>+)
+           | (variant "<label>" <val>?)
+           | (list <val>*)
+           | (tuple <val>+)
+           | (flags "<label>"*)
+           | (enum "<label>")
+           | none | (some <val>)
+           | ok | (ok <val>) | error | (error <val>)
+           | (binary <core:datastring>)
+f32canon ::= <core:f32> without the `nan:0x` case.
```
lukewagner, line 1397:
```suggestion
```

design/mvp/Explainer.md
```diff
@@ -1355,22 +1360,116 @@ number of threads that can be expected to execute concurrently.
...
+f64canon ::= <core:f64> without the `nan:0x` case.
+```
+
+The validation rules for `value` require the `val` to match the `valtype`.  For example:
```
lukewagner, line 1401: Could you add a sentence that the `(binary ...)` expression provides an alternative syntax allowing the binary contents of the value definition to be written directly in the text format, analogous to data segments, avoiding the need to understand type information when encoding or decoding?

design/mvp/Explainer.md
```diff
@@ -1355,22 +1360,116 @@ number of threads that can be expected to execute concurrently.
...
 
-Like modules, components can have start functions that are called during
-instantiation. Unlike modules, components can call start functions at multiple
-points during instantiation with each such call having parameters and results.
-Thus, `start` definitions in components look like function calls:
 ```ebnf
-start ::= (start <funcidx> (value <valueidx>)* (result (value <id>?))*)
+value    ::= (value <id>? <valtype> <val>)
+val      ::= false | true
+           | <core:u8>
```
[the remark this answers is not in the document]
lukewagner: Since we're in the text format here and back to assuming we have a target type, I think we only need `<core:u64>` and `<core:s64>`, and values too big for the target integer type will be rejected as a normal part of validation.

design/mvp/Binary.md
rvolosatovs: I did actually use `uN` in:
```ebnf
value ::= t:<valtype> len:<uN> v:<val(t)> => (value t v) (where len = ||v||)
```
idea being that `N` is potentially larger than 32, since it may encode an arbitrary amount of values, which may be up to `0xffff_ffff` bytes long

I do think, however, we should define `N` as a `ceil(sqrt(|| v ||))` because otherwise it's unbound

design/mvp/Binary.md
rvolosatovs: I believe this is now addressed?

design/mvp/Binary.md
rvolosatovs: Lost this comment in the discussion, addressed in https://github.com/WebAssembly/component-model/pull/336/commits/a56e79239b8f7a989e8c6f7def1803c8bf19e0d8

design/mvp/Binary.md
rvolosatovs: this seems to be addressed now?

rvolosatovs:
I think this is ready for the next review round

design/mvp/Binary.md
lukewagner: Ah, gotcha.  Since it's always backwards-compatible to relax a `core:u32` to a `core:64` and a `core:64` to a `core:u128` in the future, what we usually do in core wasm is use a `core:u32` and say "if we hit the 4gb limit in the future, we can always relax this", so I'd recommend that.  If you are anticipating >4gb values, it'd also be totally fine to plug in `core:u64` too (which Should Be Enough For Anyone).

design/mvp/Explainer.md
```diff
@@ -1355,22 +1360,112 @@ number of threads that can be expected to execute concurrently.
...
+           | (enum "<label>")
+           | none | (some <val>)
+           | ok | (ok <val>) | error | (error <val>)
+           | (binary <core:datastring>)
+f64canon ::= <core:f64> without the `nan:0x` case.
+```
+
+The validation rules for `value` require the `val` to match the `valtype`.
+
+`(binary ...)` expression provides an alternative syntax allowing the binary contents
```
lukewagner, line 1393:
```suggestion
The `(binary ...)` expression form provides an alternative syntax allowing the binary contents
```

design/mvp/Binary.md
lukewagner: Yes, perfect, thanks!

lukewagner (approved):
Awesome, thanks again!  I think that addresses all the feedback I'm aware of.  I'll merge at the end of the week if nothing else pops up.

PR #337, opened 2024-04-10
Re-add case-insensitive uniqueness

lukewagner:
This PR (re)adds the validation rule that two names in a scope can't differ only in case.  This rule was discussed somewhere in a WIP PR, implemented in the current 0.2 Wasmtime implementation, but somehow it was lost on the path to getting merged in `main`, so this PR re-adds it, bringing the spec in sync with the impl.


alexcrichton (approved):

lukewagner:
(Updated WIT.md to match)

Issue #313, opened 2024-02-28
Support inline packages


tschneidereit:
@alexcrichton brought up the idea of supporting packages in-line in a file in addition to the `package` statement. This would help with bundling world definitions into single files as part of things like runtime / language SDKs, which is often very useful.

As a concrete example, wasmtime is [considering](https://github.com/bytecodealliance/wasmtime/pull/7929) to include binary representations of the two worlds it implements instead of the `deps` folder to ease artifact management. That however makes handling these artifacts more difficult in other ways, including simple human inspection.

If we were able to bundle all the wit files into a single one, containing inline packages, that'd be a much better solution.

The syntax for this seems pretty straightforward to me, but I might be missing something:

```wit
package wasi:http@0.2.0 {
  world proxy { ... }
  interface incoming-handler { ... }
  interface types { ... }
}

package wasi:random@0.2.0 {
  world imports { ... }
  interface random { ... }
  interface insecure { ... }
  interface insecure-seed { ... }
}
[... other packages as needed]
```

lukewagner:
Yep, that use case and syntax makes sense to me.  I think we can even encode the WIT syntax you wrote today as:
```wit
(component
  (type (export "wasi:http/proxy@0.2.0") ...)
  (type (export "wasi:http/incoming-handler@0.2.0") ...)
  (type (export "wasi:http/types@0.2.0") ...)
  (type (export "wasi:random/imports@0.2.0") ...)
  (type (export "wasi:random/random@0.2.0") ...)
  (type (export "wasi:random/insecure@0.2.0") ...)
  (type (export "wasi:random/insecure-seed@0.2.0") ...)
)
```
Rendering that back to WIT would then just group `interface`/`world` definitions by package.  So this could be purely a WIT-level addition and not even require a change in encoding scheme.

Another closely-related use case that I think wants the same feature is "show me a single self-contained (no deps) .wit for a given component".  Right now this isn't possible because you can't put multiple packages into a single `.wit`, so this feature would also unlock this use case.  But there is an interesting question:

Let's say I'm rendering the WIT from a `component.wasm` that looks like:
```wat
;; component.wasm
(component
  (import "wasi:http/outgoing-handler" ...)
  (import "custom-callback" (func (param "s" string) (result string)))
  ...
  (export "wasi:http/incoming-handler" ...)
)
```
[...]

PR #340, opened 2024-04-12
Describe "explicit"-style WIT package grammar

azaslavsky:
This enables the inclusion of multiple package definitions in a single `.wit` file, as discussed in
https://github.com/WebAssembly/component-model/issues/313.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -55,9 +55,22 @@ or
 package wasi:clocks@1.2.0;
 ```
 
-WIT packages can be defined in a collection of files and at least one of them
-must specify a package name. Multiple files can specify a `package` and
-they must all agree on what the package name is.
+WIT packages can be defined in a collection of files. At least one of these
+files must specify a package name. Multiple files can specify the `package`,
+though they must all agree on what the package name is.
+
+Alternatively, many packages can be declared consecutively in one or more
+files, if the "explicit" package notation is used:
+
+```wit
+package local:a {
+  interface foo {}
+}
+
+package local:b {
+  interface bar {}
+}
+```
 
 Package names are used to generate the [names of imports and exports]
 in the Component Model's representation of [`interface`s][interfaces] and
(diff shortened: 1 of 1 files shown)


alexcrichton:
An interesting consequence of being able to intermix files that have `package foo:bar;` and files that have `package foo:baz { ... }` is that we may no longer need the `deps` directory in `wit-parser`. You could, for example, have `wasi-*.wit` in the `wit` folder directly perhaps. Not that this is necessarily a good idea, but it's an idea this would open up.

ydnar:
This is great—it would allow round-trip serialization of a `Resolve` back into WIT in a single consumable file.

lukewagner (approved):
Looks great, thanks for working on this!

lukewagner:
Does this look good to merge, or would we like to wait for any additional implementation feedback?

azaslavsky:
I'm happy to merge - do y'all do contributor-triggered merges, or
maintainer?

lukewagner:
Mostly the latter.  For WIT changes, I've generally been trying to keep WIT.md reflective of what is actually in the WIT toolchains (or, if not, gating the feature via a unicode emoji), so I was mostly curious if the implementation of this feature has been implemented and merged.

alexcrichton:
This hasn't yet been merged into wasm-tools (or others as far as I know). That being said I think it's reasonable to merge here 👍

lukewagner:
Ah thanks.  Do you think it might be merged soonish, or would it make sense to emoji-gate this feature?

alexcrichton:
I've not been working on it myself, but @azaslavsky I recall you were working on an initial implementation I think?

azaslavsky:
@alexcrichton I am. Its fallen a bit on my priority list due to $dayjob concerns, but I should have the revised PR up by end of week.

azaslavsky:
https://github.com/bytecodealliance/wasm-tools/pull/1577 implementing this has merged, so I think this can land as well?

lukewagner:
Awesome, thanks again!

Issue #276, opened 2023-11-29
Comments/Questions


rossberg:
I finally did an end-to-end read of all (well, most of) the docs in this repo. Since GH doesn't offer a means to comment on committed files, I created a dummy repo with a dummy PR of the entire content [here](https://github.com/rossberg/dummy/pull/1) and left my comments & questions there. Hope that indirection works well enough. :)

Overall, this looks great. I found the explainer particularly helpful and very well-written.

Happy to open individual issues for some of my comments/suggestions if they turn out to be relevant.

lukewagner:
Awesome, thanks so much!  I'll definitely go through the comments and either turn them into PRs or ask questions here or open new issues, as appropriate.

lukewagner:
Oops, this fell off my radar, but I started going through all the comments just now.  I started with a first pass that simply fixes typos in [this commit](https://github.com/WebAssembly/component-model/commit/4bafd0101b1a9fe3f6e8db520f3221259e8bb771).  For the rest, I'll submit some PRs so they can be reviewed.

lukewagner:
Ok, I went through all the comments and either replied inline or folded the feedback into #339 or #341.  Happy to discuss more inline in those comment threads or on the above PRs or here.

rossberg:
Thanks, I left a few more replies.

lukewagner:
(Also filed #351.)

lukewagner:
Ok, with #353 filed, I think I've gone through all the comments.  Thanks a lot for all the feedback and suggestions!

lukewagner:
Great, thanks for filing #356 for discussion.  I'll close this issue b/c I think that's everything, but feel free to reopen if there's anything missing, or file new issues.

PR #341, opened 2024-04-13
Tweak syntax for how component-level export identifiers are bound

lukewagner:
A good observation in #276 is that the way we bind new `$identifiers` in component-level `export` definitions is a bit irregular, being the only place in WAT where we bind an identifier without the identifier being right after the `<sort>` of the identifier (e.g., in `(import "f" (func $f))` or `(alias $c "e" (type $t))`).  As proposed in this PR, instead of binding a new identifier `$t'` like so:
```wat
(type $t (resource (rep i32)))
(export $t' "t" (type $t))
```
we would instead write:
```wat
(type $t (resource (rep i32)))
(export "t" (type $t) (type $t'))
```
and this would dovetail nicely with the type ascription, so that when you wanted to both bind an identifier and explicitly declare a type, you could write:
```wat
(type $t (resource (rep i32)))
(export "t" (type $t) (type $t' (sub resource)))
```
This PR doesn't affect the AST, semantics or binary format, just how identifier-binding works in the text format, but it is a breaking change.  Thus, if we did go forward with this change, we'd probably want to support both the current and new style for a long time.

I don't think there is a big rush to make this change, though; I think it's mostly just a good cleanup to make before everything is finalized.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -311,10 +311,11 @@ there are three kinds of "targets" for an alias: the `export` of a component
 instance, the `core export` of a core module instance and a definition of an
 `outer` component (containing the current component):
 ```ebnf
-alias            ::= (alias <aliastarget> (<sort> <id>?))
+alias            ::= (alias <aliastarget> bind-id(<sortexpr>))
 aliastarget      ::= export <instanceidx> <string>
                    | core export <core:instanceidx> <core:name>
                    | outer <u32> <u32>
+sortexpr         ::= (<sort>)
 ```
 If present, the `id` of the alias is bound to the new index added by the alias
 and can be used anywhere a normal `id` can be used.
@@ -758,7 +759,7 @@ declarators to be used by subsequent declarators in the type:
 ```wasm
 (component
   (import "fancy-fs" (instance $fancy-fs
-    (export $fs "fs" (instance
+    (export "fs" (instance $fs
       (export "file" (type (sub resource)))
       ;; ...
     ))
@@ -788,8 +789,8 @@ definitions:
     (export "h" (func (result $U)))
     (import "T" (type $T (sub resource)))
     (import "i" (func (param "x" (list (own $T)))))
-    (export $T' "T2" (type (eq $T)))
-    (export $U' "U" (type (sub resource)))
+    (export "T2" (type $T' (eq $T)))
+    (export "U" (type $U' (sub resource)))
     (export "j" (func (param "x" (borrow $T')) (result (own $U'))))
   ))
 )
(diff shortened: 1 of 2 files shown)


rossberg (commented):
LGTM

alexcrichton:
Ok in implementing this I think I've stumbled across an ambiguity which I'm not sure how best to resolve.

This is an existing test case (copied randomly from the list of failures) of what we have today before this PR:

```wasm
(component $C
  (component $m)
  (alias outer $C $m (component $target))
  (export "v" (component $target))
)
```

That is desugared and/or printed with `wasm-tools print`, before this PR, as:

```wasm
(component $C
  (component $m (;0;))
  (alias outer $C $m (component $target (;1;)))
  (export (;2;) "v" (component $target))
)
```

Here the `(;2;)` on the `export` annotation is there to indicate that it's creating a component with index 2. After this PR the above is printed as:

```wasm
(component $C
  (component $m (;0;))
  (alias outer $C $m (component $target (;1;)))
  (export "v" (component $target) (component (;2;)))
)
```

The intention here is to continue to print the index in a comment, hence the `(component (;2;))` here which would otherwise be replaced with `(component $foo)` if it were named. This, however, conflicts with the "inline definition of type syntax". For example if the above module is again piped through `wasm-tools print` it outputs:

```wasm
(component $C
  (component $m (;0;))
  (alias outer $C $m (component $target (;1;)))
  (type (;0;)
    (component)
  )
  (export "v" (component $target) (component (;2;) (type 0)))
)
```

So basically what I'm getting at is that 

```wasm
(export "foo" (component $c) (component $other_name))
```

is ambiguous as to whether you're naming the new component `$other_name` or ascribing it with an empty component type.

---

I believe the above concern is basically only applicable to non-type exports. For types we've got `(eq ..)` or `(sub resource)` to disambiguate.

lukewagner:
Ah, thanks for implementing and pointing this out!  Yes, that ambiguity makes sense.

The least-bad way to resolve this I could think of is to say that `<sortexpr>` is given precedence over `<externdesc>` in `<externopt>`, so that `(sort)` `(sort $id)` and `(sort (; 0 ;))` are never parsed as a type ascription.  With this rule:
* Tooling can unconditionally emit `(sort $id (; idx ;))`, which seems like the priority use case
* If you want to ascribe an empty type, you can, with `(sort (type ...))` (although there's practically zero use case for this)

How does that sound?

alexcrichton:
While I agree that's possible, personally I feel like the current situation is less-bad in the sense that while it's syntactically more wonky it doesn't involve special cases like that. I was trying to think of a syntax that doesn't run into these issues but I was also turning up blanks...

PR #343, opened 2024-04-15
Use little-endian byte ordering for internal conversions.

sunfishcode:
When implementing reinterpret-cast functions, use `python.struct`'s `<` (little-endian) instead of `!` (network byte order).

This has no effective semantic change, because the conversions are just converting between i32/f32 and i64/f64 and are always done in pairs, so it's only required that the decoding match the encoding. However, using little-endian more clearly describes the behavior as corresponding to a Wasm store followed by a Wasm load, which would both be little-endian.

In theory this could become significant in the future if we add SIMD values where endianness conversions are partitioned by SIMD lanes, or other complex types.


lukewagner (approved):
LGTM

rvolosatovs (approved):

PR #346, opened 2024-04-22
enforce enum and variant types being non-empty

rvolosatovs:
Enforce enum and variant type definitions to be non-empty in `Binary.md`

refs https://github.com/WebAssembly/component-model/pull/336#discussion_r1573010660

This is similar to https://github.com/WebAssembly/component-model/pull/218


lukewagner (approved):
Ah, good catch, thanks!

PR #347, opened 2024-04-22
Bump the binary format version to 0xd.

sunfishcode:
Bump the version in the Binary.md document from 0x0a to 0x0d, so that it matches the version [currently in use].

[currently in use]: https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasmparser/src/parser.rs#L25


rvolosatovs (approved):

lukewagner (approved):
Thanks!

sunfishcode:
Taking the liberty of merging this now, to help address some confusion about the relationship between the doc here and current tools. If anyone has comments, please do post them here or file new issues.

PR #350, opened 2024-04-30
Update core type explainer text to match implementation

lukewagner:
I noticed a small discrepancy between the explainer text for how core type definitions work with what's in Binary.md and what's implemented.  What's implemented is actually a bit nicer (symbolic identifiers work, which I just checked), so this PR (should) bring the Explainer up to date.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -493,17 +493,13 @@ implicitly-created `func` type referring to both.
 Lastly, the `core:alias` module declarator allows a module type definition to
 reuse (rather than redefine) type definitions in the enclosing component's core
 type index space via `outer` `type` alias. In the MVP, validation restricts
-`core:alias` module declarators to *only* allow `outer` `type` aliases but,
-in the future, more kinds of aliases would be meaningful and allowed.
+`core:alias` module declarators to *only* allow `outer` `type` aliases (into an
+enclosing component's or component-type's core type index space). In the
+future, more kinds of aliases would be meaningful and allowed.
 
 As an example, the following component defines two semantically-equivalent
 module types, where the former defines the function type via `type` declarator
-and the latter refers via `alias` declarator. Note that, since core type
-definitions are validated in a Core WebAssembly context that doesn't "know"
-anything about components, the module type `$C2` can't name `$C` directly in
-the text format but must instead use the appropriate [de Bruijn] index (`1`).
-In both cases, the defined/aliased function type is given index `0` since
-module types always start with an empty type index space.
+and the latter refers via `alias` declarator.
 ```wasm
 (component $C
   (core type $C1 (module
@@ -513,7 +509,7 @@ module types always start with an empty type index space.
   ))
   (core type $F (func (param i32) (result i32)))
   (core type $C2 (module
-    (alias outer 1 $F (type))
+    (alias outer $C $F (type))
     (import "a" "b" (func (type 0)))
     (export "c" (func (type 0)))
   ))
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -160,9 +160,6 @@ Notes:
   initially-empty type index space.
 * `alias` declarators currently only allow `outer` `type` aliases but
   would add `export` aliases when core wasm adds type exports.
-* Validation of `outer` aliases cannot see beyond the enclosing core type index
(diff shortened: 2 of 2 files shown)


design/mvp/Explainer.md
```diff
@@ -493,17 +493,13 @@ implicitly-created `func` type referring to both.
 Lastly, the `core:alias` module declarator allows a module type definition to
 reuse (rather than redefine) type definitions in the enclosing component's core
 type index space via `outer` `type` alias. In the MVP, validation restricts
-`core:alias` module declarators to *only* allow `outer` `type` aliases but,
-in the future, more kinds of aliases would be meaningful and allowed.
+`core:alias` module declarators to *only* allow `outer` `type` aliases to the
+immediately-enclosing component or component-type. In the future, more kinds of
```
rossberg, line 497: If I understand the intention correctly, then you may rather want to say
```suggestion
innermost enclosing component or component-type. In the future, more kinds of
```
since the "immediately-enclosing" scope may not be a component scope but a core module scope.

But why restrict to the innermost?

lukewagner: Here I'm just updating to match the current impl.  But the basic reason is that it's less work and also sufficient to achieve the goal which is sharing type definitions.

design/mvp/Explainer.md
```diff
@@ -493,17 +493,13 @@ implicitly-created `func` type referring to both.
 Lastly, the `core:alias` module declarator allows a module type definition to
 reuse (rather than redefine) type definitions in the enclosing component's core
 type index space via `outer` `type` alias. In the MVP, validation restricts
-`core:alias` module declarators to *only* allow `outer` `type` aliases but,
-in the future, more kinds of aliases would be meaningful and allowed.
+`core:alias` module declarators to *only* allow `outer` `type` aliases to the
+innermost enclosing component or component-type. In the future, more kinds of
```
alexcrichton, line 497: I'm a bit confused by this sentence as well insofar as wasm-tools considers this valid

```wasm
(component $A
  (core type $t (func))
  (component $B
    (component $C
      (alias outer $A $t (core type $t2))
    )
  )
)
```

it was my intention to support arbitrary levels of nesting for aliases, but I wouldn't put it past me to have made a mistake here. 

Basically I think it's reasonable to keep the spec as-is in this regard and fix the implementation instead.

lukewagner: Ah yes, I guess I would need to further nuance the wording to allow that case.  What I was mostly thinking about is that, atm, this example fails:
```wat
(component $A
  (core type $t (func))
  (component $B
    (core type (module
      (alias outer $A $t (type $t2))
    ))
  )
)
```
So are you saying you think this should work too?

alexcrichton: Aha, I see now! I didn't even realize that was an error. Looks like the implementation traces itself back to https://github.com/bytecodealliance/wasm-tools/pull/640 which implements https://github.com/WebAssembly/component-model/pull/44 which specifically mentions the 1-depth limit. (I'm learning this all as I'm going here too!)

It looks like that wording is still in `Binary.md`, although I can't immediately recall why such wording is there. That may also mean that `wasm-tools` isn't validating my component I pasted above correctly since it should be rejecting it like the component you have. The component you have though is rejected at the `wast` layer though rather than the `wasmparser` binary validation layer, so I'd also need to play around with things to see if the same validation is in `wasmparser`.

lukewagner: If we're changing implementation anyways (to either accept more or reject more), if it's not too much work (b/c this doesn't seem like a huge priority, other than getting spec and impl to line up), would it make sense to remove the 1-depth limit?

alexcrichton: Yeah that's my thinking as well. I'll get to that probably next week and make sure it works out and post here?

alexcrichton: As probably expected [this was easy enough to implement](https://github.com/bytecodealliance/wasm-tools/pull/1539) so I don't think there's any issue lifting the restrictions here

lukewagner: Ok, great to hear, thanks!

lukewagner:
Ok, I updated the PR to relax the restrictions and match the forthcoming implementation.  @alexcrichton @rossberg PTAL

alexcrichton (approved):

Issue #239, opened 2023-08-29
[Documentation] static core linking


penzn:
@pchickey @lukewagner, we had a discussion of static linking in context of WebAssembly/WASI#549, and I just wanted to make sure I understood it correctly:

Currently static core linking works by shipping static libs or wasm 'object files' in the component binary and then running `wasm-ld` in the runtime during component instantiation.

Is this an accurate description? Can you please point me to where this is currently documented?

pchickey:
`wasm-ld` is able to statically link wasm object files and archive files (static libs) into a core wasm module.

You can then transform a core wasm module into a component using a tool such as `wit-component`.

All of these things happen ahead of time, before the component is loaded into a runtime for instantiated.

penzn:
So what is the relationship between `wasm-ld` and `wit-component`, I thought you mentioned that the latter uses the former or something like that. Let me know if there is already a place to read about it, so I don't bug you with questions already answered elsewhere :)

alexcrichton:
I'm not aware of any specific documentation myself, but these tools are quite low-level so it's sort of similar to asking what the relationship is between `clang` and `ld` in a sense. It's probably best to understand the tools from an inputs/outputs perspective where:

* `wasm-ld` takes in compiler object files, those generated by `clang` or `rustc`, and generates a core wasm module
* `wit-component` takes a core wasm module and some type information, previously embedded in the core wasm module, and generates a component

The `wit-component` support is expoed through the `wasm-tools component new` subcomand (in [this repo](https://github.com/bytecodealliance/wasm-tools)) and you can explore that with `wasm-tools component new -h`. For `wasm-ld` you can also explore with `wasm-ld -h` and binaries I know are available from the wasi-sdk repository but if you already have LLD installed you may already have `wasm-ld` installed as well.

penzn:
@pchickey made a reference to `wasm-ld` being used by either a wit tool or runtime, and I am trying to follow up on the mechanics of that.

I know what `wasm-ld` does and what wasm 'object files' really are.

pchickey:
wasm-ld is not used by any of the wit tools or runtimes. On our call last night I described the process by which which you can statically link programs prior to getting involved with any component runtime.

[...]

PR #367, opened 2024-06-12
Add Linking.md to provide a better high-level picture of linking

lukewagner:
This PR adds the diagram discussed in #239 along with some surrounding explanatory text that tries to paint a higher-level picture of the combined use of shared-everything/-nothing linking allowed by the component model.  It also tries to define some of the terms used to describe linking.  Definitely folks let me know if anything sounds off or doesn't accurately reflect the current state of tooling.

(cc @sunfishcode @alexcrichton @dicej )


alexcrichton (approved):

pchickey (commented):
Given that `wasm-compose` is already deprecated in favor of `wac`, I think it should be left out of this explainer.

Another piece of the puzzle that this explainer is missing is that the dynamic shared-everything linking can also be achieved with a component which imports a core module of a given signature. That question came up in Zulip recently and I was able to point to the libc example in the canonical ABI section of the Explainer. I know that tooling support for that is basically nonexistent but at least one person was wondering if it was possible. I don't know exactly how to fit that into this explainer, though.

dicej:
> Another piece of the puzzle that this explainer is missing is that the dynamic shared-everything linking can also be achieved with a component which imports a core module of a given signature.

Minor quibble: that would still be an instantiation-time operation rather than a runtime one, so I don't think it quite qualifies as "dynamic".  But yes, I expect that will be a hugely useful feature; I'd love for `componentize-py` to optionally produce components that _import_ `libpython312.so` instead of bundling it.  And being able to host such modules in a well-known registry would make that a no-brainer.

design/mvp/Linking.md
```diff
@@ -0,0 +1,143 @@
...
+
+Next, each C/C++ translation unit is compiled to a [WebAssembly Object File]
+using `clang`, optionally archived together using `ar`, and finally
+**shared-everything statically-linked** using [`wasm-ld`], all without any of
+these tools knowing about the Component Model.
+
+A single Core WebAssembly module can be trivially wrapped into a component
+using a tool like the [`wasm-tools`] `component new` command. Multiple Core
+WebAssembly modules can be **shared-everything dynamically-linked** together
+and loaded imperatively at runtime via emulated `dlopen()` using the `component
```
dicej, line 62: Might want to clarify that `dlopen` emulation is optional and specified on a library-by-library basis when using `component link`.  I.e. you can enable it for none, some, or all of the libraries as desired.

dicej (approved):

lukewagner:
> Given that wasm-compose is already deprecated in favor of wac, I think it should be left out of this explainer.

Done

> Another piece of the puzzle that this explainer is missing is that the dynamic shared-everything linking can also be achieved with a component which imports a core module of a given signature.

Ah, good point.  That is yet another orthogonal aspect (since, as long as you have a registry available) you can import any time you could have nested.  I suppose that would be good to cover.  I'll have to think about how to succinctly describe and visualize that in the diagram...

design/mvp/Linking.md
lukewagner: Ah, interesting!  If you don't enable it for a library, what happens (how do its symbols get exposed to the other modules)?

dicej: `component link`'s main job is to match up imports with exports and create a linear instantiation graph based on those match-ups (plus break any cycles via a synthesized module and `call.indirect`) -- all at build time.

`dlopen` emulation just allows symbol resolution to be delayed until runtime instead of forcing it to happen at build time (or more generally, you might want to resolve symbols in a given library at build time and _also_ allow them to be resolved at runtime).  If you don't need that runtime capability (or only need it for a subset of your libraries), we don't bother creating the lookup tables for libraries that don't need it.

lukewagner: Ohhh, well that's great and my paragraph then is totally wrong (I assumed the thing you just said it does isn't implemented yet).

lukewagner: I [rewrote](https://github.com/WebAssembly/component-model/pull/367/commits/04cc31faef10a9650f99d580c42b106e383ce9ce) that paragraph, PTAL.

lukewagner:
@pchickey and others: I added two paragrams and some visuals to the diagram in [this commit](https://github.com/WebAssembly/component-model/pull/367/commits/44ddf2d25e8d438eb3d6dc2c39bb7ee90f7bab94), PTAL

lukewagner:
Also @lann, it'd be useful to hear if the blurbs make sense to you, since the PR mentions `wasm-pkg-tools` now.

design/mvp/Linking.md
```diff
@@ -0,0 +1,180 @@
...
+an external registry. This registry toolchain integration is still in progress,
+but by reusing common support libries such as [`wasm-pkg-tools`], higher-level
+tooling can uniformly interact with multiple kinds of storage backends such as
+local directories, [OCI Wasm Artifacts] stored in standard [OCI Registries] and
+[warg registries]. Of note, even when modules or components are stored inline
+by earlier stages of the build pipeline, when creating an OCI Wasm Artifact, a
+toolchain can enable deduplication by content-hash of common modules or
+components by placing them in separate OCI [`layers`] which are imported via
+[`hashname`] by the root component stored in the first layer of the OCI Wasm
+Artifact.
```
lann, line 99: It may not matter here but this is not yet implemented anywhere afaik.

lukewagner: True, to better set expectations (or maybe entice contributions), I added a [parenthetical](https://github.com/WebAssembly/component-model/pull/367/commits/74c1b9a98b313e2e6cf1db93aee12c417f458370) qualifying this as NYI.

lukewagner:
(I added an FAQ entry.)

Issue #356, opened 2024-05-08
Make function types more general and symmetric


rossberg:
Currently, the grammar for function types is as follows:
```
functype      ::= (func <paramlist> <resultlist>)
paramlist     ::= (param "<label>" <valtype>)*
resultlist    ::= (result "<label>" <valtype>)* | (result <valtype>)
```
All parameters and results must be named, except a singleton result.

There is broad variety across programming languages in whether and how they allow/require/distinguish named vs unnamed parameters, as well as unnamed vs unnamed results, at a function's def site, use site, or both. The current design is rather specific in that regard and somewhat biased. From purely an interface perspective, there is no reason to treat parameters differently from results, or allow omitting names in some cases but not others. And at least for some languages, it would be useful to make explicit which components have "proper" names and which ones haven't, since that enables more idiomatic bindings.

See [here](https://github.com/rossberg/dummy/pull/1#discussion_r1407621103) for more discussion.

There are several degrees to which the grammar could be generalised:
```
paramlist     ::= (param "<label>" <valtype>)* | (param <valtype>)
resultlist    ::= (result "<label>" <valtype>)* | (result <valtype>)
```
or
```
paramlist     ::= (param "<label>" <valtype>)* | (param <valtype>)*
resultlist    ::= (result "<label>" <valtype>)* | (result <valtype>)*
```
or
```
paramlist     ::= (param "<label>"? <valtype>)*
resultlist    ::= (result "<label>"? <valtype>)*
```
I'd suggest one of the latter two, which make names uniformly optional, and then specify a canonical scheme for synthesised names in contexts/bind-gens that need them, for example, `"_1"`, `"_2"`, etc. based on position. This would apply symmetrically to parameters and results.

lukewagner:
Personally, I like the symmetry in the abstract, and I can think of a few times where, when writing a WIT interface, I feel like I'm being forced to add a parameter name that adds no value (e.g., `handle: func(request: request) -> result<response, error-code>`).  I do worry, though, that the additional degree of freedom invites more subjective stylistic variation (some folks are going to want to give everything a parameter name, others are going to want to leave them off by default).  E.g., WASI would need to establish a style guideline on this with the criteria for named-vs-unnamed.

[...]

PR #368, opened 2024-06-14
Remove named multi-return from function types (for now at least)

lukewagner:
Currently function types are slightly asymmetric with results having an extra single-unnamed-result option.  As #356 points out, we could make params and results fully symmetric by saying that both could either be a list of all-named- or all-unnamed types.  I think that topic requires some more discussion to motivate and understand the implications for bindings generators in various languages.

In the meantime, though, I don't think anyone uses or even knows about named multi-return -- the much more common thing to do is to return a `tuple` or `record` (almost always wrapped in a `result`).  I think this means we can expect to see
an ad hoc mix of multi-return and `tuple`/`record` in practice which seems like needless inconsistency (and wasted time arguing over what if any convention there should be).  It also seems like this path will be either unimplemented or untested in current bindings generators, which bodes poorly if anyone ever does decide to use this feature.

Thus, this PR proposes that, at least until we decide to go for full symmetry, we lean into the asymmetry (which is already present) and disable this additional case of named multi-return that (afaik) noone is using.  If however anyone is using and appreciating named multi-return, I'd be very interested to hear about it!  (I'll leave this PR open for a while to collect feedback.)

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -204,7 +204,7 @@ resourcetype  ::= 0x3f 0x7f f?:<funcidx>?                 => (resource (rep i32)
 functype      ::= 0x40 ps:<paramlist> rs:<resultlist>     => (func ps rs)
 paramlist     ::= lt*:vec(<labelvaltype>)                 => (param lt)*
 resultlist    ::= 0x00 t:<valtype>                        => (result t)
-                | 0x01 lt*:vec(<labelvaltype>)            => (result lt)*
+                | 0x01 0x00                               =>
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):

Mossaka (approved):
This will simplify the specification a lot. Thanks

lukewagner:
Ok, since noone seems to be using multi-return afaics and this is a conservative change we can always relax later (via #356), I'll go ahead and merge now (which may shake out more use cases, which will be useful data).

alexcrichton:
This change is going to be shipped with Wasmtime 24, to be released August 20, as part of the update in https://github.com/bytecodealliance/wasmtime/pull/8965

Issue #301, opened 2024-02-07
Import name projections don't support exports by interface name


peterhuene:
Currently the [projection rule](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions) for import names looks like:

```ebnf
projection    ::= '/' <label>
label         ::= <word>
                | <label> '-' <word>
```

This prevents projections for exports using the `interfacename` format, as these cannot be expressed simply as a `label`.

Say component `foo:bar` exports an instance with name `bar:baz/qux` (i.e. in `interfacename` format) and we wanted to import this instance from another component, but reference a possible implementation using an `unlocked-dep` import rather an import of an interface name. We're unable to express this because `bar:baz/qux` cannot be represented as a projection.

One possible solution:

Extend the `projection` rule as follows:

```ebnf
projection          ::= '/' <projection-segment>
projection-segment  ::= <label> | '<' <interfacename> '>' 
```

Or using some other delimiter to denote that the segment is an interface name and not a label.

lukewagner:
Interesting question!  Could one solution be to say that, if you are wanting to import the, let's say, `wasi:http/incoming-handler` interface of a component with package name `ns:foo`, you could generate an import:
```wat
(import "unlocked-dep=<ns:foo>" (instance
  (export "wasi:http/incoming-handler" (instance
    (export "handle" (func ...))
  ))
))
```
Hence, we're using nested instance types and subtyping to project out just the subset of the `ns:foo/bar` component that we want.

Do you think that could work?  What's kindof neat about this approach is that is supports arbitrary subsets of a component's exports while still maintaining that they came from the same instance and share resource types.

peterhuene:
That's the ideal solution at the component model level, but WIT cannot express that (i.e. an instance exporting an instance), which is what source bindings generation is based off of and that ultimately controls the component type one can build in the guest languages that use WIT for "componentization".

If we ditched WIT to describe the resulting component type, then we could encode a more complex import.

lukewagner:
Gotcha, makes sense.  Since we haven't yet extended WIT to express locked/unlocked deps at all (mea culpa), perhaps the right fix here is to consider what the right new WIT syntax is for locked/unlocked deps and then see how this use case fits into the WIT syntax so that it can be encoded as a component-level import like I showed above.  Does that sound right to you?

peterhuene:
[...]

PR #372, opened 2024-06-24
Nested interfaces

macovedj:
Here is an outline for the various ways in which one could describe nested interfaces in wit.

For additional context re: motivation, see [this issue](https://github.com/WebAssembly/component-model/issues/301), though generally what this enables is the ability to target nested instances which are expressible in wat and binary formats, and not in wit.

An initial implementation that still needs some work can be found [here](https://github.com/bytecodealliance/wasm-tools/pull/1624), though at the moment, it would only add support for the case of nesting interfaces from foreign packages.

Perhaps this PR can also be a place to hash out a bit of what the implications would be from a guest lang bindgen perspective as well.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -662,8 +662,8 @@ world w2 {
 
 ## Nesting WIT interfaces
 [nest]: #nesting-wit-interfaces
-Interfaces can also export other interfaces via the `nest` keyword.
-With the `nest` keyword, one can reference other interfaces in the same package, foreign packages, or simply define anonymous interfaces inline.
+Interfaces can also contain other interfaces via the `nest` keyword.
+With the `nest` keyword, one can reference other interfaces in the same package, foreign packages, or define anonymous interfaces inline.
 
 ```wit
 package local:example;
@@ -673,7 +673,7 @@ interface foo {
 }
 
 interface top {
-  nest foo
+  nest foo;
   nest foreign:pkg/bar;
   baz: interface {
     ...
@@ -685,31 +685,45 @@ Each of these forms of nesting interfaces are encoded as:
 
 ```wasm
 (component
-  (type (;0;)
-    (instance
-      ... `types from foo`
-    )
-  )
-  (export "local:example/foo" (type 0))
-  (type (;1;) 
-    (instance
-      (alias outer 0 0 (type (;0;)))
-      (export "local:example/foo" (instance (type 0)))
-      (type (;1;) 
-        (instance 
-          ... `types from foreign:pkg/bar`
+  (type (;0;) 
+    (component 
+      (type (;0;)
+        (instance
+          ... `types from foo`
         )
       )
-      (export "foreign:pkg/bar" (instance (type 1)))
-      (type (;2;) 
-        (instance 
-          ... `types from baz`
-        )
-      )
-      (export "baz" (instance (type 2)))
+      (export "local:example/foo" (type 0))
     )
   )
-  (export "local:example/top (type 1))
+  (export (;1;) "foo" (type 0))
+  (type (;2;)
+    (component
+      (type (;0;) 
+        (instance
+          (type (;0;)
+            (instance
+              ... `types from foo`
+            )
+          )
+          (export "local:example/foo" (instance (type 0)))
+          (type (;1;) 
+            (instance 
+              ... `types from foreign:pkg/bar`
+            )
+          )
+          (export "foreign:pkg/bar" (instance (type 1)))
+          (type (;2;) 
+            (instance 
+              ... `types from baz`
+            )
+          )
+          (export "baz" (instance (type 2)))
+        )
+      )
+      (export "local:example/top (type 1))
+    )
+  )
+  (export (;3;) "top" (type 2))
 )
 ```
 
@@ -1159,6 +1173,7 @@ interface-items ::= gate interface-definition
 interface-definition ::= typedef-item
                        | use-item
                        | func-item
+                       | nest-item
 
 typedef-item ::= resource-item
                | variant-items
@@ -1181,6 +1196,9 @@ named-type-list ::= ϵ
                   | named-type ( ',' named-type )*
 
 named-type ::= id ':' ty
+
+nest-item ::= gate nest use-path ';'
+            | id: interface-item
 ```
 
 
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -116,7 +116,7 @@ would correspond to:
 )
 ```
 
-An `interface` can contain [`use`][use] statements, [type][types] definitions,
+An `interface` can contain [`use`][use] and [`nest`][nest] statements, [type][types] definitions,
 and [function][functions] definitions. For example:
 
 ```wit
@@ -136,7 +136,7 @@ interface types {
 }
 ```
 
-More information about [`use`][use] and [types] are described below, but this
+More information about [`use`][use], [`nest`][nest], and [types] are described below, but this
 is an example of a collection of items within an `interface`. All items defined
 in an `interface`, including [`use`][use] items, are considered as exports from
 the interface. This means that types can further be used from the interface by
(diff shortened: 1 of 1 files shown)


design/mvp/WIT.md
```diff
@@ -660,6 +660,59 @@ world w2 {
...
+
+```wit
+package local:example;
+
+interface foo {
+  ...
+}
+
+interface top {
+  nest foo
```
alexcrichton, line 676: I think this might be missing a semicolon?

design/mvp/WIT.md
```diff
@@ -660,6 +660,59 @@ world w2 {
...
+  nest foreign:pkg/bar;
+  baz: interface {
+    ...
+  }
```
alexcrichton, line 680: Syntactically I might bikeshed this as `nest baz { ... }` perhaps? (to keep the word "nest" around it and not have too too many keywords)

alexcrichton (commented):
Thanks! Could this also update the pseudo-ebnf [below too](https://github.com/WebAssembly/component-model/blob/9eda88291cac30a3aa6d69aed0ab0f147e0719db/design/mvp/WIT.md#L796)?

rossberg:
It’s great seeing this expressivity gap in WIT addressed. However, I don’t think the proposed extension is the right solution for this. It is very special-cased and fails to address closely related or more general use cases. For example, it's quite natural to want to express an interface of the form
```
    interface {
       a1 : A
       a2 : A 
    }
```
where both sub elements have the same interface A, but _without_ also defining a standalone A on the outer level. Similarly, a common case are components like
```
    world {
        import i : A;
        export e : A;
    }
```
where the same signature occurs for a generic import and an export. Or perhaps even only two imports:
```
    world {
        import a1 : A;
        import a2 : A;
    }
```
And endless other variations.

The actual root of the problem is that WIT — unlike the raw component system — conflates two entirely different forms of declaration, namely, the named definition of an interface _type_ (an actual interface per se) and declaring the presence of an _instance_ of an interface (the implementation of an interface).

For perspective, compare this to a regular programming language, where we usually distinguish
```
    type T = struct { a : A; b : B }
```
from
```
    export T : struct { a : A; b : B }
```
and these are entirely different things, declaring very different categories of things. I doubt anybody would consider merging both into a single form of declaration a helpful thing to do.

An addition like the one proposed in this PR is not fixing this underlying category error, but arguably makes the conflation worse, very likely requiring further ad-hoc work-around features in the future.

The more adequate and scalable solution would be to properly separate these notions in WIT as well. That is, introduce a new form of declaration for naming an interface type, that then can be used in all places where an interface description would occur, but does not by itself declare an instance.

The example from the PR would become something along the lines of
```
    interface type Foo {
        ...
    }

    interface foo : Foo

    interface top {
        foo : Foo;
        bar : foreign:pkg/Bar;
        baz : interface {
            ...
        }
    }
```
but importantly, you don’t _have to_ declare the existence of a global `foo` if all you actually want is to name the type `Foo`, because it's used in multiple places.

macovedj:
Hmm, can we talk about this in terms of how it would be encoded?
Given that today the `interface` syntax is encoded as follows:

```wit
interface foo {
  ...
}
```

```wasm
(component
  (type (;0;)(instance 
  ...
  ))
  (export "ns:id/foo" (type 0))
)
```

We can see that `interface` is currently encoded as a type that is exported.

Note in the proposal, the `nest` keyword is exporting an `instance` rather than a `type`, which is a difference between nested interfaces and top level interfaces.

I may be misinterpreting, but it feels like the `interface` keyword currently behaves how I would intuitively expect the proposed `interface type` syntax to behave, and it feels like `nest` is implicitly operating in a way that feels similar to `foo: Foo`.

Given the current usage of `interface`, if we were to address the larger concerns here, would it maybe make more sense to go the opposite direction?

Thinking of `interface` as operating currently how `interface type` is proposed to behave, instead we could do something like

```wit
interface foo-type {
  ...
} 

foo: instance foo-type;

interface top {
  foo-inst: foo,
  top-foo-type: foo-type
}
```

Or using @rossberg's initial suggestion, not even introduce the `instance` keyword and just have

`foo: foo-type`

but allow it to act on an `interface` rather than an `interface type`

Then `top` has the liberty of exporting either or both `top-foo-type` as a `type` like what we currently see in top level interfaces,  or a `foo-inst` as an `instance`, as is originally proposed in this PR.  If I'm reading correctly, this feels like it accomplishes disambiguating without changing the current semantics of `interface`.

lukewagner:
@rossberg In your example WIT, you're giving all 3 nested interfaces a `plainname`, i.e., they turn into the following instance type:
```wat
(type $top (instance
  (export "foo" (instance ...))
  (export "bar" (instance ...))
  (export "baz" (instance ...))
))
```
That's fine, and we can quibble about the right syntax for addressing all the use cases you're getting at, but that is all quite independent from the use case we are trying to address in this PR which is that we want an export of an `interfacename` (that has been separately defined and named).  e.g., we want:
```wat
(type $top (instance
  (export "wasi:http/incoming-handler" ...)
))
```
which is this PR is proposing looks like:
```wit
interface top {
  nest wasi:http/incoming-handler;
}
```
Notice that we are quite intentionally *not* assigning a locally made-up `plainname`; we want to export the nested `instance` by specifying only its `interfacename`.  I don't think any of the syntax you're proposing addresses this use case?

design/mvp/WIT.md
lukewagner: I can see both arguments, but while writing `nest a: interface { ... };` is more symmetric with `nest foo:bar/baz;` writing `a: interface { ... };` is more symmetric with `a: func() -> blah;` (in interfaces) and `import a: interface { ... };`  (in worlds).  Since the reason for having `nest` in the first place is because `<interfacename>;` is lexically ambiguous (due to the `:` inside `<interfacename>`) and also looks odd, and since with `a: interface { ... };` there's no lexical ambiguity and it looks roughly like what it means in worlds, my vague preference is to go with the latter symmetry.

lukewagner (commented):
Thanks for working on this!  Looks generally good to me; here's a few initial comments, I'll need to review a bit more carefully later.

design/mvp/WIT.md
macovedj: Left as is for now, though I noticed Luke was using semicolons for anonymous interfaces, while we weren't using them before... not sure if people feel one way or another about that, but I don't think it's ambiguous to leave them out, similarly to how records are handled.

macovedj:
Some of the other concerns raised in the PR that implements nesting foreign packages included if we've considered what this should look like from the perspective of guest bindings.  Spitballing with rust, currently if we have the following wit

```wit
package foo:bar@1.0.0;

interface things {
  record my-record {
    foo: string
  }
}
world my-world {
  export things;
}
```

We end up with the following

```rs
#[allow(dead_code)]
pub mod exports {
  #[allow(dead_code)]
  pub mod foo {
    #[allow(dead_code)]
    pub mod bar {
      #[allow(dead_code, clippy::all)]
      pub mod things {
        ...
     }
   }
  }
}
```

Does it make sense to just continue to nest `mods`?
So using thing examples where we have 

```wit
nest foo;
nest foreign:pkg/bar;
nest baz;
...

then `things` above could look as follows:

```
pub mod things {
  ...
  pub mod foo {
    ...
  }
  pub mod foreign {
    pub mod pkg {
      pub mod bar {
        ...
      }
    }  
  }
  pub mod baz {
    ...
  }
}

I'm guessing I may be missing something, though I'd guess that other languages namespace things based on packages/interfaces similarly...

design/mvp/WIT.md
lukewagner: Oops, I hadn't noticed the precedent for anonymous interfaces in worlds was to leave off the semicolon, so I'd go with that.

lukewagner: Coincidentally, working through the proposed additions in #308 made me realize that a semicolon after anonymous interfaces (in both interfaces and worlds) is probably a good idea for long-term future compatibility.  See e.g. the code examples in [this comment](https://github.com/WebAssembly/component-model/pull/308#issuecomment-2207055542) where I think you end up wanting the `;` always.

macovedj:
I guess the other thoughts worth proposing for bindgen are more detailed mechanics of generating interfaces/traits.

I'd guess that if we have a situations where say if interface A has a generated trait definition `GuestA` and also nests interface `B` with a generated trait definition `GuestB`, then `GuestA` would probably have `GuestB` as an associated type, where the associated type is namespaced as outlined above.

lukewagner:
Good points!  Because nested interfaces have a totally distinct identity (of contained types and functions) from any other nesting or top-level `import`/`export` of the same `interfacename`, I think you're right in your above Rust code to nest `foreign:pkg/bar` under `things` so that it's distinct from, e.g., an `import foreign:pkg/bar` in the same `world`.  To your second question about traits, though: I think they would also need to be kept distinct so that there is a separate trait (which can have separate concrete resource types) for each nested occurrence of an `interfacename`.

Maybe not interesting, but one corner case worth considering is what if we have:
```wit
interface things {
  foreign: interface { ... };
  nest foreign:pkg/bar;
}
```
This is allowed b/c the first nested interface is given the `plainname` `foreign` and the second nested interface is given the `interfacename` `foreign:pkg/bar` (which are distinct and thus allowed) but with the above scheme, both will try to stick a `pub mod foreign` into the `things` module.  I assume the right solution here is to handle this rare collision with some avoidance scheme (maybe add a suffix character to the `foreign` namespace).  I believe this sort of collision also arises at world-level (replacing `nest` with `import`), so it'd be good to know what we do there and probably copy that.

PR #376, opened 2024-07-09
Simplify and tighten up reentrance/backpressure rules for async

lukewagner:
I realized that the current CABI logic underspecifies a few cases and that would seem to allow, in some cases, async calls to block in cases where they intuitively shouldn't.  The two commits in this PR fix this by specifying how async-lowered calls work in a lot more detail (without a lot more code; I found some latent simplifications at part of this) and removing the hand-wavey-ness (particularly around `asyncio.create_task`).

The first commit also makes one semantic change: instead of async tasks implicitly signalling backpressure by `task.wait`ing before `task.start`ing the call, the first commit instead introduces an explicit `task.backpressure` built-in that can be called any time by guest code to toggle a bool on the `ComponentInstance` for whether to exert backpressure or not.  This both simplifies the logic quite a bit and provides more expressivity to the guest code (e.g., you might only realize at the end of a task that you're nearing OOM and need backpressure, and this change lets you do that).  This commit also makes guest-signaled backpressure apply to *all* export calls (not just async-lifted ones) since, if you're explicitly signalling backpressure due to nearing OOM, sync calls are also a problem.

cc @dicej

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -18,6 +18,7 @@ summary of the motivation and animated sketch of the design in action.
   * [Subtask and Supertask](#subtask-and-supertask)
   * [Structured concurrency](#structured-concurrency)
   * [Waiting](#waiting)
+  * [Backpressure](#backpressure)
   * [Starting](#starting)
   * [Returning](#returning)
 * [Examples](#examples)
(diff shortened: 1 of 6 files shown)


design/mvp/Async.md
```diff
@@ -217,12 +232,10 @@ asynchronously-lifted Core WebAssembly function is passed an empty list of
...
 and receive its arguments.
 
-The main reason to have `task.start` is so that an overloaded component
-instance can [wait](#waiting) for existing tasks to finish before starting a
-new task. Once a component instance waits to start a task, the Component Model
-will automatically apply backpressure, preventing new tasks from starting.
-See the [`AsyncTask`] class and [`canon_task_start`] function in the Canonical
-ABI explainer for more details.
+The main reason to have `task.start` is so that an overloaded component can
+[wait](#waiting) and/or exert [backpressure](#backpressure) before accepting
```
dicej, line 236: Can you clarify under what circumstances a component would choose to call `task.wait` vs. `task.backpressure` vs. both when overloaded?

dicej (approved):

design/mvp/Async.md
lukewagner: Good idea, added some prose [here](https://github.com/WebAssembly/component-model/pull/376/commits/e71ef224df4b34134da129e46dcae18da475587e).

PR #377, opened 2024-07-12
Add support for `@deprecated` gates to WIT

### LINKED ISSUE #374: Support `@deprecated` gates in WIT
## Motivation

Following up on the conversation in https://github.com/WebAssembly/wasi-http/issues/107 (renaming `field-key` to `field-name`), we should support some form of `@deprecated` notation to mark individual items as deprecated. In the linked issue I proposed the following flow for that:

> ### Navigating deprecations within a WASI release family
> If we assume for a second this is a change we want to make, that opens up the question of how we would want to navigate a change like this. WASI 0.2 was just ratified, and WASI 0.3 is likely still a while out. And if we were to implement for example #94 (add a `fields.keys` method), we'd be doubling-down on the existing terminology instead of the terminology we would prefer.
> 
> This is a question which is generally broader than this issue, but I think it would be good if we could establish a process for making changes like these _within_ a minor WASI version. An example flow I'm thinking of is:
> 
> 1. Introduce a new alias `type field-name = field-key` (or alias it directly to `string` if that's legal in our projections).
> 2. Change all APIs which take `field-key` to take a `field-name` instead.
> 3. Mark `field-key` as deprecated for WASI 0.2.
> 4. When we release WASI 0.3 we remove `field-key` from the API surface.
> 
> If we were able to do something like this we could release the changes as part of a WASI `0.2.x` release without breaking any existing code. And then in a WASI `0.3.0` release we could finalize the changes by actually removing deprecated APIs. I don't know whether we already have a way to support deprecating APIs in WIT; but if we don't that may be a good thing to consider adding.

## Feature design

Just like `@since` and `@unstable`, it would make sense to add a third gate: `@deprecated`. I think it would make sense for this feature to also carry a `version =` field (just like `@since`) to mark when in which version it was deprecated. As well as some way to provide a message why this API was deprecated. I'm suggesting we use the `message = ` field name for that.

This matches [Rust's approach](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) fairly closely, which has a `#[deprecated(since="", note="")]` notation. Swift is [a little fancier](https://sarunw.com/posts/deprecate-old-api-in-swift/), and has an `@available` notation which supports a number of configurations. What we're proposing here would in Swift be written as `@available(*, deprecated, message: "")`. Swift also supports a `rename` field which exists specifically to point to other symbols. But unless we have an existing path notation for it, I'm hesitant to propose we add it straight away.

Put together I'm proposing we should change [WIT's feature gate notation](https://github.com/WebAssembly/component-model/blob/9eda88291cac30a3aa6d69aed0ab0f147e0719db/design/mvp/WIT.md#feature-gates) to the following:

```ebnf
gate ::= unstable-gate
       | since-gate
       | deprecated-gate

unstable-gate ::= '@unstable' '(' feature-field ')'
since-gate ::= '@since' '(' version-field ( ',' feature-field )? ')'
deprecated-gate ::= '@deprecated' '(' version-field ( ',' message-field )? ')'

feature-field ::= 'feature' '=' id
version-field ::= 'version' '=' <valid semver>
message-field ::= 'message' '=' message
```

cc/ @lukewagner I'd be keen to get your thoughts on this - anything this might be missing, or you think we should do differently?

yoshuawuyts:
Closes #374. Adds support for `@deprecated` gates to WIT. While the text in this PR is ready for review now, we should hold off on merging this until we have implementation feedback. Thanks!

r? @lukewagner

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -923,8 +923,9 @@ package-items ::= toplevel-use-item | interface-item | world-item
 ### Feature Gates
 
 Various WIT items can be "gated", to reflect the fact that the item is part of
-an unstable feature or that the item was added as part of a minor version
-update and shouldn't be used when targeting an earlier minor version.
+an unstable feature, that the item was added as part of a minor version
+update and shouldn't be used when targeting an earlier minor version, or that a
+feature has been deprecated and should no longer be used.
 
 For example, the following interface has 4 items, 3 of which are gated:
 ```wit
@@ -939,6 +940,9 @@ interface foo {
 
   @unstable(feature = fancier-foo)
   d: func();
+
+  @deprecated(version = 0.2.1)
+  e: func();
 }
 ```
 The `@since` gate indicates that `b` and `c` were added as part of the `0.2.1`
@@ -952,6 +956,12 @@ In contrast, the `@unstable` gate on `d` indicates that `d` is part of the
 change type or be removed at any time. An important expectation set by the
 `@unstable` gate is that toolchains will not expose `@unstable` features by
 default unless explicitly opted-into by the developer.
+
+Finally, the `@deprecated` gate on `e` indicates that `e` should no longer be
+used starting version `0.2.1`. `@deprecated` gates can carry an optional
+"message" field which can be used to elaborate why the feature should no longer
+be used and which feature to use instead. Both toolchains and host runtimes may
+warn users if they detect an `@deprecated` API is being used.
 
 Together, these gates support a development flow in which new features start
 with an `@unstable` gate while the details are still being hashed out. Then,
@@ -970,9 +980,15 @@ Specifically, the syntax for feature gates is:
 ```wit
 gate ::= unstable-gate
        | since-gate
+       | deprecated-gate
+
 unstable-gate ::= '@unstable' '(' feature-field ')'
+since-gate ::= '@since' '(' version-field ( ',' feature-field )? ')'
+deprecated-gate ::= '@deprecated' '(' version-field ( ',' message-field )? ')'
+
 feature-field ::= 'feature' '=' id
-since-gate ::= '@since' '(' 'version' '=' <valid semver> ( ',' feature-field )? ')'
+version-field ::= 'version' '=' <valid semver>
+message-field ::= 'message' '=' message
 ```
 
 As part of WIT validation, any item that refers to another gated item must also
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -958,10 +958,10 @@ change type or be removed at any time. An important expectation set by the
 default unless explicitly opted-into by the developer.
 
 Finally, the `@deprecated` gate on `e` indicates that `e` should no longer be
-used starting version `0.2.1`. `@deprecated` gates can carry an optional
-"message" field which can be used to elaborate why the feature should no longer
-be used and which feature to use instead. Both toolchains and host runtimes may
-warn users if they detect an `@deprecated` API is being used.
+used starting version `0.2.1`. Both toolchains and host runtimes may
+warn users if they detect an `@deprecated` API is being used. An `@deprecated`
+gate is required to always be paired up with an `@since` gate. It is
+semantically invalid to use `@deprecated` without also using `@since`.
 
 Together, these gates support a development flow in which new features start
 with an `@unstable` gate while the details are still being hashed out. Then,
@@ -985,7 +985,7 @@ gate-item ::= unstable-gate
 
 unstable-gate ::= '@unstable' '(' feature-field ')'
 since-gate ::= '@since' '(' version-field ( ',' feature-field )? ')'
-deprecated-gate ::= '@deprecated' '(' version-field ( ',' message-field )? ')'
+deprecated-gate ::= '@deprecated' '(' version-field ')'
 
 feature-field ::= 'feature' '=' id
 version-field ::= 'version' '=' <valid semver>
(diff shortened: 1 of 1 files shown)


design/mvp/WIT.md
```diff
@@ -970,9 +980,15 @@ Specifically, the syntax for feature gates is:
...
+       | deprecated-gate
+
 unstable-gate ::= '@unstable' '(' feature-field ')'
+since-gate ::= '@since' '(' version-field ( ',' feature-field )? ')'
+deprecated-gate ::= '@deprecated' '(' version-field ( ',' message-field )? ')'
+
 feature-field ::= 'feature' '=' id
-since-gate ::= '@since' '(' 'version' '=' <valid semver> ( ',' feature-field )? ')'
+version-field ::= 'version' '=' <valid semver>
+message-field ::= 'message' '=' message
```
lukewagner, line 991:
```suggestion
message-field ::= 'message' '=' <core:string>
```
And then right below this could you add a sentence linking `core:string` to https://webassembly.github.io/spec/core/text/values.html#text-string?

lukewagner (approved):
lgtm!  Let me know once there's any implementation to validate this and then we can merge.

design/mvp/WIT.md
```diff
@@ -970,11 +980,20 @@ Specifically, the syntax for feature gates is:
...
+       | deprecated-gate
+
 unstable-gate ::= '@unstable' '(' feature-field ')'
+since-gate ::= '@since' '(' version-field ( ',' feature-field )? ')'
+deprecated-gate ::= '@deprecated' '(' version-field ( ',' message-field )? ')'
+
 feature-field ::= 'feature' '=' id
-since-gate ::= '@since' '(' 'version' '=' <valid semver> ( ',' feature-field )? ')'
+version-field ::= 'version' '=' <valid semver>
+message-field ::= 'message' '=' <core:string>
```
alexcrichton, line 991: One point to note here which is nontrivial in the implementations is that there's no previous grammatical production in WIT which needs a string right now. There is relatively consistently use cases that pop up here and there which aren't necessarily critical but would be nice to have strings, but so far none of them have stuck and right now WIT has no preexisting string syntax.

That means that there's both no existing implementation of this and there's additionally no specification of the string syntax in WIT at this time. We've worked around this in other context (e.g. `version = "1.2.3"` was changed to `version = 1.2.3`), but I don't think it can be worked around easily in this case given how-free form "message" typically is.

yoshuawuyts: Oh that's a good question; I didn't realize that was an issue here. Given that `message` is optional here, perhaps one way to break up this work is to leave it out of an initial design until we've figured out how to handle strings in WIT?

That way we can still introduce the gate sooner, and while we wait for the `message` field to be added, we can include the deprecation message as part of the docs?

alexcrichton: I think that's reasonable yeah 👍

design/mvp/WIT.md
```diff
@@ -970,11 +980,20 @@ Specifically, the syntax for feature gates is:
 ```wit
 gate ::= unstable-gate
        | since-gate
+       | deprecated-gate
```
alexcrichton, line 983: I want to confirm the implications of this grammar which is that previously you're only allowed to have a single `gate` per item. For example you can't have both `@since` and `@unstable`. That makes sense given just those two attributes but most folks typically expect that in languages with attributes you can add multiple attributes. With `@deprecated` it might make sense to have both `@since` and `@deprecated` on the same item.

I say this because the grammar as written doesn't allow having both `@since` and `@deprecated`, but I think it might be reasonable to allow this.

alexcrichton:
I've got an initial implementation of this at https://github.com/bytecodealliance/wasm-tools/pull/1687 with the most notable difference being my prior comment where `@deprecated` is required to be paired with `@unstable` or `@stable`. Easy to change, but figured I'd put a stick in the sand somewhere.

design/mvp/WIT.md
lukewagner, line 983:
```suggestion
gate ::= gate-item*
gate-item ::= unstable-gate
            | since-gate
            | deprecated-gate
```
To wit, this also implicitly fixes a previous "bug" in the spec where we didn't allow an empty `gate`.

yoshuawuyts:
@alexcrichton thanks for putting that PR up; that's great!

Is there a use case when people might want to combine `@unstable` and `@deprecated`? The way I was thinking about this is that in order to mark something as `@deprecated` it must first be marked as stable. And by definition anything marked with `@unstable` is not stable?

(Also I wasn't sure what you meant by "stick in the sand" here - I assumed something like: "starting point" and not like "line in the sand"? I had trouble looking it up 😅)

alexcrichton:
My thinking for `@unstable` and `@deprecated` is that in my experience in Rust if something stays as `@unstable` for too long but it's widely used then it's defacto-stable and a transition plan of sorts is needed for the eventual replacement. IIRC there's been `#[deprecated]` and `#[unstable]` APIs in Rust, and this was intended to mirror that. It's definitely a bit of a niche use case for sure.

> Also I wasn't sure what you meant by "stick in the sand" here

Oh that's just me mixing up idioms and producing nonsense, you can ignore me.

yoshuawuyts:
Okay, great - that's a good enough motivation for me. I think that means that once we've merged https://github.com/bytecodealliance/wasm-tools/pull/1687 we should be ready to merge this too!

design/mvp/WIT.md
```diff
@@ -939,6 +940,9 @@ interface foo {
 
   @unstable(feature = fancier-foo)
   d: func();
+
+  @deprecated(version = 0.2.1)
+  e: func();
```
alexcrichton, line 945: Mind updating this to have `@since` as well?

alexcrichton (approved):

PR #380, opened 2024-07-18
Explicit package declarations should require root package declaration

macovedj:
Recently, support for [explicit package declarations](https://github.com/WebAssembly/component-model/pull/340) was added to wit, enabling multiple package declarations in a single wit file.  This PR proposes that using this new syntax should still require a "root" package declaration.

Today, mixing declarations below isn't yet supported.  This PR proposes that it not only be supported, but required, though just as today, the top level declaration could appear in another file in a wit project directory.

```wit
package root:package;

package first:explicit {
  ...
}

package second:explicit {
  ...
}
```

With this requirement, we can retain the concept of _the_ package that is defined by a wit project, while also being able to define multiple packages in a single project.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -891,9 +891,9 @@ readability but this isn't required.
 Concretely, the structure of a `wit` file is:
 
 ```ebnf
-wit-file ::= package-decl? (package-items | explicit-package-definition)*
+wit-file ::= package-decl? (package-items | nested-package-definition)*
 
-explicit-package-definition ::= package-decl '{' package-items* '}'
+nested-package-definition ::= package-decl '{' package-items* '}'
 
 package-items ::= toplevel-use-item | interface-item | world-item
 ```
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -77,7 +77,7 @@ package local:b {
 }
 ```
 
-It is worth noting that when defining packages "explicitly" with a scope, that it does not remove the need for the "root" package declaration above. That is, there must be a single "root" package that is the container of all such "explicitly" defined packages. 
+It is worth noting that defining nested packages does not remove the need for the "root" package declaration above. These nested package definitions simply provide the contents of other packages inline so that they don't have to be otherwise resolved via the filesystem or a registry. 
 
 Package names are used to generate the [names of imports and exports]
 in the Component Model's representation of [`interface`s][interfaces] and
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -65,7 +65,7 @@ WIT packages can be defined in a collection of files. At least one of these
 files must specify a package name. Multiple files can specify the `package`,
 though they must all agree on what the package name is.
 
-Additionally, many packages can be declared consecutively in one or more files, if the "explicit" package notation is used:
+Additionally, many packages can be declared consecutively in one or more files, if the following nested package notation is used:
 
 ```wit
 package local:a {
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -65,8 +65,7 @@ WIT packages can be defined in a collection of files. At least one of these
 files must specify a package name. Multiple files can specify the `package`,
 though they must all agree on what the package name is.
 
-Alternatively, many packages can be declared consecutively in one or more
-files, if the "explicit" package notation is used:
+Additionally, many packages can be declared consecutively in one or more files, if the "explicit" package notation is used:
 
 ```wit
 package local:a {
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Personally I'm in favor of this change in talking with @macovedj. It's been difficult to rationalize how tooling should be updated when the "thing you just parsed" returns a list of packages when there should probably just be a "main package" somewhere in there. I think that this change returns us back to that original state of having a "main package" at all times but we still have all the benefits of inline curly-braced package blocks for ease of use, reading, single-file storage, etc.

Rolling this change out will be a little weird on the `wasm-tools` side of things but I'm hoping that the usage of multi-package forms hasn't spread too too much.

Also cc @azaslavsky since you worked on the `package .. { ... }` forms as well

design/mvp/WIT.md
```diff
@@ -65,8 +65,7 @@ WIT packages can be defined in a collection of files. At least one of these
 files must specify a package name. Multiple files can specify the `package`,
 though they must all agree on what the package name is.
 
-Alternatively, many packages can be declared consecutively in one or more
-files, if the "explicit" package notation is used:
+Additionally, many packages can be declared consecutively in one or more files, if the "explicit" package notation is used:
```
lukewagner, line 68:
```suggestion
Additionally, many packages can be declared consecutively in one or more files, if the following nested package notation is used:
```

lukewagner: Just thinking of what we want the official terminology to be here, I don't think "explicit" is quite right.  "Nested" seems a bit more intuitive to me.

design/mvp/WIT.md
```diff
@@ -78,6 +77,8 @@ package local:b {
 }
 ```
 
+It is worth noting that when defining packages "explicitly" with a scope, that it does not remove the need for the "root" package declaration above. That is, there must be a single "root" package that is the container of all such "explicitly" defined packages.
```
lukewagner, line 80:
```suggestion
It is worth noting that defining nested packages does not remove the need for the "root" package declaration above. These nested package definitions simply provide the contents of other packages inline so that they don't have to be otherwise resolved via the filesystem or a registry. 
```

design/mvp/WIT.md
```diff
@@ -890,36 +891,15 @@ readability but this isn't required.
...
-
-Files may be organized in two arrangements. The first of these is as a series
-of multiple consecutive "explicit" `package ... {...}` declarations, with the
-package's contents contained within the brackets.
-
-```ebnf
-explicit-package-list ::= explicit-package-definition*
+wit-file ::= package-decl? (package-items | explicit-package-definition)*
 
 explicit-package-definition ::= package-decl '{' package-items* '}'
```
lukewagner, line 896:
```suggestion
wit-file ::= package-decl? (package-items | nested-package-definition)*

nested-package-definition ::= package-decl '{' package-items* '}'
```

lukewagner (approved):
This makes sense to me too, thanks for working through this.

azaslavsky:
I'm generally on board with this change. Is there a companion PR with the implementation in the main wasm-tools repository? The reason I ask is that some tooling for producing WIT, like the pretty printer, might struggle to decide what the "main" package is when printing some set of packages for view. So we'd want to either provide a means to specify that, or else loosen the "must have a root" rule in such cases. Also, today we have users specify their "main" package on the command line when using the all-explicit mode, which should probably be deprecated, lest it collide with this new canonical method of "main" specification.

I suspect there are other similar tricky bits, so it would be good to have a working implementation to work out these corner cases alongside this spec change.

design/mvp/WIT.md
macovedj: Yeah I've noticed that it strikes people as counterintuitive.  I've also been considering "scoped".

macovedj:
> I'm generally on board with this change. Is there a companion PR with the implementation in the main wasm-tools repository?

Been noodling on it.  Hoping I can have something up for you to follow this coming week.

alexcrichton:
My hope is to definitely deprecate/remove the current "set of packages" input where it's just a bunch of `package ... { ... }` blocks in favor of this feature of "must have a top-level `package ...;` to explicitly indicate what the main package is. That'll hopefully simplify the various tooling pieces where everything is still working with a single world/package at a time rather than a set

macovedj:
> Is there a companion PR with the implementation in the main wasm-tools repository?

Still working through some of the tests and tidying up a bit, but figured I'd put it up here so you could get a feel for the changes

alexcrichton:
Ok I believe that the implementation at https://github.com/bytecodealliance/wasm-tools/pull/1700 is good to go now. It's somewhat urgent to get that out because the C# bindings generator was about to start relying on multi-package output in a way that is incompatible with this proposal. I'd like to head that off and avoid any workaround/compat bits by only having in-the-wild C# things use this new format of package-at-the-top and optionally-nested-packages-afterwards

PR #381, opened 2024-07-22
missing `resourcetype` rule in `deftype`

dannypsnl:


lukewagner:
Great catch, thank you!

PR #384, opened 2024-08-01
Add optional fixed length to lists

### LINKED ISSUE #181: Fixed length arrays

cpetig:
Resolves #181

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -12,7 +12,7 @@ rules, but rather merge the minimal need-to-know elements of both, with just
 enough detail to create a prototype. A complete definition of the binary format
 and validation will be present in the [formal specification](../../spec/).
 
-See the [explainer introduction](Explainer.md) for an explanation of 🪙.
+See [Gated Features](Explainer.md#gated-features) for an explanation of 🪙 and 🔧.
 
 
 ## Component Definitions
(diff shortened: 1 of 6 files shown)


jeffparsons:
Question from the peanut gallery: 

Has this change been approved in principle already — i.e. is it just the implementation that's awaiting review?

lukewagner:
Yes, I think I've heard broad agreement on this feature.  But it is a good question of why not just merge the PR now, given that it's emoji-gated (so that it's clear it's not implemented everywhere) and not controversial.  That's probably a good idea, so I'll do that next week unless anyone suggests otherwise.

lukewagner:
Rereading the PR just now, it seems like we'd want a low (but relaxable in the future) upper bound on the fixed length so that bindings generators don't have to think about rare pathological cases (similar to what we did for `flags`).  Maybe 16?

jeffparsons:
I'd argue for at least 64 to support the largest variants of SHA-2, SHA-3 and BLAKE3, which are all 512 bits, as `list<u8, 64>`.

I know this is a very narrow use case, but most of the others that are on my radar are smaller arrays, e.g., IPv6, UUID, coordinates and color values for "graphics stuff", etc.

lukewagner:
Yes, that's a good point.  Given that I'm overdue from merging after my previous comment and this is a minor refinement, I'll merge this PR now as-is and file a new PR to add the length limits in a bit, and we can agree on 64 or whatever else there.

PR #387, opened 2024-08-01
Add explicit scenarios for feature gate usage

### LINKED ISSUE #382: Clarify usage patterns for feature gates
Hey all, after some discussion in the following PRs:

https://github.com/bytecodealliance/wasm-tools/pull/1689
https://github.com/bytecodealliance/jco/pull/459

**It seems like now would be a good time to discuss and really pin down the expected usage pattern of feature gates.** 

Intuitions seem to clash on how the gates should be used, so the decision should likely be discussed and made (hopefully quickly, if possible!) with a record for people to see the whys.

## Context

At present, the [WIT design feature gates](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#feature-gates) section describes a reality that requires/implies a few things:

**1. Package versions may not be present**
There is no mention that a package containing `@since()` that is not itself versioned is a bug.

There are currently some tests in `wasm-tools` that use `@since` in packages that have no version specified (i.e. this is expected to be "right" or at least legal).

**2. `@since(version = v, feature = f)` works as an OR**

> Then, once the feature is stable (and, in a WASI context, voted upon), the @unstable gate is switched to a @since gate. To enable a smooth transition (during which producer toolchains are targeting a version earlier than the @since-specified version), the @since gate contains an optional feature field that, when present, says to enable the feature when either the target version is greator-or-equal or the feature name is explicitly enabled by the developer.

This necessarily means that version is used *or* the feature is used.

**3. `@since()` may specify versions in the future

> Thus, c is enabled if the version is 0.2.2 or newer or the fancy-foo feature is explicitly enabled by the developer.

The first part of this "if" can only work if you were able to specify a future (unreleased) version. It doesn't make sense to describe it like this unless a situation like this is possible:

```wit
// there's obviously quite a gap in these versions but hopefully it's useful for the sake of discussion.
package some:pkg@0.1.0;

@since(version = 0.2.2, feature = access-early-a)
interface a { ... }
```

## Options for moving forward

There *seem* to be at least these three ways of handling it:

**1. OR (current)**
In this situation, we keep the functionality as it exists now. This requires the caveats mentioned above.

One of the questions associated with this approach is whether it makes sense for a package at version `0.1.0` to be able to have a `@since(version = 0.2.0)` in it *at all*.

**2. AND (new)**

In this situation `version` and `feature` are always *both* considered for enabling an item.

This has a few benefits I can think of:
- Always knowing an import requires *all* the `@since` related options makes it easy to do things like infer requirements from an import (versus not knowing if that means you need a version or a feature)
- It's possible to write features that actually change across versions, and versions that act differently w/ certain features enabled

**3. Strict `@unstable` -> `@since` with ignored features after stabilization**

Alex has laid out [his intuition for this](https://github.com/bytecodealliance/wasm-tools/pull/1689#issuecomment-2248527698), which makes also makes sense.

Strictly moving from `@unstable` to `@since` and then ignoring features once enabled, along with making sure that future unreleased versions cannot be used in `@since` removes a lot of ambiguity.

## What needs to be clarified

It would be nice if we could lay out the actual intended evolution of a feature through gates -- this is kind of already spelled out already in the WIT design doc, but there seems to be two levels of progression here:

- unstable -> stable (experimental features)
- not enabled version  -> enabled version (future feature prep)

I think we're done at the point when we have:
- A flow (probably best written out by steps) that people agree with that most importantly seems *intuitive* for users to follow (`@unstable` -> `@since`, `@since(version = v)` -> `@since(version = v + 1)` ?)
- *Which* levels of progression the feature gates feature should be handling

vados-cosmonic:
This commit addds some explanation of the intended usage pattern of feature gates in order to make the transition points and functionality easier to identify/reason about.

Resolves #382

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -643,7 +643,7 @@ interface a {
   resource r;
 }
 interface b {
-  use a.{r}; 
+  use a.{r};
   foo: func() -> r;
 }
 
@@ -946,17 +946,17 @@ required to always be paired up with either a `@since` or `@deprecated` gate.
 Together, these gates support a development flow in which new features start
 with an `@unstable` gate while the details are still being hashed out. Then,
 once the feature is stable (and, in a WASI context, voted upon), the
-`@unstable` gate is switched to a `@since` gate. To enable a smooth transition
-(during which producer toolchains are targeting a version earlier than the
-`@since`-specified `version`), the `@since` gate contains an optional `feature`
-field that, when present, says to enable the feature when *either* the target
-version is greator-or-equal *or* the feature name is explicitly enabled by the
-developer. Thus, `c` is enabled if the version is `0.2.2` or newer or the
+`@unstable` gate is switched to a `@since` gate.
+
+Thus, `c` is enabled if the version is `0.2.2` or newer or the
 `fancy-foo` feature is explicitly enabled by the developer. The `feature` field
 can be removed once producer toolchains have updated their default version to
 enable the feature by default.
 
-Specifically, the syntax for feature gates is:
+#### Feature gate syntax
+
+The grammar that governs feature gate syntax is:
+
 ```wit
 gate ::= gate-item*
 gate-item ::= unstable-gate
@@ -964,15 +964,18 @@ gate-item ::= unstable-gate
             | deprecated-gate
 
 unstable-gate ::= '@unstable' '(' feature-field ')'
-since-gate ::= '@since' '(' version-field ( ',' feature-field )? ')'
+since-gate ::= '@since' '(' version-field ')'
 deprecated-gate ::= '@deprecated' '(' version-field ')'
 
 feature-field ::= 'feature' '=' id
 version-field ::= 'version' '=' <valid semver>
 ```
 
+#### Rules for feature gate usage
+
 As part of WIT validation, any item that refers to another gated item must also
 be compatibly gated. For example, this is an error:
+
 ```wit
 interface i {
   @since(version = 1.0.1)
@@ -981,6 +984,7 @@ interface i {
   type t2 = t1; // error
 }
 ```
+
 Additionally, if an item is *contained* by a gated item, it must also be
 compatibly gated. For example, this is an error:
 ```wit
@@ -992,6 +996,157 @@ interface i {
   bar: func();  // also error: weaker gate
 }
 ```
+
+The following rules apply to the use of feature gates:
+
+- Either `@since` *or* `@unstable` should be used, but not both (exclusive or).
+- If a package contains a feature gate, it's version must be specified (i.e. `namespace:package@x.y.z`)
+
+#### Scenario: Stabilization of a new feature
+
+This section lays out the basic flow and expected usage of feature gate machinery
+ when stabilizing new features and deprecating old ones.
+
+Assume the following WIT package as the initial interface:
+
+```wit
+package examples:fgates-calc@0.1.0;
+
+@since(version = 0.1.0)
+interface calc {
+  @since(version = 0.1.0)
+  variant calc-error {
+    integer-overflow,
+    integer-underflow,
+    unexpected,
+  }
+
+  @since(version = 0.1.0)
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+**First, add new items under an `@unstable` annotation with a `feature` specified:**
+
+```wit
+package examples:fgates-calc@0.1.1;
+
+@since(version = 0.1.0)
+interface calc {
+  @since(version = 0.1.0)
+  variant calc-error {
+    integer-overflow,
+    integer-underflow,
+    unexpected,
+  }
+
+  @since(version = 0.1.0)
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+
+  /// By convention, feature flags should be prefixed with package name to reduce chance of collisions
+  ///
+  /// see: https://github.com/WebAssembly/WASI/blob/main/Contributing.md#filing-changes-to-existing-phase-3-proposals
+  @unstable(feature = fgates-calc-minus)
+  sub: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+At this point, consumers of the WIT can enable feature `fgates-calc-minus` through their relevant tooling and get access to the `sub` function.
+
+Note that, at least until subtyping is relaxed in the Component Model, if we had to *add* a new case to `calc-error`, this would be a *breaking change* and require either a new major version or adding a second, distinct `variant` definition used by new functions.
+
+**Second, when the feature is ready to be stabilized, switch to a `@since` annotation:**
+
+```wit
+package examples:fgates-calc@0.1.2;
+
+@since(version = 0.1.0)
+interface calc {
+  @since(version = 0.1.0)
+  variant calc-error {
+    integer-overflow,
+    integer-underflow,
+    unexpected,
+  }
+
+  @since(version = 0.1.0)
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+
+  @since(version = 0.1.2)
+  sub: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+#### Scenario: Deprecation of an existing stable feature
+
+This section lays out the basic flow and expected usage of feature gate machinery when stabilizing a new feature.
+
+Assume the following WIT package as the initial interface:
+
+```wit
+package examples:fgates-deprecation@0.1.1;
+
+@since(version = 0.1.0)
+interface calc {
+  @since(version = 0.1.0)
+  variant calc-error {
+    integer-overflow,
+    integer-underflow,
+    unexpected,
+  }
+
+  @since(version = 0.1.0)
+  add-one: func(x: i32) -> result<i32, calc-error>;
+
+  @since(version = 0.1.1)
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+**First: Add the `@deprecated` annotation to the relevant item in a new version**
+
+```wit
+package examples:fgates-deprecation@0.1.2;
+
+@since(version = 0.1.0)
+interface calc {
+  @since(version = 0.1.0)
+  variant calc-error {
+    integer-overflow,
+    integer-underflow,
+    unexpected,
+  }
+
+  @deprecated(version = 0.1.2)
+  add-one: func(x: i32) -> result<i32, calc-error>;
+
+  @since(version = 0.1.1)
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+At this point, tooling consuming this WIT will be able to appropriately alert users to the now-deprecated `add-one` function.
+
+**Second: completely remove the deprecated item in some future SemVer-compliant major version**
+
+```wit
+package examples:fgates-deprecation@0.2.0;
+
+@since(version = 0.1.0)
+interface calc {
+  @since(version = 0.1.0)
+  variant calc-error {
+    integer-overflow,
+    integer-underflow,
+    unexpected,
+  }
+
+  @since(version = 0.1.1)
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+In this new "major" version (this is considered a major version under SemVer 0.X rules) -- the `add-one` function can be fully removed.
 
 ## Package declaration
 [package declaration]: #package-declaration


design/mvp/WIT.md
```diff
@@ -1013,6 +1019,116 @@ interface i {
...
+  @deprecated(version = 0.1.2)
+  add-one: func(x: i32) -> result<i32, calc-error>;
+
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+At this point, tooling consuming this WIT will be able to appropriately alert users to the now-deprecated `add-one` function.
+
+**Second: cmpletely remove the deprecated item in the next SemVer-compliant major version**
```
calvinrp, line 1118: Minor typo "cmpletely"

vados-cosmonic: Ah thank you -- fixed!

design/mvp/WIT.md
```diff
@@ -1013,6 +1019,116 @@ interface i {
...
+
+```wit
+package examples:fgates@0.1.1;
+
+// ... elided ...
+
+interface calc {
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+
+  @unstable(feature = minus)
```
ricochet, line 1052: For wasi, we established a convention to use the package as a prefix for features in order to avoid conflicts. This is not required by the spec, but it might be helpful to include a sentence that this is a good practice.

https://github.com/WebAssembly/WASI/blob/main/Contributing.md#filing-changes-to-existing-phase-3-proposals

vados-cosmonic: Ah thanks! will include this:

- [ ] Add package name as prefix to feature in example,

design/mvp/WIT.md
```diff
@@ -1013,6 +1019,119 @@ interface i {
...
+#### Scenario: Stabilization of a new feature
+
+This section lays out the basic flow and expected usage of feature gate machinery when stabilizing a new feature.
+
+Assume the following WIT package as the initial interface:
+
+```wit
+package examples:fgates@0.1.0;
+
+variant calc-error {
```
alexcrichton, line 1031: Could this be moved within the `interface` below to be valid WIT syntax?

design/mvp/WIT.md
```diff
@@ -1013,6 +1019,119 @@ interface i {
...
+
+  /// By convention, feature flags should be prefixed with package name to reduce chance of collisions
+  ///
+  /// see: https://github.com/WebAssembly/WASI/blob/main/Contributing.md#filing-changes-to-existing-phase-3-proposals
+  @unstable(feature = fgates-calc-minus)
+  sub: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+At this point, consumers of the WIT can enable feature `minus` through their relevant tooling and get access to the `sub` function.
```
alexcrichton, line 1060: s/minutes/fgates-calc-minus/

alexcrichton (approved):

design/mvp/WIT.md
vados-cosmonic: Ah absolutely, thanks for this

- [ ] fix invalid `examples:fgates` WIT

design/mvp/WIT.md
vados-cosmonic: Thanks good catch

`fix reference to minutes w/ `fgates-calc-minus`

design/mvp/WIT.md
```diff
@@ -1013,6 +1019,119 @@ interface i {
...
 ```
 
+#### Scenario: Stabilization of a new feature
+
+This section lays out the basic flow and expected usage of feature gate machinery when stabilizing a new feature.
+
+Assume the following WIT package as the initial interface:
+
+```wit
+package examples:fgates@0.1.0;
```
lukewagner, line 1029:
```suggestion
package examples:fgates-calc@0.1.0;
```

design/mvp/WIT.md
```diff
@@ -1013,6 +1019,119 @@ interface i {
...
+  ///
+  /// see: https://github.com/WebAssembly/WASI/blob/main/Contributing.md#filing-changes-to-existing-phase-3-proposals
+  @unstable(feature = fgates-calc-minus)
+  sub: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+At this point, consumers of the WIT can enable feature `fgates-calc-minus` through their relevant tooling and get access to the `sub` function.
+
+Note that if we had to *add* a new variant to `calc-error`, this would be a *breaking change* for most if not all downstream bindgen tooling (i.e. the enum would become a new object, incompatible with previous code).
```
lukewagner, line 1062:
```suggestion
Note that, at least until subtyping is relaxed in the Component Model, if we had to *add* a new case to `calc-error`, this would be a *breaking change* and require either a new major version or adding a second, distinct `variant` definition used by new functions.
```

design/mvp/WIT.md
```diff
@@ -1013,6 +1019,119 @@ interface i {
...
+
+interface calc {
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+
+  @since(version = 0.1.2)
+  sub: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+While you *may* include the `feature` portion of `@since` (i.e. `@since(version = 0.1.2, feature = fgates-calc-minus)`), it is *ignored* by tooling and is only informational once the feature is stabilized.
```
lukewagner, line 1079: To simply WIT authors' lives and save them from having to ask "should I keep the feature-gate name around... and for how long?", should we just remove the optional `feature` field?  If we can't identify a scenario where we recommend keeping it, then I would think yes.

design/mvp/WIT.md
```diff
@@ -1013,6 +1019,119 @@ interface i {
...
+package examples:fgates-deprecation@0.1.1;
+
+interface calc {
+  variant calc-error {
+    integer-overflow,
+    integer-underflow,
+    unexpected,
+  }
+
+  add-one: func(x: i32) -> result<i32, calc-error>;
```
lukewagner, line 1097: Maybe add a `@since(version = 0.1.0)` to `add-one` to make the example more representative of WASI?

lukewagner (commented):
Looks great, thanks for doing this!  Just one question below which, depending on what folks here think, suggests a few more changes in the PR.

design/mvp/WIT.md
vados-cosmonic: Hmnn, here I was hoping to convey that the `add-one` function was actually there at "inception" -- `fgates-deprecation@0.1.0`, and I'd imagine that most people wouldn't add `@since` gates for the very first version. 

I can add some more prose to make this clearer -- something like:

> For example, consider the following scenario (illustrated as WIT below):
>   - The interface `calc` was created with only one function (`add-one`), in package version `0.1.0`
>   - A more generic `add` was added in package version 0.1.1
>   - We'd like to deprecate `add-one` as the more generic version of `add` is more flexible

What do you think?

design/mvp/WIT.md
vados-cosmonic: This I'd definitely be in favor of (I can't think of a real use case for the feature field) -- would this be worth making another issue as space to get feedback for the removal as a follow on to the work done so far?

I'm not quite sure what the original intent was for the optional feature to be added (historical context?), but a new issue might be a easier-to-reference future place for the "why"/intent

design/mvp/WIT.md
lukewagner: At least currently in WASI, we seem to be adding `@since` to everything, including functions present from inception.

design/mvp/WIT.md
lukewagner: IIUC, this was the resolution of our previous discussion in #382 ([this comment](https://github.com/WebAssembly/component-model/issues/382#issuecomment-2251236972) was my attempt to summarize), so I think this has already been discussed and tentatively concluded.  Since this PR is intended to resolve #382 it could make sense here.

design/mvp/WIT.md
vados-cosmonic: Thanks for pointing that out -- I'll make this consistent with the other WASI WITs!

vados-cosmonic: Just added this here: https://github.com/WebAssembly/component-model/pull/387/commits/ef754f097fd58d8adf06d8265bb0714c97a3d640

design/mvp/WIT.md
vados-cosmonic: Ah yes, thanks for reminding me -- I'll remove them here to match (and the code changes hopefully follow soon after for stuff like `wasm-tools`, etc)

vados-cosmonic: OK -- I've removed the mention of the `feature` portion of `@since` for now! I considered explicitly discouraging people from using it, but thought having the example just do the right thing should be fine.

design/mvp/WIT.md
```diff
@@ -946,17 +946,21 @@ required to always be paired up with either a `@since` or `@deprecated` gate.
...
-field that, when present, says to enable the feature when *either* the target
-version is greator-or-equal *or* the feature name is explicitly enabled by the
-developer. Thus, `c` is enabled if the version is `0.2.2` or newer or the
+`@unstable` gate is switched to a `@since` gate.
+
+To enable a smooth transition (during which producer toolchains are targeting a
+version earlier than the `@since`-specified `version`), the `@since` gate contains
+an optional `feature` field that, when present, says to enable the feature when
+*either* the target version is greater-or-equal *or* the feature name is explicitly
+enabled by the developer.
```
lukewagner, line 955: Would it make sense to delete this paragraph and also the `feature` field of `@since` from the grammar as well?

vados-cosmonic: Ah thanks for pointing this out, I should definitely add that in here -- just made the update!

design/mvp/WIT.md
```diff
@@ -993,6 +993,151 @@ interface i {
...
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+
+  @since(version = 0.1.2)
+  sub: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+#### Scenario: Deprecation of an existing stable feature
+
+This section lays out the basic flow and expected usage of feature gate machinery when stabilizing a new feature.
```
lukewagner, line 1072:
```suggestion
This section lays out the basic flow and expected usage of feature gate machinery when deprecating a new feature.
```

design/mvp/WIT.md
```diff
@@ -993,6 +993,151 @@ interface i {
...
+  add-one: func(x: i32) -> result<i32, calc-error>;
+
+  @since(version = 0.1.1)
+  add: func(x: i32, y: i32) -> result<i32, calc-error>;
+}
+```
+
+At this point, tooling consuming this WIT will be able to appropriately alert users to the now-deprecated `add-one` function.
+
+**Second: completely remove the deprecated item in the next SemVer-compliant major version**
```
lukewagner, line 1120:
```suggestion
**Second: completely remove the deprecated item in some future SemVer-compliant major version**
```

lukewagner: (Just because the "next" major version may be too soon, depending on timing/community.)

lukewagner (approved):
Great, thanks!

lukewagner:
I just realized that one other change we discussed in #382 that maybe we want to make in this PR (if we want it to fully resolve #382) is the validation requirements (attached to the various gates) of: `@since` xor `@unstable` and gates-imply-top-level-`package` (just as an extra sentence or two in the section that introduces the gates).

design/mvp/WIT.md
vados-cosmonic: Is this right? I wouldn't normally think of "deprecation" in the case of a new feature -- maybe I'll change this to "stabilizing new features and deprecating old ones".

vados-cosmonic:
@lukewagner thanks that's a great catch -- that little requirement was definitely an eye opener on the code side -- will add!

vados-cosmonic:
Kept this particular change separate so it might be easy to review (I added some headings):

https://github.com/WebAssembly/component-model/pull/387/commits/af0c60ef819c1714fd4f3c63354cb934a82dae61

Would appreciate just a bit more feedback!

lukewagner:
Awesome, lgtm, thanks again!  Once things are implemented enough to feel good about this PR, lmk and I'll merge this.

vados-cosmonic:
Yup! I'll squash here and see what I can do about updating implementation in other spots!

vados-cosmonic:
Hey @lukewagner so at this point the optional feature is removed (in a semi-breaking manner, those changes should percolate through but have rarely few people affected), anything else I'm missing? We don't have a new version of `wit-parser` and/or `wit-component` out yet of course, but it *should* be relatively real now going forward.

lukewagner:
Great, thanks again for all the work!

PR #393, opened 2024-09-05
Unlocked deps

macovedj:
This PR servers as an initial description of what the desired behavior of an `unlocked-dep` would include.  As exemplified in the PR, most language toolchains would probably use it in combination with [nested interfaces](https://github.com/WebAssembly/component-model/pull/372), which is currently in the process of being implemented.

The primary motivation behind this syntax is enabling a way for registry tooling concerned with package resolution to resolve implementations of components (rather than wit).  This should hopefully enable workflows where users can depend on specific components, rather than interfaces whose implementations need to be specified at a later time.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -663,7 +663,7 @@ world w2 {
 
 ## Unlocked Dependency Imports
 
-When working with a registry, the keyword `unlocked-dep` is available to specify version requirements as ranges.
+When working with a registry, the keyword `unlocked-dep` is available to specify dependencies with package name and version requirements.  For example:
 
 ```wit
 world w {
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -661,6 +661,60 @@ world w2 {
 > configure that a `use`'d interface is a particular import or a particular
 > export.
 
+## Unlocked Imports (semver)
+
+When working with a registry, the keyword `unlocked-dep` is available to specify version requirements as ranges.
+
+```wit
+world w {
+  unlocked-dep foo:bar@{>=x.x.x <y.y.y};
+}
+```
+
+The binary format has a corresponding [import definition](Explainer.md#import-and-export-definitions) and this WIT syntax informs
+bindgen tooling that it should be used.
+
+The key idea here is to be able to specify a dependency on a _component_, rather than on a wit interface.  Sometimes, as a component author, the goal is to have a dynamic import, where at a time after development is done, one of many implementations of a wit interface is specified, so that in essence your dependency makes your component configurable.  This workflow is well documented across a variety of tools.  Unlocked imports, on the other hand, are available for
+specifying a dependency on a specific implementation of an interface, with a semver range
+
+### Example Unlocked Workflow
+Each language has its own toolchain for creating wasm components that should feel familiar to users of that language.  As an example, somebody authoring a rust component would add the component they're interested in to their `Cargo.toml`.
+
+```
+"foo:bar" = "x.x.x"
+```
+
+Say that the exports of this component match the exports of some `world exports`.  Then the wit used for the component being authored would end up as follows:
+
+```wit
+package my:component
+
+package foo:bar {
+  interface exports {
+    nest some:other/interfacename
+    ...
+  }
+}
+
+world w {
+  unlocked-dep foo:bar/exports@{>=x.x.x <y.y.y}
+}
+```
+
+
+Once this wit is synthesized by the language toolchain based on the language's package file, bindings can be generated and a wasm binary can be compiled which will contain unlocked imports as defined in the [import definition section](Explainer.md#import-and-export-definitions) of the explainer.  Below is an example of the unlocked imports that will be present in the `wat`
+
+```wat
+(import "unlocked-dep=<foo:bar/exports@{>=1.0.0}>")
+```
+
+A wasm component binary that has unlocked imports is referred to as an unlocked component. In general, unlocked components are what would be published to a registry.   Registry aware tools have a command `lock` that will produce a  "locked" component when it resolves dependency versions.  As an example, today the warg cli has a [lock command](https://github.com/bytecodealliance/registry/blob/main/src/commands/lock.rs). This "locked"  component will use locked import statements, rather than unlocked import statements, with pinned version numbers and integrity hashes discovered during package resolution, as can be seen below.
+
+```wat
+(import "locked-dep=<foo:bar/example@1.0.0>,integrity=<sha256-7b582e13fd1f798ed86206850112fe01f837fcbf3210ce29eba8eb087e202f62>")
+```
+
+Locked components aren't runnable unless they are being run by a registry aware runtime.  They serve the role of a reproducible deployment artifact.  In order to run them with a runtime that is not registry aware, one would need to use a `bundle` command, also made available by registry aware toolchains, that will inline component definitions where locked imports exist in a locked component.  The warg cli also has a reference implementation of a [bundle command](https://github.com/bytecodealliance/registry/blob/main/src/commands/bundle.rs).
 ## WIT Functions
 [functions]: #wit-functions
 
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -676,44 +676,65 @@ bindgen tooling that it should be used.
 
 The point of `unlocked-dep` is to specify a dependency on a _component implementation_ (or a semver *range* of implementations), rather than on an abstract WIT interface (with an unspecified implementation).
 
-### Example Unlocked Workflow
-Each language has its own toolchain for creating wasm components that should feel familiar to users of that language.  As an example, somebody authoring a rust component would add the component they're interested in to their `Cargo.toml`.
+### Example Unlocked Dependency Workflow
+Let's say someone authoring a Rust component `my:component` targeting the `wasi:http/proxy` world adds a dependency on another component `foo:bar` by adding the following (hypothetical) lines to their `Cargo.toml`:
 
 ```
-"foo:bar" = "x.x.x"
-```
+[package.metadata.component.target]
+target = "wasi:http/proxy@0.2.0"
 
-Say that the exports of this component match the exports of some `world exports`.  Then the wit used for the component being authored would end up as follows:
-
+[package.metadata.component.dependencies]
+"foo:bar" = "1.2"
+The language toolchain would first generate a new world that augments the target world with an additional unlocked dependency:   
 ```wit
-package my:component
+package my:component;
+world generated-world {
+  include wasi:http/proxy@0.2.0;
+  unlocked-dep foo:bar@{>=1.2.0};
+}
+```
 
-package foo:bar {
+Now say that `foo:bar@1.2.0` implements the following `exports` interface:
+```wit
+package foo:bar@1.2.0;
+interface exports {
+  calc: func(x: u64) -> u64;
+}
+```
+
+The next step is to expand `generated-world` with nested packages fetched from all the relevant registries, producing an all-in-one WIT file with no external references:  
+```wit
+package my:component;
+
+package wasi:http@0.2.0 {
+  ...
+  world proxy { ... }
+}
+
+package foo:bar@1.2.0 {
   interface exports {
-    nest some:other/interfacename
-    ...
+    calc: func(x: u64) -> u64;
   }
 }
 
-world w {
-  unlocked-dep foo:bar/exports@{>=x.x.x <y.y.y}
+world generated-world {
+  include wasi:http/proxy@0.2.0;
+  unlocked-dep foo:bar/exports@{>=1.2.0};
 }
 ```
-
-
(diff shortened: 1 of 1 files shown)


design/mvp/WIT.md
```diff
@@ -661,6 +661,60 @@ world w2 {
...
+
+```wat
+(import "locked-dep=<foo:bar/example@1.0.0>,integrity=<sha256-7b582e13fd1f798ed86206850112fe01f837fcbf3210ce29eba8eb087e202f62>")
+```
+
+Locked components aren't runnable unless they are being run by a registry aware runtime.  They serve the role of a reproducible deployment artifact.  In order to run them with a runtime that is not registry aware, one would need to use a `bundle` command, also made available by registry aware toolchains, that will inline component definitions where locked imports exist in a locked component.  The warg cli also has a reference implementation of a [bundle command](https://github.com/bytecodealliance/registry/blob/main/src/commands/bundle.rs).
```
lukewagner, line 717:
```suggestion
A wasm component that contains `unlocked-dep` imports is referred to as an "unlocked component".  Unlocked components are what you normally would want to publish to a registry, since it allows users of the unlocked component to perform the final dependency solving across a DAG of components.
```
Until we have some CLI command documentation (in particular, for subcommands of `wkg` that do `lock` and `bundle`), maybe it's best to just leave it here, and we can always add more later.  Importantly though, I think this traces the path of `unlocked-dep` from WIT into WAT, which I think is all we need for this section.

design/mvp/WIT.md
```diff
@@ -661,6 +661,60 @@ world w2 {
 > configure that a `use`'d interface is a particular import or a particular
 > export.
 
+## Unlocked Imports (semver)
+
+When working with a registry, the keyword `unlocked-dep` is available to specify version requirements as ranges.
```
lukewagner, line 666:
```suggestion
When working with a registry, the keyword `unlocked-dep` is available to specify and package name and version requirements.  For example:
```

design/mvp/WIT.md
```diff
@@ -661,6 +661,60 @@ world w2 {
...
+  interface exports {
+    nest some:other/interfacename
+    ...
+  }
+}
+
+world w {
+  unlocked-dep foo:bar/exports@{>=x.x.x <y.y.y}
+}
+```
```
lukewagner, line 702:
````suggestion
Now say that `foo:bar@1.2.0` implements the following `exports` interface:
```wit
package foo:bar@1.2.0;
interface exports {
  calc: func(x: u64) -> u64;
}
```
````
I'm thinking it's useful to not also introduce `nested` here, when trying to show the most basic example.

File: design/mvp/WIT.md [hunk 4]
```diff
@@ -661,6 +661,60 @@ world w2 {
...
+
+
+Once this wit is synthesized by the language toolchain based on the language's package file, bindings can be generated and a wasm binary can be compiled which will contain unlocked imports as defined in the [import definition section](Explainer.md#import-and-export-definitions) of the explainer.  Below is an example of the unlocked imports that will be present in the `wat`
+
+```wat
+(import "unlocked-dep=<foo:bar/exports@{>=1.0.0}>")
+```
```
lukewagner, line 709:
````suggestion
The next step is to expand `generated-world` with nested packages fetched from all the relevant registries, producing an all-in-one WIT file with no external references:  
```wit
package my:component;

package wasi:http@0.2.0 {
  ...
  world proxy { ... }
}

package foo:bar@1.2.0 {
  interface exports {
    calc: func(x: u64) -> u64;
  }
}

world generated-world {
  include wasi:http/proxy@0.2.0;
  unlocked-dep foo:bar/exports@{>=1.2.0};
}
```
Note that `exports` is an arbitrary name and can be anything because the WIT bindings generation will always strip off the final interface name, leaving only the package name.  In particular, the Component Model type for this world is:
```wat
(component
  (import "wasi:http/types@0.2.0" (instance ...))
  (import "wasi:http/outgoing-handler@0.2.0" (instance ...))
  (import "unlocked-dep=<foo:bar@{>=1.2.0}>" (instance
    (export "calc" (func (param "x" u64) (result u64)))
  ))
  (export "wasi:http/incoming-handler@0.2.0" (instance ...))
)
```
````
Note that `exports` is stripped by the time we get into the component.

**Review by lukewagner — COMMENTED** [review]
Thanks for starting this!  I have some suggested changes to the examples, WDYT:

**Comment by macovedj** [comment]
Appreciate the review!  Makes sense to me.

File: design/mvp/WIT.md [hunk 5]
```diff
@@ -661,6 +661,80 @@ world w2 {
 > configure that a `use`'d interface is a particular import or a particular
 > export.
 
+## Unlocked Dependency Imports
+
+When working with a registry, the keyword `unlocked-dep` is available to specify dependencies with package name and version requirements.  For example:
+
+```wit
+world w {
+  unlocked-dep foo:bar@{>=x.x.x <y.y.y};
```
alexcrichton, line 670: Is it intended to support `foo:bar` here exactly? Or is a `/interface` required as well?

lukewagner: I was thinking that, when `/interface` is absent, the expectation could be that a registry is used to resolve `ns:package` to a component and the interface is the exports of that component (which have no name).

More hypothetically, if we get unnamed interfaces/worlds in WIT, `/interface` could also be absent without relying on a registry by using a nested package of the form:
``wit
package foo:bar {
  interface { ... }
}
world w {
  unlocked-dep foo:bar@{...}
}
```
WDYT?

Comment (macovedj) [reply design/mvp/WIT.md:670]: But perhaps this is a good opportunity to make clear that the optional `/interface` is possible as well, and maybe talk a bit about in which cases toolchains would create an `unlocked-dep` with and without an interface projection?  If we wanted to scope it in, folks could specify an interface when they add a dep and tools could only grab interfaces they need from the registry.  And also the `lock` or `bundle` command could shake out unused interfaces in cases as well.  May be more than we want for the initial discussion, but I _was_ thinking that folks may find it confusing to only see a component `foo:bar` when they're accustomed to depending on interfaces at the moment.

Comment (alexcrichton) [reply design/mvp/WIT.md:670]: Could this spec out the `interface { ... }` change as well in that case? That's not currently implemented or sketched out here, so I think that should be included too if that's the intention. (also could this update the ebnf for `unlocked-dep` in worlds too?)

Comment (lukewagner) [reply design/mvp/WIT.md:670]: I'd be up for that; it would simplify things.

Comment (macovedj) [reply design/mvp/WIT.md:670]: Great I should be able to add some of these updates soon.

File: design/mvp/WIT.md [hunk 6]
```diff
@@ -661,6 +661,80 @@ world w2 {
...
+  (import "wasi:http/types@0.2.0" (instance ...))
+  (import "wasi:http/outgoing-handler@0.2.0" (instance ...))
+  (import "unlocked-dep=<foo:bar@{>=1.2.0}>" (instance
+    (export "calc" (func (param "x" u64) (result u64)))
+  ))
+  (export "wasi:http/incoming-handler@0.2.0" (instance ...))
+)
+```
+
+A wasm component that contains `unlocked-dep` imports is referred to as an "unlocked component".  Unlocked components are what you normally would want to publish to a registry, since it allows users of the unlocked component to perform the final dependency solving across a DAG of components.
```
James-Mart, line 737: Related to my comment on this PR, but an "unlocked component" really doesn't seem self-explanatory to me. Language like, "coupled component" or "tightly coupled component" seem to more readily express the fact that the component depends on implementations rather than interfaces.

macovedj:
I think this is a great place for you to voice such an opinion @James-Mart, really appreciate the feedback!

>Seems to me that if the WIT author is concerned with expressing the simple concept of a dependency on a specific component vs on any component that conforms to the specified interface, then perhaps wit could have more self->explanatory keywords like specified-dep and specified-dep-range?

Something that may impact your opinion... At the moment we don't really anticipate that developers would ever author this syntax in WIT themselves, or at least not often.  More commonly, toolchains would deduce that this WIT should be used for bindings generation based on a dependency specification in a package file.

As for the syntax suggestions, I don't feel super strongly one way or the other immediately.  It's valuable to hear that some of the verbiage/syntax doesn't feel intuitive to you, but I also think we have yet to fully document the entire lifecycle of an unlocked dep/unlocked component.  I was having a bit of difficulty deciding how much of this information belongs in this PR vs in other places, but I'll try and add some context that helps provide some clarity around some of the points you're raising.

lukewagner:
@James-Mart Even if mostly folks won't manually write their component dependencies in WIT often, I'm definitely happy to discuss what is the most intuitive syntax/naming that we can find for WIT (which I agree can be distinct from what goes into the raw WAT) since I think folks will at least occasionally *read* synthesized WITs that use this syntax.  So thanks for bringing this up!

One observation is that both `unlocked-dep` and your proposed `specified-dep`/`speified-dep-range` share the (abbreviation of the) word "dependency".  What if we used just the word "dependency" and inferred "locked" vs. "unlocked"/"range" from the syntax after the `@`.  E.g.:
```wit
world w {
  dependency ns:foo@{>=1.0.0}; // => unlocked-dep=<ns:foo@{>=1.0.0}>
  dependency ns:foo@1.0.0;     // => locked-dep=<ns:foo@1.0.0>
}
```
(We'd also need to invent some syntax for where to include the `integrity=` content hash for a `locked-dep`.)

Separately, I've wondered whether there's value in having an explicit `import` prefix:
```wit
world w {
  import dependency ns:foo@{>=1.0.0};
```
Syntactically, there's no need for `import` (dependencies can only be imported, not exported), but I wonder if there's value in having it for regularity dependencies are in fact imported?  Happy to hear other opinions on this.

As a last detail, I think that `dependency`s would *always* emits an `import` with an `instance` type.  Eventually (I don't think we need to add or choose it in this PR), we'll also need a syntax that emits a `component` type (which *tends* to show up when you lock your dependencies, but, at the component-model level, is orthogonal).  Observing that a `component`-typed import allows you to create private child instances (not shared with any other component) whereas an `instance`-typed import implies/expects that the same imported `instance` can be shared with other components, maybe `private dependency` could imply `component` type?  Whatever the precise syntax, this is the rare case which I expect folks won't see often (even in synthesized WIT), since `component` imports mostly only show up at the end of building an application (out of unlocked components).

James-Mart:
@macovedj 
> At the moment we don't really anticipate that developers would ever author this syntax in WIT themselves, or at least not often.

Ah, that does make sense for general use. Although I _think_ in the context of my use-case for component development this may be the mechanism that developers are instructed to use pretty regularly, since they are building components that are always dynamically composed, and there can be multiple implementations of each interface, so developers need some way to indicate to the runtime which instance should be linked.

@lukewagner 
> What if we used just the word "dependency" 

What about `instance`? Seems potentially better than `dependency` at conveying the idea that it's specifically a dependency on an instance, rather than an interface.

> and inferred "locked" vs. "unlocked"/"range" from the syntax after the @

Definitely +1 for this

> I wonder if there's value in having [`import`] for regularity dependencies are in fact imported?

Without the explicit import, I think a reader is more likely to assume that `dependency` is some kind of top-level directive (like `import` and `export`), rather than a specialization of import. So I like the explicit `import`, If not required I think it would be helpful if it were at least allowed.

lukewagner:
> What about just `instance`? Seems potentially better than `dependency` at conveying the idea that it's specifically a dependency on an instance, rather than an interface.

That's a good thought, and it makes technical sense if you're familiar with how `instance` definitions and types work in the C-M, but in my experience saying "instance" almost always ends up confusing folks if they're not already intimately familiar with the C-M, so I've learned to avoid leaning on it when trying to be accessible.

The nice thing about the word "dependency" is that, from what I've seen, the word "dependency" is used by package managers and their associated build-config files (e.g., `npm`/`package.json`, `cargo`/`Cargo.toml`, etc) to exclusively refer to *implementations*.  And in addition to being able to import these *dependencies*, you can also import things directly from the host that are *not* in the dependencies (e.g., in Node, `import { readFile } from 'node:fs';`), suggesting that `import` is the more general "pull in functionality" word whereas `dependency` is the "pull in a particular implementation" word.

Just to see an example of `dependency` imports mixed in with other imports:
```wit
world w {
  import wasi:keyvalue/store;
  import wasi:http/handler;
  import dependency gh:sqlite@{>=1.0.0};
}
```
IMO, that seems decently unambiguous (and an improvement from the original proposed syntax).  But that's just from my limited perspective; happy to discuss more!

design/mvp/WIT.md
lukewagner: As an alternative to adding unnamed interfaces/worlds (which open up another set of design questions), we could instead add a bit of also-useful syntax that allows the WIT to capture precisely how the range query was resolved:
```wit
package gh:sqlite@1.1.1 {
  interface exports { ... }
}
world w {
  import dependency gh:sqlite@{>=1.0.0} = gh:sqlite/exports@1.1.1;
}
```
This `= <pkgname>` right-hand-side of the `import dependency` is necessary for advanced cases where the WIT needs to nest multiple versions of the same package such that the `<pkgnamequery>` alone would be ambiguous.  Importantly for our purposes here, though, none of the name/version info in the `= <pkgname>` shows up in the import so our world `w` above just contains `(import "unlocked-dep=<gh:sqlite@{>=1.0.0}>" (instance ...))` (no `/exports`, no `1.1.1`), and we didn't need anonymous interfaces to achieve that.

PR #405, opened 2024-10-14
Add 'stream' and 'future' types

lukewagner:
This PR builds on #363 to add `future` and `stream` type constructors along with the associated canon built-ins (`{stream,future}.new`, `{stream,future}.{read,write}`, `{stream,future}.cancel-{read,write}`).  It also generalizes the "`async_subtasks`" table to also hold futures and streams, renaming it to be the "`waitables`" table and renaming `subtask.drop` to `waitable.drop`.  Lastly, a new `canonopt` called `always-task-return` is added so that sync-lifted exports can use `task.return` to return their value, which you need if you want to return a stream or future from a synchronous function.

I'd suggest reading the new text in Explainer.md/Binary to get a short overview of the concrete syntactic/binary additions, then Async.md to get the high-level summary, then CanonicalABI.md to get the full details.


design/mvp/CanonicalABI.md
```diff
@@ -535,28 +554,35 @@ using a `callback`, by returning to the event loop) to learn about progress
...
+either return a tuple of scalar values to deliver to core wasm, or `None`. This
+flexibility allows multiple redundant events to be collapsed into one (e.g.,
+when a `Subtask` advances `CallState` multiple times before the event enqueued
+by the initial state change is delivered) and also for events to be
+retroactively removed (e.g., when a `stream.cancel-read` "steals" a pending
+`STREAM_READ` event that was enqueued but not yet delivered). Although this
+Python code represents `events` as a list of closures, an optimizing
+implementation should be able to avoid dynamically allocating this list and
+instead represent `events` as a linked list embedded in the elements of the
+`waitables` table (noting that, by design, any given `watiables` element can be
```
dicej, line 584:
```suggestion
`waitables` table (noting that, by design, any given `waitables` element can be
```

design/mvp/Binary.md
```diff
@@ -290,7 +292,19 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift
            | 0x0a m:<core:memdix>                                => (canon task.wait (memory m) (core func)) 🔀
            | 0x0b m:<core:memidx>                                => (canon task.poll (memory m) (core func)) 🔀
            | 0x0c                                                => (canon task.yield (core func)) 🔀
-           | 0x0d                                                => (canon subtask.drop (core func)) 🔀
+           | 0x0d                                                => (canon waitable.drop (core func)) 🔀
+           | 0x0e t:<typeidx>                                    => (canon stream.new t (core func)) 🔀
+           | 0x0f                                                => (canon stream.read (core func)) 🔀
```
dicej, line 297: BTW, would it be too out-of-scope to add `{stream|future}.write-error` to this list just to reserve the byte codes, with the semantics left for a later PR?

design/mvp/Binary.md
```diff
@@ -290,7 +292,19 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift
...
+           | 0x10                                                => (canon stream.write (core func)) 🔀
+           | 0x11 async?:<async?>                                => (canon stream.cancel-read async? (core func)) 🔀
+           | 0x12 async?:<async?>                                => (canon stream.cancel-write async? (core func)) 🔀
+           | 0x13 t:<typeidx>                                    => (canon future.new t (core func)) 🔀
+           | 0x14                                                => (canon future.read (core func)) 🔀
+           | 0x15                                                => (canon future.write (core func)) 🔀
+           | 0x16 async?:<async?>                                => (canon future.cancel-read async? (core func)) 🔀
```
dicej, line 304: Maybe we talked about this earlier and I forgot, but could we consider leaving the cancel functions as post-WASIp3 (or nice-to-have-for-WASIp3) features?

design/mvp/Async.md
```diff
@@ -180,18 +203,80 @@ invocation of an export by the host. Moreover, at any one point in time, the
 set of tasks active in a linked component graph form a forest of async call
 trees which e.g., can be visualized using a traditional flamegraph.
 
-The Canonical ABI's Python code enforces Structured Concurrency by maintaining
-a simple per-[`Task`] `num_async_subtasks` counter that traps if not zero when
-the `Task` finishes.
+The Canonical ABI's Python code enforces Structured Concurrency by incrementing
+a per-[`Task`] counter when a `Subtask` is created, decrementing when a
```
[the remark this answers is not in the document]
lukewagner: You're definitely right that the impl can use a variety of techniques; it's not semantically observable whether it uses a literal integer counter.  I try to capture this spec-vs-impl distinction by saying "The Canonical ABI's Python code" (where the Python code serves as the spec), but let me know if you'd suggest a better way to clarify.

design/mvp/Explainer.md
```diff
@@ -1404,9 +1470,40 @@ switch to another task, allowing a long-running computation to cooperatively
...
+read or written to the pointer or the sentinel "`BLOCKED`" value otherwise.
+(See [`canon_future_read`] in the Canonical ABI explainer for details.)
+
+The `{stream,future}.cancel-{read,write}` built-ins have type `[i32] -> [i32]`
+and take an index to the matching [readable or writable end](Async.md#streams-and-futures)
+of a stream or future that has an outstanding "`BLOCKED`" read or write. If
+cancellation finished eagerly, the return value is the number of elements read
+or written into the given buffer (`0` or `1` for a `future`). If cancellation
+blocks, the return value is the sentinel "`BLOCKED`" value and the caller must
+`task.wait` for a `{STREAM,FUTURE}_{READ,WRITE}` event to indicate the
```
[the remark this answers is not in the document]
lukewagner: You're right, I forgot to enumerate both options.  In general, any time it says "`task.wait`", there should be a ", or return to event loop if using `callback`".

design/mvp/Binary.md
lukewagner: We talked about postponing *task* cancellation, but this read/write-cancellation is I think different and also more acutely necessary (in particular, I had thought there are cases where we're in a synchronous Drop (or C++ dtor) context and we need to reacquire ownership of a buffer that was previously handed to a `stream.read` or `stream.write`).

dicej: Right, I understand that this is different than task cancellation, and that there needs to be some kind of guest/host coordination to avoid the guest unilaterally reclaiming a buffer while the host still expects to read from or write to it.  Just wondering if the minimum viable coordination might be simpler -- e.g. could calling `waitable.drop` tell the host "any pending reads or writes must be cancelled immediately, and you must not use those buffers anymore"?

design/mvp/Binary.md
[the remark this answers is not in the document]
dicej: To be clear: I realized we do still need a runtime type check anyway -- just saying it helps a lot to know what the type is _supposed_ to be at (pre-)instantiation time.

design/mvp/Binary.md
lukewagner: One tricky detail is that, when a read or write is cancelled, we need to communicate any partial read/write progress (that has already occurred or occurs racily) so that this information isn't lost.  With `cancel-{read,write}`, this partial information can be in the return value (in the synchronous or async-unblocked cases) or in the payload of the associated `READ`/`WRITE` event (in the async-blocked case).

Also, I would have thought that there is a basic need (to implement existing source interfaces) to allow cancelling a read or write without also closing the stream?

design/mvp/Async.md
```diff
@@ -180,18 +203,80 @@ invocation of an export by the host. Moreover, at any one point in time, the
...
+Streams and Futures have two "ends": a *readable end* and *writable end*. When
+*consuming* a `stream` or `future` value as a parameter (of an export call
+with a `stream` or `future` somewhere in the parameter types) or result (of an
+import call with a `stream` or `future` somewhere in the result type), the
+receiver always gets *unique ownership* of the *readable end* of the `stream`
+or `future`. When *producing* a `stream` or `future` value as a parameter (of
+an import call) or result (of an export call), the producer can either
+*transfer ownership* of a readable end it has already received or it can
+create a fresh writable end (via `stream.new` or `future.new`) and lift this
+writable end (maintaining ownership of the writable end, but creating a fresh
```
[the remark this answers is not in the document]
lukewagner: I tried to improve this wording in [this commit](https://github.com/WebAssembly/component-model/commit/4abcbd418e1f7ffb0edebe5e369cc2d45ee275f2); lmkwyt.

design/mvp/Explainer.md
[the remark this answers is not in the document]
lukewagner: Clarified in [this commit](https://github.com/WebAssembly/component-model/commit/50cf971718b8747f097d33cb7e6255af7b7712cd); lmkwyt.

design/mvp/Binary.md
lukewagner: Yes, good idea.

design/mvp/Binary.md
dicej: > Also, I would have thought that there is a basic need (to implement existing source interfaces) to allow cancelling a read or write without also closing the stream?

I've been operating under the assumption that the _only_ time we'd need to cancel a read or write is when we're closing the stream (in which case the app presumably doesn't care about partial read/write progress); and that if you're planning to do more reading or writing you'll be content to wait for any outstanding reads or writes to finish.

In other words, while I agree that being able to cancel read and write operations without closing is useful, I'm not sure it's a must-have feature for p3.  Do you have a specific guest API (e.g. in the JS world) in mind where this feature would be essential?

design/mvp/Binary.md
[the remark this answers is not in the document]
lukewagner: Yes, we can definitely add type immediates if it supports a simpler or more-efficient compilation strategy.  Naively, I had been assuming that there would be some "is it just a memcpy?" fast path with a general fallback path, but thinking about it some more, I could see how, even in the `memory.copy` path, the `typeidx` is useful to provide `elem_size(t)` statically.

Incidentally, one quite-related question I had been wondering about just recently is whether there might be a pragmatic simplification to the CABI (esp. for [BuildTargets.md](https://github.com/WebAssembly/component-model/pull/378)) by removing the `typeidx` immediates from `stream.new`/`future.new`, allowing us to have just 1 core function for each built-in instead of 1 per type (with the requisite name-mangling involved...).  Happy to discuss this more so we can make the right tradeoff.

design/mvp/Binary.md
lukewagner: > (in which case the app presumably doesn't care about partial read/write progress)

I don't know if we can make this assumption in general: if the other side of the stream thinks that N elements have been transferred, but the dropping side doesn't learn about N or has a different N, that seems like it will cause a problem in practice.

> Do you have a specific guest API (e.g. in the JS world) in mind where this feature would be essential?

The use case I had in my mind was if the future returned by calling a Rust `async` `read` or `write` function is dropped: in this case, the stream isn't being closed, but the `read` or `write` needs to be cancelled (without simply waiting for it to complete).

dicej: Rust doesn't have a great cancellation story right now; if you drop a `Future` representing a read or write, you have no way in general to know how much progress it made.  We'd say such `Future`s are not cancellation-safe, i.e. you can't use them in a context where cancellation without closing is a possibility.

And since there's no such thing as an async drop yet, I'm not sure how useful an async cancellation process for CM `stream`s and `future`s would be.  In short, Rust doesn't have a good way to handle such situations yet, and I don't think there's much we can do in the component model to compensate.

dicej: > I don't know if we can make this assumption in general: if the other side of the stream thinks that N elements have been transferred, but the dropping side doesn't learn about N or has a different N, that seems like it will cause a problem in practice.

I can only imagine that being a problem if the dropping side cares about knowing what N is.  If it doesn't (and wouldn't do anything with that information even if it had it), then it doesn't matter.  I agree that if it _does_ care and is only dropping without cancelling because it has no way to express cancellation, then we have a problem.

I guess this hasn't come up much (or at all?) when I've worked on network-based apps.  Generally, I just give up on a connection and close it if e.g. it times out.

lukewagner: Ah, so are you saying that, in Rust, if an `async` `read` or `write` call is cancelled, that it is *already* a dynamic error to attempt to use the stream (being read-from or written-to) again?  I suppose that would simplify the Rust use case.

Considering other languages, it seems like there are various ways in Go to use `context.Context.Done` to cancel `Read()`s or `Write()`s.  In C#, it looks like there's a `ReadAsync` that accepts a `CancellationToken` and doesn't (afaics) close the `Stream` receiver.  This makes me worried to couple cancellation of a read or write with closing a stream.

At an OS level, the analogous operation would be Windows's [`CancelIoEx`](https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-cancelioex) or io_uring's [`IORING_OP_ASYNC_CANCEL`](https://man7.org/linux/man-pages/man2/io_uring_enter.2.html).

dicej: > Ah, so are you saying that, in Rust, if an `async` `read` or `write` call is cancelled, that it is _already_ a dynamic error to attempt to use the stream (being read-from or written-to) again? I suppose that would simplify the Rust use case.

I don't know if it actually raises an error or just behaves non-deterministically; I haven't tried cancelling such a future without dropping the connection -- just assumed it wouldn't go well.

Fair points about Go and C#, though.

design/mvp/Async.md
```diff
@@ -180,18 +203,81 @@ invocation of an export by the host. Moreover, at any one point in time, the
...
+so that it can do something else. Similarly, if `stream.read` or `stream.write`
+would block, they return a "blocked" code so that the caller can continue to
+make progress on other things. But eventually, a task will run out of other
 things to do and will need to **wait** for progress on one of the task's
-subtasks. While a task is waiting, the runtime can switch to other running
-tasks or start new tasks by invoking exports.
+subtasks, readable stream ends, writable stream ends, readable future ends or
+writable future ends, which are collectively called its **waitables**. While a
+task is waiting on its waitables, the Component Model runtime can switch to
+other running tasks or start new tasks by invoking exports.
```
pavelsavara, line 280: > or start new tasks by invoking exports

Does that mean on the same component instance ?
Would it call its exports also when the component is blocked with (synchronous) stack frames by calling `Task.block()` ?

I'm trying to understand if components must be re-entrant on the same thread. Also in case of single-threaded component.

pavelsavara: Or maybe `Task.block` doesn't exist in p3? We always need to yield to host to wait for some promise to resolve ?

That would be better for single-threaded. Because I'm worried that in ST, when you block on a promise, none of the other promises and CPU bound jobs can make progress in the meantime. And that would be a footgun.

lukewagner: > Does that mean on the same component instance ?

Yep!

> Would it call its exports also when the component is blocked with (synchronous) stack frames by calling Task.block() ? [ ... ] I'm trying to understand if components must be re-entrant on the same thread. Also in case of single-threaded component.

Everything is currently single-*threaded* assuming "thread" means "a preemptively scheduled thing". (And when we add multi-threading, there would be some form of opt-in, either at the function-type level or by lifting/lowering with a `shared` option.)

But it sounds like you might also be asking about reentrance on the same *stack*.  Speaking to that question: if one task (= export call) blocks (either by synchronously calling `task.wait` or by returning to its event loop if it's using the `callback` ABI), and then a new task starts because an async export was called, that new export call will be on a separate stack, and thus you don't have to worry about the gnarly reentrance cases that arise when you share a stack.

> Because I'm worried that in ST, when you block on a promise, none of the other promises and CPU bound jobs can make progress in the meantime. And that would be a footgun.

Yes, that's a very important concern.  This design of P3 here where "other tasks can run when one task blocks" is intended to address this, but happy to dig into more cases.

design/mvp/Binary.md
lukewagner: Following up: I think maybe the simplest thing to do is just add a baby version of `error` and the associated built-ins (b/c there aren't many), so I'll go work on that.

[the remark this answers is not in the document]
lukewagner: It sounds like the hot-path is greatly improved, though, so that seems reason-enough to keep the `typeidx`; we have to do per-type mangling for resource types anyhow.

[Fixed](https://github.com/WebAssembly/component-model/commit/924f180185e50f464fe3aceb60c27a62acb2960c), PTAL

design/mvp/Explainer.md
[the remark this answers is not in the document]
dicej: LGTM, thanks.  I don't seem to have the ability to resolve conversations on this repo, but feel free to do that.

design/mvp/Binary.md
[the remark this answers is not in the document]
dicej: Looks good; thanks!

design/mvp/Explainer.md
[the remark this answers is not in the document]
lukewagner: Great.  (Just added you to the repo, so resolve away!)

design/mvp/Explainer.md
```diff
@@ -656,6 +661,47 @@ immediate of a handle type must refer to a `resource` type (described below)
...
+are used to deliver values incrementally over the course of a single async
+function call, instead of copying the values all-at-once as with other
+(synchronous) value types like `list`. The mechanism for performing these
+incremental copies avoids the need for intermediate buffering inside the
+`stream` or `future` value itself and instead uses buffers of memory whose
+size and allocation is controlled by the core wasm in the source and
+destination components. Thus, in the abstract, `stream` and `future` can be
+thought of as inter-component control-flow or synchronization mechanisms.
+
+Just like with handles, in the Component Model, async value types are
```
sunfishcode, line 676: Should this say "Canonical ABI" instead of "Component Model"? I guess the same would apply to the existing text for handles too.

design/mvp/Async.md
pavelsavara: > But it sounds like you might also be asking about reentrance on the same _stack_. 
> that new export call will be on a separate stack

By separate stack, you probably mean WASM stack. But something also must allocate another shadow stack in _guest_ linear memory. Is that LLVM ? Is that linker for each export ?

> Everything is currently single-threaded assuming "thread" means "a preemptively scheduled thing"

So from guest perspective, this is like thread, except we can't create new one ?
Would wasi-libc include pthread headers. 
Maybe `THREAD_MODEL=single` is [close enough](https://github.com/WebAssembly/wasi-libc/pull/518) ?

Would that have another TLS from libc perspective ?

> > Because I'm worried that in ST, when you block on a promise, none of the other promises and CPU bound jobs can make progress in the meantime. And that would be a footgun.
> 
> Yes, that's a very important concern. This design of P3 here where "other tasks can run when one task blocks" is intended to address this, but happy to dig into more cases.

Thank you. 

In dotnet in single-threaded, we simulate thread-pool by running (user code) jobs on the main event loop.
If one of the jobs is synchronously blocked on `task.wait`, the WASM host doesn't know that there are still **pending CPU bound jobs** in the dotnet thread-pool queue. Those would not be processed until host has some other reason to call into the component.

How can we signal to host to call back to us ? Create `wasi-clock` with zero timeout ? Is that cheap enough ?

pavelsavara: Note that threads need to be registered and un-registered inside dotnet VM. It's not exactly cheap operation.

dicej: > By separate stack, you probably mean WASM stack. But something also must allocate another shadow stack in _guest_ linear memory. Is that LLVM ? Is that linker for each export ?

Yeah, LLVM will need to learn to create and switch between multiple shadow stacks in order to support the reentrancy needed when lifting exports using async _without_ a callback.  However, no stack switching is needed when lifting using async _with_ a callback, and I've been assuming that Rust, JS, Python, and C# would all use the callback approach, leaving languages like Go (which is designed around stackful coroutines) to use the non-callback approach.

design/mvp/Async.md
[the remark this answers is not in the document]
dicej: Looks good, thanks.  GitHub still won't let me resolve these conversations myself, BTW.

design/mvp/Binary.md
dicej: BTW, I think we'll need three built-ins for "baby `error`": `error.create`, `error.debug-string`, and `error.drop`.  Does that sound correct?  And are you planning to add a `waitable.close-with-error` (or equivalent, e.g. an optional param to `waitable.drop` or whatever) while you're at it?

lukewagner: Yes and yes!

design/mvp/Async.md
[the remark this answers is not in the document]
dicej: Nevermind, I just hadn't noticed the invite email.  Gonna start resolving now!

design/mvp/Async.md
lukewagner: > So from guest perspective, "re-enter" is like a another thread, except we can't create new one our self ?

Yes, a new green/cooperative thread (and, agreeing with Joel above who notes that `callback`-ABI code doesn't have to worry about multiple linear memory stacks and can continue to reuse just one, since all task-switching happens when you return to the event loop).

> Would wasi-libc include pthread headers ?
> Maybe THREAD_MODEL=single is https://github.com/WebAssembly/wasi-libc/pull/518 ?

Independently, we've talked about using [stack-switching](https://github.com/WebAssembly/stack-switching) (or, if that's not ready in time, a new canon built-in that is implementable-in-terms-of-stack-switching) to provide a pthreads implementation to help in porting code that needs pthreads for concurrency, but not performance/parallelism.  Then, to your point, from a C/wasi-sdk perspective, each new async export call could work as-if this cooperative `pthread_create()` were being called for the exported function.

> Would that have another TLS from libc perspective ?

If we did it via pthreads, yeah, I think it would.  Note that we wouldn't need actual wasm support for this TLS since there are no hardware threads and so we could just use a plain `global` that was kept up-to-date by the libc code.

> In dotnet in single-threaded [...] How can we signal to host to call back to us ? Create wasi-clock with zero timeout ? Is that cheap enough ?

This may or may not answer your question but P3 async has the ability to yield (either by calling `task.yield` or, if using the `callback` ABI, by returning an `i32` with the high bit set).  When yielding, the C-M can switch let other tasks run (other exports from the same component instance, or other tasks in other component instances), but will switch back eventually without blocking for an indeterminate amount of time.  I think perhaps this avoids the need use wasi-clocks with a zero timeout?

design/mvp/canonical-abi/definitions.py
```diff
@@ -1557,9 +1841,110 @@ async def canon_task_yield(task):
...
+  return await cancel_async_copy(ReadableFutureHandle, sync, task, i)
+
+async def canon_future_cancel_write(sync, task, i):
+  return await cancel_async_copy(WritableFutureHandle, sync, task, i)
+
+async def cancel_async_copy(HandleT, sync, task, i):
+  trap_if(not task.inst.may_leave)
+  h = task.inst.waitables.get(i)
+  trap_if(not isinstance(h, HandleT))
+  trap_if(not h.copying_buffer)
```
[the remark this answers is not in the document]
lukewagner: Wow, fantastic catch, thanks!  [Fixed](https://github.com/WebAssembly/component-model/pull/405/commits/3ab3c208c33a3621214fa65ad360ae4fb87afa0b).  The only subtly is that if there was already some progress, we need to report that before reporting `CLOSED`, which is what `pack_async_copy_result` already does, so we just reuse that (which I switched to using for all return paths, for regularity).  PTAL

design/mvp/Async.md
pavelsavara: >  However, no stack switching is needed when lifting using async with a callback, and I've been assuming that Rust, JS, Python, and C# would all use the callback approach

I don't follow. It seems to me that the host could start another async call to export at any time when the component is not running. That means also when it's blocked with a stack.

Would guest (dotnet) be able to opt out from re-entering via export when there is call to `task.wait` ?

Because the gnarly re-entrancy problems continue:
- since the code that is on the blocked stack expects that it "locked" some data structures, which would not mutate before the static call returns. Re-entering could break that invariant, if the locks are just NOOP.
- If re-enter was another pthread, with TID and TLS (even emulated by LLVM+libc), would it also emulate the locks/mutextes ? This path on single (parallel) thread leads to deadlocks.

If we never call `task.wait` from dotnet, that would solve the problem. But there are use case, in which it's useful.
- file system access in dotnet->posix assumes blocking API
- dotnet crypto API is blocking, we need to consume wasi-crypto in blocking way
- DNS name resolution is also blocking in dotnet
- We also can't really prevent user code from interacting with component imports directly calling `task.wait` on something.

dicej: > I don't follow. It seems to me that the host could start another async call to export at any time when the component is not running. That means also when it's blocked with a stack.

If all the async exports use a callback, then no, the host will _not_ reenter the component.  Async exports which use a callback should not block (i.e. no sync-on-async) -- they should return as soon as they can't make further progress, expecting to have their callbacks called when one of the pending async import calls has made progress (at which point the export call can presumably also make progress).  At any given moment, only one core Wasm call is ever happening, and the host will defer any other calls to async functions until the current core Wasm call returns (either because it has completed or because it can't make progress).

In other words, an async export using a callback should _not_ call `task.wait`.  That's sync-on-async, analogous to a poorly-behaved C# program that makes a blocking call in an async function.  And the consequences of doing a sync-on-async call are the same in WASIp3 as in a traditional C# program: it limits concurrency by monopolizing a thread that could be doing other useful work.

> DNS name resolution is also blocking in dotnet

`Dns.GetHostAddressesAsync` can and should be implemented completely asynchronously on WASIp2 and p3.  As for the other blocking APIs: assuming there is no async alternative, then yes, we'll need to look at ways to make .NET reentrancy-safe from LLVM on up.  How do C# async programs handle such APIs today?  Thread workers?  We might need to consider making those APIs out of scope for concurrent WASI programs until we have multithreading of some kind (even if it's just green threads).

design/mvp/canonical-abi/definitions.py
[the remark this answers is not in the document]
sunfishcode: Should the `h.stream.closed()` case in `async_copy` also use `pack_async_copy_result`, for the same reason?

Would it be possible to add tests for these cases?

A design choice here is whether to extend the return value of `cancel-read`/`cancel-write`, and perhaps also plain `read`/`write`, to have a way to indicate "the stream closed after completing N elements". That might require returning more than a single `i32`, but it's doable. In the case where a cancel "races" with a close, this might let successful closes win in more cases. That said, if users really want to see those closes, it seems they could ask for them with the current API as well, by calling `stream.read` with a length of zero.

Fwiw, the WASIp2 API has this, although it was primarily motivated by an early assumption that each `read` host call would have some overhead, and indicating `closed` on the last `read` that returns elements can save an extra `read` call at the end.

design/mvp/Async.md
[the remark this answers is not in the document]
vados-cosmonic: I don't think this is done much in the specs produced by the BCA, but you could note that this section/description is "non-normative".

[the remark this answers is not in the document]
lukewagner: Most of the Python class fields in the "Runtime State" section has the same property of being "spec-internal" and only used to specify the `canon` definitions' externally-visible behavior, freeing an implementation to represent things however it wants, so I added a little blurb in [this commit](https://github.com/WebAssembly/component-model/pull/405/commits/8fd4d55e2c72bd7abc1b7ae48b87629be3363f58) that says so explicitly.  PTAL!

(As a nit: I wouldn't say `todo` is "non-normative", since it is used transitively to define the (normative) behavior of the various `canon_*` definitions.  But hopefully "spec-internal" captures this nuance.)

design/mvp/Async.md
```diff
@@ -180,18 +203,81 @@ invocation of an export by the host. Moreover, at any one point in time, the
...
+a per-[`Task`] counter when a `Subtask` is created, decrementing when a
+`Subtask` is destroyed, and trapping if the counter is not zero when the `Task`
+attempts to exit.
+
+### Streams and Futures
+
+Streams and Futures have two "ends": a *readable end* and *writable end*. When
+*consuming* a `stream` or `future` value as a parameter (of an export call with
+a `stream` or `future` somewhere in the parameter types) or result (of an
+import call with a `stream` or `future` somewhere in the result type), the
```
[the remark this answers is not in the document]
lukewagner: After #368, component-level functions now only have a single result type.  (Types form an expression tree, ofc, but the "somewhere in the" text here was meant to suggest "anywhere in the return type expression".)

design/mvp/Binary.md
```diff
@@ -290,7 +292,19 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift
            | 0x0a m:<core:memdix>                                => (canon task.wait (memory m) (core func)) 🔀
            | 0x0b m:<core:memidx>                                => (canon task.poll (memory m) (core func)) 🔀
            | 0x0c                                                => (canon task.yield (core func)) 🔀
-           | 0x0d                                                => (canon subtask.drop (core func)) 🔀
+           | 0x0d                                                => (canon waitable.drop (core func)) 🔀
+           | 0x0e t:<typeidx>                                    => (canon stream.new t (core func)) 🔀
+           | 0x0f t:<typeidx>                                    => (canon stream.read t (core func)) 🔀
```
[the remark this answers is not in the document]
lukewagner: At the moment, the readable/writable end of a future/stream inherits the `canonopts` of the import/export call that produced/consumed the stream, fixing those `canonopts` for the lifetime of the stream (instead of allowing each `stream.read`/`stream.write` to potentially use a different `canonopts`.  In the future when we start to optimize producer+consumer pairs via little fused copy adapters (beyond the basic `memory.copy` case), it seemed like this might be a useful invariant since it lets us "set up" once up-front instead of every time `stream.read`/`stream.write` is called.  But happy to discuss how you're seeing this work in practice.

design/mvp/Async.md
pavelsavara: > How do C# async programs handle such APIs today? Thread workers? 

- in the browser we can't use async `SubtleCrypto` in single-threaded build for this reason. Maybe JSPI could help but I have same worries about re-entrancy. Good news is that we could test that already in the browser. We probably should.
- emscripten't virtual file system is synchronous posix style
- there is no DNS lookup in the browser
- in multi-threading build we can solve it by blocking the thread and doing the promise waiting on UI thread.
- In the normal desktop/server dotnet, there are always threads to rescue this. I think that overlapped IO is equivalent of promises, but I don't have experience with it.

> We might need to consider making those APIs out of scope for concurrent WASI programs until we have multithreading of some kind (even if it's just green threads).

😭

Not having crypto in the browser is barely acceptable. Not having it in server will limit use-cases.

Going without synchronous file system would be interesting challenge for dotnet.

We could have WASI-only platform-abstraction-layer in same way as we have for HTTP client, that would be async only C# APIs. (As opposed to shared posix/wasi-libc blocking one)

As a consequence, that would rule out many many existing C# user applications, which use synchronous FS APIs.

And I'm not even confident that C# file system async-only APIs are complete enough.

I think synchronous file system access problem will hit all other async guest languages. 

And probably also console stream ?

> As for the other blocking APIs: assuming there is no async alternative, then yes, we'll need to look at ways to make .NET reentrancy-safe from LLVM on up. 

Yeah, that sounds like tall order.

design/mvp/Explainer.md
lukewagner: At this point in the explainer, we haven't introduced "the Canonical ABI", and so this text is meant to be symmetric to the preceding text in the "Handle Types" section.  There could be a good refactoring that does introduces "the Canonical ABI" properly (perhaps before introducing the component-level types, describing the relationship between compoennt-level types and core types), and I even started doing this just now, but then I thought I should probably keep that refactoring out of this streams PR so it can stay focused.  But that would be a good follow-on.

design/mvp/Binary.md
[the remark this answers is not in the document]
dicej: Ah, okay; I missed that.  I'm having trouble wrapping my head around it, though.  If I have two imported functions that return `stream<string>` but use different string encodings, I kind of have to treat them as two different types, right?  It's not clear to me how either the guest binding generator or the host would determine which encoding to use for an arbitrary `stream.read` -- it depends on which imported function I got the stream handle from, and I don't know how we'd connect the two statically.

In my prototype, I specify the options on each call, so there's no chance of confusion.  That _seems_ simpler to me, but maybe I'm missing something.

design/mvp/Async.md
lukewagner: One extra affordance we have in this proposal that I think addresses some or all of those use cases is the ability for an otherwise-async-using component to make a *synchronous* import call during which the component instance cannot execute or be reentered.  (In the Canonical ABI code, this is modeled by the `interruptible` flag that is set when starting the sync import, cleared when the sync import returned, and automatically applies backpressure when anyone tries to make an export call.). Of course this fully blocks that one component instance, but not all the other components.  In a browser, this would be equivalent to using JSPI and maintaining a per-component-instance flag that prevents exports from being called when this flag is set by immediately returning a Promise (to be resolved at some point in the future once the sync call finishes and execution can resume in the component instance).

dicej: > Indeed, even more than "should", `task.wait` (and `task.yield`, which is the other built-in that allows reentrance) [`trap_if`](https://github.com/WebAssembly/component-model/blob/add-stream/design/mvp/CanonicalABI.md#-canon-taskwait) called from an export that has a `callback`, so if a producer toolchain uses the `callback` ABI (like, I believe, .NET would), it can rely on not getting reentered while core wasm is on the stack and thus continue to use a single shadow stack.

Given @pavelsavara's concerns, perhaps we should allow `task.wait` to be called from an export that has a callback, but still disallow reentrance (i.e. defer any would-be-reentering calls until the export returns).  That would come at a cost to concurrency (proportional to how long `task.wait` actually needs to block), but provide flexibility for apps that need to do blocking things here and there in an otherwise async context.

dicej: > the ability for an otherwise-async-using component to make a _synchronous_ import call during which the component instance cannot execute or be reentered

Per my comment above (which I wrote before I saw yours), perhaps we should make calls to `task.wait` do the same thing.

pavelsavara: > defer any would-be-reentering calls until the export returns

Yes please

design/mvp/Binary.md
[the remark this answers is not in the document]
lukewagner: So what's currently-proposed is that the readable and writable handles both contain a copy of the `canonopts` of the corresponding `lower`/`lift` that produced/consumed them, resp., and so these would be fed in as dynamic parameters to the generic copy routine noting that, when the `read` or `write` is blocked, you sortof have to bottle up these immediates as data (while you wait for the other side) anyhow.  (And then in an optimizing future, I thought that these `canonopt`s could be replaced by a `funcref` pointing to a copy thunk, where the `funcref` was chosen once by the fused adapter that produced/consumed the readable/writable end, resp.)

If we wanted to lock things down even further, we could give `stream.read`/`stream.write` and `opts` immediate and then `trap_if` the immediate was different from how the stream was initially lifted/lowered (kindof like this PR now does for the element type `t`), so that we have all the information statically both up-front and at each `read` or `write`, but I'd be curious to hear more of how much that helped.

design/mvp/Async.md
lukewagner: Yeah, we could totally do that (giving `task.wait` an optional `async` immediate, just like `lower`).  I had wanted to see whether there was a concrete use case for this (where you want to wait on *multiple* async things, but not activate the general async scheduler, so you're half-async) before speculatively adding such an option, but if yall are pretty sure there is, it's easy enough to spec.

design/mvp/Binary.md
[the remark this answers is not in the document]
dicej: I think my main concern is how to generate guest bindings.  I need to generate lifting and lower code for each payload type, and now it sounds like I'll potentially need to generate such code not just for each payload type but for each unique (payload type, canonical options) pair.  Which gets truly intimidating when you consider that code can be separately compiled by distinct toolchains for distinct languages (each with it's own favorite string encoding) and then either statically or shared-everything-dynamically linked together into a single component, with handles passed between those separately-compiled pieces.  Keeping track of which handle needs which options sounds like a tall order.

Seems much simpler to let each `read`/`write` call use whatever options it wants to and _not_ attach options dynamically to each handle.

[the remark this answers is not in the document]
lukewagner: Got it, thanks for explaining!  So iiuc, the use case here is: I have a single callsite where I want to `stream.read` or `stream.write` with a fixed `canonopts` but I'm consuming streams that might've been lowered from different `canon lower`s with potentially-different `canonopts` and this could Just Work.  I think I agree with that, so I'll add `canonopts` to `{stream,future}.{read,write}`.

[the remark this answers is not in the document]
lukewagner: Added in [this commit](https://github.com/WebAssembly/component-model/pull/405/commits/81591728e2fcc416d8c77807f2f5ed0513438195).  Working on this surfaced a tricky interaction (caught by tests) for `stream<borrow<R>>`: we do need to remember the `Subtask` or `Task` that is consuming/producing this stream (so we can account for the `borrow` rules as-if it were a `list<borrow<R>>`), but we don't want to use the `canonopt`s of that `Subtask`/`Task`.  This required some superficial refactoring of `CallContext` which was the majority of the commit.

design/mvp/Binary.md
dicej: I'm going to mark this conversation resolved, since it's clear that at least some popular languages support cancelling read and write operations without closing the stream, which is enough reason to make that part of the initial spec.

design/mvp/Async.md
lukewagner: Ok, added in [this commit](https://github.com/WebAssembly/component-model/pull/405/commits/fd9fbf1b95ff569dd059746c4bfa7b6c8678d3dd).  With this, `task.{wait,poll,yield}` and `{stream,future}.{read,write}` can all set (or not set) the `async` attribute.  Let me know if this addresses these concerns.

dicej: LGTM; thanks!  @pavelsavara any other concerns about this topic?

lukewagner:
@dicej Great point, and agreed.  Split back up in [this commit](https://github.com/WebAssembly/component-model/pull/405/commits/916bb30fc9b05764a3033010d75b4ee00d19b2a9).

design/mvp/Explainer.md
```diff
@@ -1500,11 +1509,11 @@ blocks, the return value is the sentinel "`BLOCKED`" value and the caller must
...
 
-The `waitable.drop` built-in has type `[i32] -> []` and removes the indicated
-[subtask](Async.md#subtask-and-supertask) or [stream or future](Async.md#streams-and-futures)
-from the current instance's [waitables](Async.md#waiting) table, trapping if
-the subtask isn't done or the stream or future is in the middle of reading
-or writing.
+The `{stream,future}.close-{readable,writable}` built-ins have type
+`[i32] -> []` and removes the indicated [stream or future](Async.md#streams-and-futures)
+from the current component instance's [waitables](Async.md#waiting) table,
+trapping if the stream or future has a mismatched direction or type or are in
```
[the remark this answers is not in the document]
lukewagner: My impression is that it's better to keep these orthogonal instead of partially-overlapping.  This would also help producer toolchains understand that cancellation is potentially blocking.

design/mvp/Binary.md
lukewagner: Ok, done in [this commit](https://github.com/WebAssembly/component-model/pull/405/commits/f55b5906f7f47e3fd3e02fb86a6295342c4c1fed).  PTAL!

lukewagner:
For visibility: [this commit](https://github.com/WebAssembly/component-model/pull/405/commits/f55b5906f7f47e3fd3e02fb86a6295342c4c1fed) adds `error` and its associated canon built-ins, closing the TODOs in the original PR.

design/mvp/canonical-abi/definitions.py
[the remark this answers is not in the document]
lukewagner: (Note that the hard-coded `CLOSED` has now been replaced with a `pack_async_copy_result` to pull in the `error` logic.)

design/mvp/CanonicalABI.md
```diff
@@ -2423,12 +3019,359 @@ For a canonical definition:
...
+
+For canonical definitions:
+```wasm
+(canon stream.close-readable $t (core func $f))
+(canon stream.close-writable $t (core func $f))
+(canon future.close-readable $t (core func $f))
+(canon future.close-writable $t (core func $f))
+```
+validation specifies:
+* `$f` is given type `(func (param i32))`
```
dicej, line 3263: The `close-writable` functions should have a different core signature now that they take optional errors, correct?

design/mvp/Explainer.md
```diff
@@ -630,6 +634,29 @@ a single NaN value. And boolean values in core wasm are usually represented as
...
+
+The intention of this highly-non-deterministic semantics is to provide hosts
+the full range of flexibility to:
+* append a basic callstack suitable for forensic debugging in production;
+* optimize for performance in high-volume production scenarios by slicing or
+  discarding error messages;
+* optimize for developer experience in debugging scenarios when debug metadata
+  is present by appending expensive-to-produce symbolicated callstacks.
+
+A consequence of this, however, is that components *must not* depend on the
```
dicej, line 654: We could potentially elaborate on this by mentioning that, when defining a WIT interface, best practice for supporting recoverable errors for a stream is to pair that stream with a future of the appropriate error type, e.g. `tuple<stream<u8>, future<my-error>>`.

design/mvp/Binary.md
dicej: LGTM overall; I left a couple of comments below.

design/mvp/canonical-abi/definitions.py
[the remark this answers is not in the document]
sunfishcode: Thinking more about the question of "the stream closed after completing N elements", I convinced myself that we don't need it for now. It only helps in a "you can get lucky more often" way, and it probably is rare in practice. And we can add it in the future if we find a need for it.

design/mvp/CanonicalABI.md
lukewagner: Good catch, thanks!  [fixed](https://github.com/WebAssembly/component-model/pull/405/commits/78d2ece13f0351cfcf526cd7ec0b06b35dde559e)

design/mvp/Explainer.md
lukewagner: Great suggestion, [done](https://github.com/WebAssembly/component-model/pull/405/commits/cca454bdeaf41d70ea121757d53003f93e48ba4c)

lukewagner:
Thanks for all the excellent feedback everyone!  I think the comments are dying down and it sounds like this is on a convergence path with @dicej's impl, so if nothing new pops up, I think I'll merge mid next week (and then we can continue to iterate on async in new issues and PRs).

FWIW, I think the next change I'd like to make is a PR that removes the `DONE` state from subtasks, which should simplify things and allow bindings for async borrows to make sense, and, after that, go through the top bullets of Async.md#TODO.

design/mvp/Binary.md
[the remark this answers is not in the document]
lukewagner: As a further follow-up, with `opts` now coming from `stream.write $opts` instead of the lifting, the restriction of not allowing `stream.write` before the writable end is paired can be removed (and instead performed in the synchronous case, where it's a guaranteed deadlock).  Working on this I also realized an invalid assertion (that unpaired writable streams could not have an outstanding buffer -- this is possible if you start a write and receive your own stream back before the write completes), which is now removed and tested for.  See [this commit](https://github.com/WebAssembly/component-model/pull/405/commits/43d348d1f4fc13922ededd709538a754f4b67b29)

design/mvp/Async.md
pavelsavara: my concern are resolved, thanks

design/mvp/Binary.md
```diff
@@ -287,10 +290,29 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift
...
-           | 0x0b m:<core:memidx>                                => (canon task.poll (memory m) (core func)) 🔀
-           | 0x0c                                                => (canon task.yield (core func)) 🔀
+           | 0x0a async?:<async>? m:<core:memdix>                => (canon task.wait async? (memory m) (core func)) 🔀
+           | 0x0b async?:<async>? m:<core:memidx>                => (canon task.poll async? (memory m) (core func)) 🔀
+           | 0x0c async?:<async>?                                => (canon task.yield async? (core func)) 🔀
            | 0x0d                                                => (canon subtask.drop (core func)) 🔀
+           | 0x0e t:<typeidx>                                    => (canon stream.new t (core func)) 🔀
+           | 0x0f t:<typeidx> opts:<opts>                        => (canon stream.read t opts (core func)) 🔀
+           | 0x10 t:<typeidx> opts:<opts>                        => (canon stream.write t opts (core func)) 🔀
+           | 0x11 async?:<async?>                                => (canon stream.cancel-read async? (core func)) 🔀
```
dicej, line 300: `{stream|future}.cancel-{read|write}` are the only `{stream|future}` functions that don't take a `t:<typeIdx>`.  For consistency (and because it's helpful to know statically in the implementation which table we're going to look at), could we add that?

lukewagner: Yeah, that makes sense.  [Fixed](https://github.com/WebAssembly/component-model/pull/405/commits/5ba57ca3352ee09060b75beff3bf0f587dd87168)

lukewagner:
One other update: for both the pragmatic short-term benefit of avoiding parsing conflicts with the `error` resource type (defined by `wasi:io/error`) and, more generally, to help make it clear that it's just meant to convey non-deterministic, debugging-oriented contextual data, [this commit](https://github.com/WebAssembly/component-model/pull/405/commits/ee522f7a53ee04361d8999fc61df85e8f2c6de66) renames `error` to `error-context`.  This also nicely distinguishes `error-context` from the `error` case of the `result` variant (which contains the "whole" error value, including the WIT-defined error payload value).

PR #407, opened 2024-10-24
Reorganize the Canonical Built-ins section.

sunfishcode:
Reorganize the "Canonical Built-ins" section of Explainer.md.
 - Don't quote the `canon` ebpf here, as it doesn't have a lot of explanatory information.
 - Put builtins in their own subsections.
 - Give builtins synopsis tables that include their conceptual signatures, binary encoding immediates, and canonical ABI signatures.
 - Revise documentation to separate the conceptual signatures and semantics from the Canonical ABI representations.

I started out with a table in Async.md, but as I worked it ended up seeming to make sense to just evolve the content in Explainer.md. I'm curious what you think!

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -1213,6 +1213,7 @@ to wrap and a list of configuration options:
 ```ebnf
 canon    ::= (canon lift core-prefix(<core:funcidx>) <canonopt>* bind-id(<externdesc>))
            | (canon lower <funcidx> <canonopt>* (core func <id>?))
+           ...
 canonopt ::= string-encoding=utf8
            | string-encoding=utf16
            | string-encoding=latin1+utf16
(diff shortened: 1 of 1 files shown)


design/mvp/Explainer.md
```diff
@@ -1361,49 +1362,59 @@ In addition to the `lift` and `lower` canonical function definitions which
...
-resource (with resource type `typeidx`) with the given `i32` value as its
-representation and returning the `i32` index of a new handle pointing to this
-resource.
+###### `resource.new`
+
+| Synopsis                   |                           |
+| -------------------------- | ------------------------- |
+| Conceptual signature       | `func<T>(repr: ...) -> T` |
+| Binary encoding immediates | `T:<typeidx>`             |
+| Canonical ABI signature    | `[repr:i32] -> [i32]`     |
```
lukewagner, line 1375: Skimming through them, the 'Conceptual signature' row doesn't seem to add much value over the 'Canonical ABI Signature' row and sortof adds a bit of its own confusion, since it has '...' and its own custom syntax which one now has to extrapolate from.  Independently, I think the 'Binary encoding immediates' are a bit out of place, since if you're implementing a binary decoder and interested in this information, you're going to need to read Binary.md anyways to see where and how the immediates are embedded in the overall definition.  Thus, I think we should replace the 3-row table in each of these sub-headings with a single triple-backtick code block that contains only the Canonical ABI Signature.

Separately, I see the EBNF for `resource.new` and all the other built-ins is removed above, so I think we should introduce these productions in each of these subsections for reference.  So, combined with the above, maybe an example for `resource.new` could look like:

"""
The syntax for the `resource.new` built-in is:
```
canon ::= ...
        | (canon resource.new <typeidx> (core func <id>?))
```
the Core WebAssembly signature for this built-in defined by the Canonical ABI is:
```
[repr:i32] -> [handleidx:i32]
```
... prose describing both together ...
"""

design/mvp/Explainer.md
```diff
@@ -1361,49 +1362,59 @@ In addition to the `lift` and `lower` canonical function definitions which
...
+
+The `resource.new` built-in creates a new resource (with resource type
+`T`) with `repr` as its representation and returning the a new
+handle pointing to this resource.
```
lukewagner, line 1379:
```suggestion
`T`) with `repr` as its representation and returning the index of a new handle
in a per-resource-type-per-component-instance table that points to a resource
that contains `repr`.
```

lukewagner (commented):
Thanks for doing this, this seems like a much better organization and breakdown for this content!  One quick high-level request to sort out first:

sunfishcode:
I'll close this for now, work on incorporating your feedback, and submit a new PR on the main branch once the stream/future PR is merged.

PR #415, opened 2024-11-19
Tidy up whitespace in Binary.md.

sunfishcode:
Make a small whitespace change in Binary.md to make `section_1` consistent with the other sections, and to vertically align it when displayed in a fixed-width font.


lukewagner:
Thanks!

PR #419, opened 2024-11-25
Rename `hw_concurrency` to `available_parallelism`.

sunfishcode:
Following the general computer-science meanings of the terms "concurrency" and "parallelism", rename `thread.hw_concurrency` to `thread.available_parallelism`, because it's about the number of threads that can execute in parallel.

This differs from the existing JS API of [`navigator.hardwareConcurrency`], so add a note mentioning the correspondence.

[`navigator.hardwareConcurrency`]: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hardwareConcurrency

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -287,7 +287,7 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift 
            | 0x07 rt:<typdidx>                                   => (canon resource.drop rt async (core func))
            | 0x04 rt:<typeidx>                                   => (canon resource.rep rt (core func))
            | 0x05 ft:<typeidx>                                   => (canon thread.spawn ft (core func)) 🧵
-           | 0x06                                                => (canon thread.hw_concurrency (core func)) 🧵
+           | 0x06                                                => (canon thread.available_concurrency (core func)) 🧵
            | 0x08                                                => (canon task.backpressure (core func)) 🔀
            | 0x09 ft:<core:typeidx>                              => (canon task.return ft (core func)) 🔀
            | 0x0a async?:<async>? m:<core:memdix>                => (canon task.wait async? (memory m) (core func)) 🔀
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -3418,22 +3418,22 @@ def canon_thread_spawn(f, c):
     return [-1]
 ```
 
-### 🧵 `canon thread.hw_concurrency`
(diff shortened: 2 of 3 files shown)


yoshuawuyts:
As the person who added [`std::thread::available_parallelism`](https://github.com/rust-lang/rust/issues/74479) to Rust, you'll be unsurprised to lean I'm in favor of this change. The terminology we used there was actually based on @aturon's PhD thesis on concurrency. Here is the motivation section I wrote [at the time](https://github.com/rust-lang/rust/pull/89324):

> ## Rationale
>
> The API was initially named std::thread::hardware_concurrency, mirroring the [C++ API of the same name](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency). We eventually decided to omit any reference to the word "hardware" after https://github.com/rust-lang/rust/pull/74480#issuecomment-662045841. And so we ended up with available_concurrency instead. For a talk I was preparing this week I was reading through ["Understanding and expressing scalable concurrency" (A. Turon, 2013)](http://aturon.github.io/academic/turon-thesis.pdf), and the following passage stood out to me (emphasis mine):
> 
> > **Concurrency is a system-structuring mechanism.** An interactive system that deals with disparate asynchronous events is naturally structured by division into concurrent threads with disparate responsibilities. Doing so creates a better fit between problem and solution, and can also decrease the average latency of the system by preventing long-running computations from obstructing quicker ones.
> 
> > **Parallelism is a resource.** A given machine provides a certain capacity for parallelism, i.e., a bound on the number of computations it can perform simultaneously. The goal is to maximize throughput by intelligently using this resource. For interactive systems, parallelism can decrease latency as well.
> 
> _Chapter 2.1: Concurrency is not Parallelism. Page 30._
> 
> _"Concurrency is a system-structuring mechanism. Parallelism is a resource."_ — It feels like this accurately captures the way we should be thinking about these APIs. What this API returns is not "the amount of concurrency available to the program" which is a property of the program, and thus even with just a single thread is effectively unbounded. But instead it returns "the amount of _parallelism_ available to the program", which is a resource hard-constrained by the machine's capacity (and can be further restricted by e.g. operating systems).

lukewagner:
Agreed!  It's good to see the prior art for `available_parallelism`.  I suppose let's let this sit for a bit to collect feedback, but then I'm quite happy to merge.

lukewagner:
(I forgot above, but cc @abrown)

abrown (approved):
I think this is the right change to make, especially if other languages are doing this. The sole reason I named this `hw_concurrency` initially was to line up with the JS version of this. Given that this functionality has moved to the component model, just documenting that relationship and using the new name seems fine.

PR #420, opened 2024-11-25
Remove an unneeded `None` check.

sunfishcode:
`despecialize` doesn't return an optional type, so there's no need to check it for `None`.

### DIFF
--- a/design/mvp/canonical-abi/definitions.py
+++ b/design/mvp/canonical-abi/definitions.py
@@ -808,8 +808,6 @@ def contains_async_value(t):
 def contains(t, p):
   t = despecialize(t)
   match t:
-    case None:
-      return False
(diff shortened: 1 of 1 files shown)


lukewagner:
`None` comes up when we recurse into the `Option[ValType]` of a variant case.  Alternatively, we could guard the recursive call to `contains(c.t, p)` in the `VariantType` case for `None` which I suppose is cleaner.

sunfishcode:
Oh, I see now. I had inferred `t` was `ValType` from seeing this:
```python
def despecialize(t):
  match t:
    case TupleType(ts)       :  ...
```
so that the `return t` at the end returned `ValType` as well, but I see now that `t` is `Option[ValType]` and that the above code also works with that.

PR #424, opened 2024-12-03
Mention that the async model can be implemented using callbacks

sunfishcode:
In the High-level Approach section, mention that the component model's approach to native async can also be implemented in terms of callbacks, in addition to language-level async, stackful coroutines, and green threads.

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -58,8 +58,8 @@ additional goals and requirements for native async support:
 
 Based on the above goals, the Component Model's approach to native async starts
 by allowing components to import and export "async" functions which abstract
-over, and bindings can be generated in the idiomatic style for each programming
-language:
+over, and can be implemented by, idiomatic concurrency in a variety of
+programming languages:
 * `async` functions in languages like C#, JS, Python, Rust and Swift
   (implemented using [`callback` functions](#waiting))
 * stackful coroutines in languages like Kotlin, Perl, PHP and (recently) C++
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -61,12 +61,12 @@ by allowing components to import and export "async" functions which abstract
 over, and bindings can be generated in the idiomatic style for each programming
 language:
 * `async` functions in languages like C#, JS, Python, Rust and Swift
-  (using [`callback` functions](#waiting))
+  (implemented using [`callback` functions](#waiting))
 * stackful coroutines in languages like Kotlin, Perl, PHP and (recently) C++
 * green threads as-if running on a single OS thread in languages like Go and
   (initially and recently again) Java
 * callbacks, in languages with no explicit async support
-  (also using [`callback` functions](#waiting))
+  (also implemented using [`callback` functions](#waiting))
 
 The Component Model supports this wide variety of language features by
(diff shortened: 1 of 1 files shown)


design/mvp/Async.md
```diff
@@ -63,6 +63,7 @@ over and can literally be implemented in terms of:
 * stackful coroutines in languages like Kotlin, Perl, PHP and (recently) C++
 * green threads as-if running on a single OS thread in languages like Go and
   (initially and recently again) Java
+* callbacks, synchronous functions that are called in response to events
```
dicej, line 66: My understanding was that the first bullet point covered the callback case (i.e. their stackless async machinery would be implemented by way of persisting futures/promises/tasks to the heap so that the guest can return control to the host by returning from the exported function, and when the host calls the callback, the guest resumes the appropriate future/promise/task for that event).  I imagine only C would expose the callback to application code directly; for other languages it would be a runtime implementation detail hidden from application code.

dicej: But yeah, more words would make that much clearer.

lukewagner: Could we merge this new bullet by extending the sentence of the old bullet: "`async` functions in languages like C#, JS, Python, Rust and Swift that can compile down to stackless callback functions that are called in response to events"?

sunfishcode: It seemed to me this list was written for the perspective of users of the languages, so in C#/JS/Python/etc. users get presented with async function bindings, in Kotlin/Perl/etc. users get bindings that use coroutines, in Go users get bindings that effectively use green threads. From that perspective, C with its exposed callbacks would be a separate bullet, rather than lumped in with async functions, even if they interface with the ABI using the same mechanism.

dicej: Given that neither JS nor Python are (currently) compiled, I'd omit the "compile down to" phrase and say something like "`async" functions in languages like C#, JS, Python, Rust and Swift produce stackless tasks which may be resumed via callback functions that are called in response to events."

dicej: @sunfishcode yes, that's reasonable, as long as we can make it clear that all the stackless languages are using the callback ABI under the hood.

BTW, we could even support both modes in C if desired: stackful (i.e. no callback functions) and stackless.  The stackful mode would require guest toolchain (e.g. LLVM) support for (shadow) stack switching, of course.

sunfishcode: Ok, I've now added text that mentions that the stackless languages are using callback functions, and clarified the sentence that introduces the list.

dicej (approved):

design/mvp/Async.md
```diff
@@ -58,11 +58,15 @@ additional goals and requirements for native async support:
 
 Based on the above goals, the Component Model's approach to native async starts
 by allowing components to import and export "async" functions which abstract
-over and can literally be implemented in terms of:
+over, and bindings can be generated in the idiomatic style for each programming
+language:
```
lukewagner, line 62:
```suggestion
over, and can be implemented by, idiomatic concurrency in a variety of
programming languages:
```

lukewagner (approved):
Oops, sorry, this fell off my radar from before.  One small grammar nit:

lukewagner:
Thanks!

PR #425, opened 2024-12-03
Remove the final "done" state of subtasks from the async ABI

lukewagner:
This PR has 2 commits, the first one is pure refactoring, no observable change (moving code from `Subtask` into `canon_lower`, which makes the logic easier to follow, imo), the second one removes `CallState.DONE`.

Before this PR, excusive ownership of borrowed handles is returned when a subtask (i.e., `async` import call) reaches the "done" state.  Abstractly this made sense, but working through source-language bindings, there is no good way of expressing when this event happens in the syntax, and this is an important event to express well (since it determines whether dropping/transferring the borrowed handle traps).  Syntactically, the natural place to return borrows is when the subtask returns its value, which is the "returned" state (which precedes the "done" state), since that maps to, e.g., when a `Promise`/`Future` is visibly resolved.  So that's what this PR does.

With that change, there is no need for the "done" state, it just adds noise, so this PR removes the "done" state entirely, allowing the subtask to be dropped by its caller immediately after receiving the return value.  This does mean that subtasks can continue executing (after `task.return`) even after their supertask finishes, but this doesn't break Structured Concurrency (and the existence of an unambiguous call-tree, which is semantically used by the reentrance guard) if we view the supertask as "async tail-calling" any still-executing subtasks when it exits.  A blurb is added to Async.md by this PR explaining this more.

Two other notes:
* The rules for lifting `borrow` are tweaked (in CanonicalABI.md) to account for the fact that a supertask can now return before its subtasks have returned, so simply checking that a `borrow` handle is passed to a nested subtask is no longer a sufficient condition to ensure proper nesting.  However, the new rule is actually more expressive than the old rule, so it's probably what it should have been all along.
* Since the current implementation plan does not allow `borrow` inside `stream<T>` or `future<T>` anyways, this PR changes the validation rules in the proposal to match, which is a conservative change we relax in the future if someone actually finds a use case for this (I'm not aware of any, atm).

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -601,14 +601,12 @@ The `return_` method is called by either `canon_task_return` or `canon_lift`
 (both defined below) to lift and return results to the caller using the
 `on_return` callback that was supplied by the caller to `canon_lift`. Using a
 callback instead of simply returning the values from `canon_lift` enables the
-callee to keep executing after returning its results. However, it does
-introduce a dynamic error condition if `canon task.return` is called less or
-more than once which must be checked by `return_` and `exit`. Since ownership
-of borrowed handles is returned to the caller when the caller is notified that
-the `Subtask` is in the `RETURNED` state, `num_borrows` is guarded to be `0`.
-Since tasks can keep executing arbitrary code after calling `task.return`, this
-means that borrowed handles must be dropped potentially much earlier than the
-observable end of the task.
+callee to keep executing after returning its results. There is a dynamic error
+if `task.return` is not called exactly once and if the callee has not dropped
+all borrowed handles by the time `task.return` is called. This means that the
+caller can assume that ownership all its lent borrowed handles has been
+returned to it when it is notified that the `Subtask` has reached the
+`RETURNED` state.
 ```python
   def return_(self, flat_results):
     trap_if(not self.on_return)
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -241,14 +241,14 @@ class ResourceHandle:
   rep: int
   own: bool
   borrow_scope: Optional[Task]
-  lend_count: int
+  num_lends: int
 
   def __init__(self, rt, rep, own, borrow_scope = None):
     self.rt = rt
     self.rep = rep
     self.own = own
     self.borrow_scope = borrow_scope
-    self.lend_count = 0
+    self.num_lends = 0
 ```
 The `rt` and `rep` fields of `ResourceHandle` store the `rt` and `rep`
 parameters passed to the `resource.new` call that created this handle. The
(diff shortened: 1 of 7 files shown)


design/mvp/CanonicalABI.md
```diff
@@ -609,10 +603,16 @@ The `return_` method is called by either `canon_task_return` or `canon_lift`
 callback instead of simply returning the values from `canon_lift` enables the
 callee to keep executing after returning its results. However, it does
 introduce a dynamic error condition if `canon task.return` is called less or
-more than once which must be checked by `return_` and `exit`.
+more than once which must be checked by `return_` and `exit`. Since ownership
+of borrowed handles is returned to the caller when the caller is notified that
+the `Subtask` is in the `RETURNED` state, `num_borrows` is guarded to be `0`.
+Since tasks can keep executing arbitrary code after calling `task.return`, this
```
dicej, line 609: For my understanding: this is saying that the task _must_ drop all borrows before calling `task.return`, correct?  The wording here is about what the caller can expect rather than what the callee is responsible for doing, so I'm inferring the latter.  Maybe you make the callee's responsibilities explicit wrt borrows elsewhere and I missed it?

dicej (approved):
LGTM; thanks!

Regarding borrows, I think it would help to add a section in Async.md that enumerates how borrowed parameters interact with structured concurrency (i.e the rules which guests must follow and which hosts must enforce), especially now that we've removed the "done" state.  Currently, the topic of borrows is spread across CanonicalABI.md such that you kind of have to read the whole thing to get the whole picture; it would be nice to have a succinct summary to refer people to if they have questions.

design/mvp/CanonicalABI.md
lukewagner: Good point, it was in the code, but not prose; [fixed](https://github.com/WebAssembly/component-model/pull/425/commits/cae44a8b971e72a9a1b30e2adf2c55b59e8f3788).

design/mvp/CanonicalABI.md
```diff
@@ -607,12 +601,16 @@ The `return_` method is called by either `canon_task_return` or `canon_lift`
 (both defined below) to lift and return results to the caller using the
 `on_return` callback that was supplied by the caller to `canon_lift`. Using a
 callback instead of simply returning the values from `canon_lift` enables the
-callee to keep executing after returning its results. However, it does
-introduce a dynamic error condition if `canon task.return` is called less or
-more than once which must be checked by `return_` and `exit`.
+callee to keep executing after returning its results. There is a dynamic error
+if `task.return` is not called exactly once and if the callee has not dropped
+all borrowed handles by the time `task.return` is called. This means that the
+caller can assume that ownership all its lent borrowed handles has been
```
dicej, line 607:
```suggestion
caller can assume that ownership of all its lent borrowed handles has been
```

lukewagner:
@dicej Great point; I added a blurb to Async.md in [this commit](https://github.com/WebAssembly/component-model/pull/425/commits/44c112bd2786e8e4192376be19be9699b71c6de2).

Issue #395, opened 2024-09-16
Relax requirements on assigning integers to resource handles


alexcrichton:
Currently in the canonical ABI resources (`own<T>` and `borrow<T>`) are required to strictly follow the specification in terms of how integers are assigned to handles as they are created. Specifically it's required that engines implement per-resource tables with a slab allocation scheme which is a LIFO allocator for resource indices. In the context of [this discussion on Zulip](https://bytecodealliance.zulipchat.com/#narrow/stream/219900-wasi/topic/PSA.3A.20wasi.3Aio.400.2E2.2E0.20and.20wasi.3Aio.400.2E2.2E1.20don't.20mix.20well/near/469159187) though it's been found that while this is a compact indexing scheme it has a drawback of creating situations which can be difficult to debug. For example creating an initial resource A gives handle index 1, and then creating an initial resource B also gives handle index 1. If these handle indices are accidentally passed to a function that wants resource C then all the runtime can do is say "unknown handle index". The runtime doesn't know whether the "1" is of type A or of type B.

The specific confusing case in question that came up on Zulip was that `wasi:io/poll/pollable@0.2.1` was created with index 1 and then passed to a function that wanted `wasi:io/poll/pollable@0.2.0`. This ended up (rightfully) not working and the runtime raised a trap, but it was a difficult-to-debug situation to determine that this was happening.

In discussion with some folks we had the idea of possibly relaxing the requirements of exactly how indices are allocated. I believe our rough conclusion is that the spec algorithm would change to allocating a random, but unique, index per resource type. This is a change from the slab allocation of today to be random instead (but still unique).

This is a small and subtle change, but the intention is to enable this to preserve the best guarantees that both guests and hosts today have without actually breaking anything in practice. Notably:

* This enables true RNG generation of handle indices for "fuzzing" a component if desired. This can help weed out accidental mistakes in guests for example.
* This enables the host to have a single table of all types of resources. Because the spec says handle indices are random there's no reason that "random" can't mean "it's sequential" for example. This would help solve the above issue because with a single table of all resources the host could provide a better error message.
* All existing hosts which implement the component model already match this new specification. The "random" behavior just happens to look like a slab.

[...]

PR #427, opened 2024-12-05
Put resources handles into a single table per component instance, instead of per-resource-type-per-instance

lukewagner:
As discussed in #395.  This simplifies the implementation (avoiding likely subtle incompatibilities) and helps developers catch and understand bugs while maintaining determinism.  In context, the overhead of the extra branch should be negligible.


sunfishcode:
It helps that this isn't exposed in the public interface of components, so if in the future we decide we want different behavior, we could have options.

sunfishcode (approved):

PR #428, opened 2024-12-10
fix the name of `thread.available_parallelism` in Binary.md

liigo:
see https://github.com/WebAssembly/component-model/pull/419


sunfishcode:
Oops, thanks!

PR #431, opened 2024-12-16
Change task.return to take component-level type + canonopt immediates

lukewagner:
In the original async PR (#363), there seemed to be a reason to reuse the same core wasm type-equality test used by `call_indirect` in `task.return` to avoid a new kind of (component-level) type equality test.  However, the streams PR (#405) ended up adding a well-motivated runtime component-type-equality test.  Without that reason, passing the component-level return type + `canonopt`s seems to be the more natural choice, so this PR switches to that.

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -351,9 +351,16 @@ replaced with `...` to focus on the overall flow of function calls.
 ```wat
 (component
   (import "fetch" (func $fetch (param "url" string) (result (list u8))))
+  (core module $Libc
+    (memory (export "mem") 1)
+    (func (export "realloc") (param i32 i32 i32 i32) (result i32) ...)
+    ...
+  )
   (core module $Main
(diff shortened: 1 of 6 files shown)


design/mvp/Async.md
```diff
@@ -351,9 +351,16 @@ replaced with `...` to focus on the overall flow of function calls.
 ```wat
 (component
   (import "fetch" (func $fetch (param "url" string) (result (list u8))))
+  (core module $Libc
+    (memory (export "mem") 1)
+    (func (export "realloc") (param i32 i32 i32 i32) (result i32) ...)
+    ...
+  )
```
lukewagner, line 358: This and most of the changes below in Async.md are pre-existing fixes that were made more evident when thinking about the `<canonopt>*` immediate of `(canon task.return ...)` below.

dicej (approved):
Looks great; thanks!

PR #440, opened 2025-01-16
Make future and stream element types optional

lukewagner:
For the same reason that `variant` and `result` case payloads and `func` results are optional, it seems like the element types of `future` and `stream` should be optional, allowing them to convey (single-shot or recurring) timing events.  When the element-type is absent, the `ptr` (read-from or written-to) is simply ignored (as if `elem_size(Unit) == 0`, but without the messy implications if we did this in general).  This is already present in the current implementation for `future` (but not yet `stream`), so this helps converge the spec with the impl.

(credit to @rvolosatovs for pointing out the omission)

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -738,7 +738,7 @@ variant-case payloads and function results, when `T` is absent, the "value(s)"
 being asynchronously passed can be thought of as [unit] values. In such cases,
 there is no representation of the value in Core WebAssembly (pointers into
 linear memory are ignored) however the *timing* of completed reads and writes
-is observable and meaningful. Thus, empty futures and streams can be useful for
+and the number of elements they contain are observable and meaningful. Thus, empty futures and streams can be useful for
 timing-related APIs.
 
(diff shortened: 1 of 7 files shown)


design/mvp/Explainer.md
```diff
@@ -733,6 +733,14 @@ futures are useful in more advanced scenarios where a parameter or result
 value may not be ready at the same time as the other synchronous parameters or
 results.
 
+The `T` element type of `stream` and `future` is an optional `valtype`. As with
+variant-case payloads and function results, when `T` is absent, the "value(s)"
+being asynchronously passed can be thought of as [unit] values. In such cases,
+there is no representation of the value in Core WebAssembly (pointers into
+linear memory are ignored) however the *timing* of completed reads and writes
+is observable and meaningful. Thus, empty futures and streams can be useful for
```
dicej, line 741:
```suggestion
and the number of elements they contain are observable and meaningful. Thus, empty futures and streams can be useful for
```

dicej (approved):

rvolosatovs (approved):

Issue #235, opened 2023-08-18
Resource properties


badeend:
Most languages have either:
- language-level support for properties
    - C#: `string MyProperty { get; set; }`
    - JavaScript: `get myProperty()`
    - JavaScript: `set myProperty(value)`
- or conventions on how to name property-like methods.
    - Java: `String getMyProperty()`
    - Java: `void setMyProperty(String value)`
    - Rust: `my_property(&self) -> String`
    - Rust: `set_my_property(&mut self, value: String) -> ()`

If properties are not encoded natively in WIT, we'll probably end up with everybody choosing their own convention (based on their own language). Which is bound to end up looking foreign to the other languages.
Or we can prescribe a convention for tooling to convert from&to, but at that point we might as well just encode it into Wit itself.

My suggestion:

```wit
interface abc {
    resource blob {
        content-type1: string { get } // Read-only property
        content-type2: string { set } // Write-only property
        content-type3: string { get, set } // Read/write property

        content-type4: string // TODO: Without annotation defaults to { get } ? Or { get, set } ?

        content-type5: string { // Customized signatures
            get -> result<string, my-get-error>,
            set -> result<_, my-set-error>,
        }
        content-type6: string { // Infallible getter, fallible setter
            get, // If no custom signature is provided, the getter/setter is assumed to be infallible.
            set -> result<_, my-set-error>,
        }

        /// General doc-comment applicable to both the getter & setter.
        content-type7: string {
            /// Documentation specific to the getter.
            get,

            /// Documentation specific to the setter.
            set,
        }
    }

    // Is equivalent to:



    resource blob
    %[get]blob.content-type1: func(self: borrow<blob>) -> string
    %[set]blob.content-type2: func(self: borrow<blob>, value: string)
    %[get]blob.content-type3: func(self: borrow<blob>) -> string
    %[set]blob.content-type3: func(self: borrow<blob>, value: string)

    // %[???]blob.content-type4

    %[get]blob.content-type5: func(self: borrow<blob>) -> result<string, my-get-error>
    %[set]blob.content-type5: func(self: borrow<blob>, value: string) -> result<_, my-set-error>
    %[get]blob.content-type6: func(self: borrow<blob>) -> string
    %[set]blob.content-type6: func(self: borrow<blob>, value: string) -> result<_, my-set-error>

}
```

Or more Javascript-esque:
```wit
interface abc {
    resource blob {
        // Together these two declarations represent one logical property.
        content-type: get func() -> string
[...]

PR #442, opened 2025-01-19
Add an 'async' hint to functions

lukewagner:
This PR adds `non-blocking` to `functype` (both in WAT and WIT).  It's a pretty small addition and doesn't touch validation/runtime.

While it initially seemed attractive to enforce `non-blocking` with trap-on-block semantics, there are valid scenarios where a callee might actually want/need to block and where the loss of concurrency in the caller is fine.  Thus the PR proposes making `non-blocking` just a hint (ignored by validation/runtime) to inform bindings generation (e.g., allowing bindings generators that make all functions `async` by default emit non-`async` functions for `non-blocking`).

`constructor` implies `non-blocking` (since `new` expressions in most languages can't be async).  However, to avoid breaking wasip2, we don't (yet) require validation of `[constructor]`-named functions to contain `non-blocking` (adding this to the list of warts to remove in the next breaking change).  In any case, bindings generators can always just take the non-blocking hint directly from seeing `[constructor]`.

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -154,7 +154,40 @@ component-level function that has been [lifted] from Core WebAssembly with the
 function that does not have the `async` option set (which is the default and
 only option prior to Preview 3). Thus, the sync/async distinction appears
 only independently in how a component-level function is *implemented* or
-*called*.
+*called*. This lack of distinction helps to avoid the classic ["What Color Is
+Your Function?"][Color] problem.
+
+Functions *may* however be annotated (in both WIT and component binaries) with
+`async` as a *hint*. This hint is intended to inform the default language
+bindings generation, indicating whether to use a source-level `async` function
+or not in languages that have such a distinction (e.g., JS, Python, C# and
+Rust). In the absence of such a hint, a bindings generator would be forced to
+make a uniform decision for what to do by default for all functions or require
+manual programmer directives. However, because `async` is just a hint, there is
+no prohibition against non-`async`-exported functions calling imported `async`
+functions. This does mean that non-`async` functions may end up blocking
+their caller, but (1) any loss in performance is the callee's "fault", (2) the
+caller can still lower `async` if they want to (overriding the default hint),
+(3) any *transitive* caller an lower `async` to avoid blocking.
+
+For example, given this interface:
+```wit
+interface filesystem {
+  resource file {
+    constructor();
+    is-closed: func() -> bool;
+    read: async func(num-bytes: u32) -> result<list<u8>>;
+    from-bytes: static func(bytes: list<u8>) -> file;
+    from-stream: static async func(bytes: stream<u8>) -> file;
+  }
+}
+```
+a bindings generator in a language with `async` would only emit `async`
+functions for `read` and `fetch`. Since in many languages `new` expressions
+cannot be async, there is no `async constructor`. Use cases requiring
+asynchronous construction can instead use `static async` functions, similar to
+`from-stream` in this example.
+
 
 ### Task
 
@@ -709,11 +742,9 @@ time as applications are built with more components.
 ## TODO
 
 Native async support is being proposed incrementally. The following features
-will be added in future chunks roughly in the order list to complete the full
+will be added in future chunks roughly in the order listed to complete the full
 "async" story, with a TBD cutoff between what's in [WASI Preview 3] and what
 comes after:
-* `nonblocking` function type attribute: allow a function to declare in its
-  type that it will not transitively do anything blocking
 * `subtask.cancel`: allow a supertask to signal to a subtask that its result is
   no longer wanted and to please wrap it up promptly
 * zero-copy forwarding/splicing
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -2191,9 +2191,12 @@ importname    ::= <exportname>
                 | <urlname>
                 | <hashname>
 plainname     ::= <label>
+                | '[async]' <label> 🔀
                 | '[constructor]' <label>
                 | '[method]' <label> '.' <label>
+                | '[async method]' <label> '.' <label> 🔀
                 | '[static]' <label> '.' <label>
+                | '[async static]' <label> '.' <label> 🔀
 label         ::= <fragment>
                 | <label> '-' <fragment>
 fragment      ::= <word>
@@ -2307,16 +2310,24 @@ The `plainname` production captures several language-neutral syntactic hints
 that allow bindings generators to produce more idiomatic bindings in their
 target language. At the top-level, a `plainname` allows functions to be
 annotated as being a constructor, method or static function of a preceding
-resource. In each of these cases, the first `label` is the name of the resource
-and the second `label` is the logical field name of the function. This
-additional nesting information allows bindings generators to insert the
-function into the nested scope of a class, abstract data type, object,
-namespace, package, module or whatever resources get bound to. For example, a
-function named `[method]C.foo` could be bound in C++ to a member function `foo`
-in a class `C`. The JS API [below](#JS-API) describes how the native JavaScript
-bindings could look. Validation described in [Binary.md](Binary.md) inspects
-the contents of `plainname` and ensures that the function has a compatible
-signature.
+resource and/or being asynchronous.
+
+When a function is annotated with `constructor`, `method` or `static`, the
+first `label` is the name of the resource and the second `label` is the logical
+field name of the function. This additional nesting information allows bindings
+generators to insert the function into the nested scope of a class, abstract
+data type, object, namespace, package, module or whatever resources get bound
+to. For example, a function named `[method]C.foo` could be bound in C++ to a
+member function `foo` in a class `C`. The JS API [below](#JS-API) describes how
+the native JavaScript bindings could look. Validation described in
+[Binary.md](Binary.md) inspects the contents of `plainname` and ensures that
+the function has a compatible signature.
+
+When a function is annotated with `async`, bindings generators are expected to
+emit whatever asynchronous language construct is appropriate (such as an
+`async` function in JS, Python or Rust). Note the absence of
+`[async constructor]`. See the [async
+explainer](Async.md#sync-and-async-functions) for more details.
 
 The `label` production used inside `plainname` as well as the labels of
 `record` and `variant` types are required to have [kebab case]. The reason for
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -32,6 +32,7 @@ more user-focused explanation, take a look at the
   * [Value definitions](#-value-definitions)
   * [Start definitions](#-start-definitions)
   * [Import and export definitions](#import-and-export-definitions)
+    * [Name uniqueness](#name-uniqueness)
 * [Component invariants](#component-invariants)
 * [JavaScript embedding](#JavaScript-embedding)
   * [JS API](#JS-API)
@@ -782,8 +783,8 @@ The remaining 4 type constructors in `deftype` use `valtype` to describe
 shared-nothing functions, resources, components, and component instances:
 
 The `func` type constructor describes a component-level function definition
-that takes a list of uniquely-named `valtype` parameters and optionally returns
-a `valtype`.
+that takes a list of `valtype` parameters with [strongly-unique] names and
+optionally returns a `valtype`.
 
 The `resource` type constructor creates a fresh type for each instance of the
 containing component (with "freshness" and its interaction with general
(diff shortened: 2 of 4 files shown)


design/mvp/Async.md
```diff
@@ -151,7 +151,34 @@ component-level function that has been [lifted] from Core WebAssembly with the
...
+```wit
+interface filesystem {
+  resource file {
+    constructor();                                  // non-blocking by default
+    is-closed: non-blocking func() -> bool;         // explicitly non-blocking
+    read: func(num-bytes: u32) -> result<list<u8>>; // blocking by default
+```
+If a resource-type has a potentially-blocking constructor, it can simply use
+`static new: func(...) -> my-resource` instead; `constructor` has no advantages
+beyond more-idiomatic bindings generation in some languages.
```
sunfishcode, line 180: From a new Wit user perspective. the name "non-blocking" could sound like it means "no stopping the caller" rather than "no waiting for I/O", which would be confusing since it does the exact opposite of that :-}. At least we should clearly document the Wit-user-facing side of this keyword.

design/mvp/Explainer.md
```diff
@@ -783,7 +784,11 @@ shared-nothing functions, resources, components, and component instances:
 
 The `func` type constructor describes a component-level function definition
 that takes a list of uniquely-named `valtype` parameters and optionally returns
-a `valtype`.
+a `valtype`. Function types can optionally be annotated with a `non-blocking`
+attribute which has no semantic effect and is ignored at validation- and
+run-time, serving primarily as a hint that tells bindings generators to lift
```
sunfishcode, line 789: Is the "primarily" here redundant?

design/mvp/WIT.md
```diff
@@ -1295,7 +1296,7 @@ typedef-item ::= resource-item
 
 func-item ::= id ':' func-type ';'
 
-func-type ::= 'func' param-list result-list
```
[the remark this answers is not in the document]
lukewagner: Good idea, [fixed](https://github.com/WebAssembly/component-model/pull/442/commits/c81fbb2af4518d83a226961498fc085812107b55), PTAL

design/mvp/Async.md
lukewagner: Yeah, I suppose "blocking" does have multiple interpretations.  I added [this addition](https://github.com/WebAssembly/component-model/pull/442/commits/6de7a0ca4e00110c4b235c8e5a900c89e27c48e2) to the previous note, but it could probably be improved.

design/mvp/Explainer.md
lukewagner: Yeah, I suppose so.  It felt a bit off without any word, so I tried "only", and also tightened up the rest of the sentence in [this commit](https://github.com/WebAssembly/component-model/pull/442/commits/5920c974a504cb1dce51ec6ac9b1dd2986cd1822), PTAL

badeend:
Since this is just a hint for the binding generators and not actually enforced at runtime: as a WIT author I'd rather prefer the defaults to be swapped. I.e. functions should be generated as `non-blocking` by default unless told otherwise.

As an example I annotated the wasi-sockets proposal both defaults:
- With "blocking" as the default, it needs [34 annotations](https://github.com/badeend/wasi-sockets/commit/f67c4cda77212e00fb7edc46eb77475f3f249870)
- With `non-blocking` as the default, it needs [just 5](https://github.com/badeend/wasi-sockets/commit/e2ff81d262ab2f30ecb8a27ae594f47db3ded6c6).

Aside from the amount of work it is for me (which I can live with :P ), annotating only the "blocking" functions is more familiar to readers who have experience with `async/await` languages, since the concept maps pretty much 1:1.

lukewagner:
Scanning that wasi-sockets PR, 28 of the 34 `non-blocking` annotations would be removed by WIT getter/setter syntax.

For the case of `connect`, it seems like this should be blocking (as is, e.g., the Node.js API).  For the case of `disconnect`, I may not understand this one fully, but the docs say it's equivalent to `connect()` with `AF_UNSPEC`, and since `connect()` is async, it seems like it should be blocking too?

For the case of `bind` (x2) and `listen`, I see there are corresponding `IORING_OP_*`s for these, so if one was using an io_uring-based implementation, there could hypothetically be concurrency wins from doing these async.  In general, I've been imagining that the rough (but not hard) criteria for "should it be blocking?" is "does it require a syscall?" since this where a bunch of potentially-blocking and concurrent things can happen.

But for the last case, `receive`, this does seem like a case we might see more of where the returned `tuple<stream<u8>, future<...>>` (perhaps one day soon condensed down to a `stream<u8, _, error-code>` or something like that) captures all the asynchrony and so the `non-blocking` looks superfluous.  One idea is that, in the same way as "`constructor`s are implicitly `non-blocking`", we could say that "functions returning only `future`s+`stream`s are implicitly `non-blocking`" too.

badeend:
Regarding `connect`/`disconnect`: I suspect you're looking at the UDP variants. Those do not perform any IO, so their annotations are correct. Their unfortunate names are explained [here](https://github.com/WebAssembly/wasi-sockets/blob/8069eb996518cd064311c1d77677b708430aa6be/wit-0.3.0-draft/types.wit#L551-L554).

As for `bind` (2x) & `listen`: those do not perform any IO either. It appears their io_uring opcodes were added [recently](https://patchwork.kernel.org/project/io-uring/patch/20240614163047.31581-3-krisman@suse.de/). IIUC, the reason for inclusion in io_uring isn't to do with I/O but rather for interoperability with io_uring's "registered files" feature. My knowledge of io_uring is not sufficient enough to know how this would impact the WASI interface (if at all).

---

> I've been imagining that the rough (but not hard) criteria for "should it be blocking?" is "does it require a syscall?"

I see your point (and the "but not hard" caveat), but I don't know how to extrapolate that guidance:
- All sockets options are syscalls too, so following that logic would imply they should all be `blocking` methods instead of getters/setters.
- From the component's POV, arent't _all_ imports effectively "syscalls"?

---

Having native syntax for getters & setters could indeed help alleviate many of the annotations in wasi-sockets. As you already counted out, that would bring the amount of needed annotations down from 34 to just 6. Versus 5. So: potáto, potàto. :upside_down_face:

My other point still stands though; this is a syntax is directed at binding generators / language implementors. IMO it still makes sense to speak _"their"_ language, not "ours". From a JavaScript/Python/Rust/C#/.. developer's POV, seeing an `async` annotation is instantly familiar and (aside from an edge case here and there) will probably map 1:1 to `async` in the generated bindings for their language. That being said, I realize this is purely syntactical bikeshedding so please don't block the PR on this, as I fully agree with the semantics. :+1:

lukewagner:
@badeend Good points!

Revisiting the case of `receive` with fresh eyes (and anticipating more like it), the workaround I suggested above seems too implicit; a "`blocking` implies you get `async`" rule does seem much simpler and, to your point, "speaking their language".  So despite starting from an initial position that we should [async everything](https://joeduffyblog.com/2015/11/19/asynchronous-everything/), I'm starting to come around to your (and Joel's) POV that maybe `non-blocking` is the better default and `blocking` should get the explicit opt-in.

Happy to hear more thoughts on this!

lukewagner:
Polling more folks during the BA component-model implementation meeting, lots of agreement on switching the default.  But the meeting also highlighted that @sunfishcode was right [above](https://github.com/WebAssembly/component-model/pull/442#discussion_r1924380183) in saying that "blocking" can naturally be interpreted to have the opposite meaning that I'm assuming in the PR.  Thinking of what a good short intuitive alternative keyword is, `async` seems like the obvious choice.  Of course some languages (like Go) don't have `async`, but these same languages are ignoring this hint anyways (at least in generated bindings; a Go programmer may still benefit from seeing the hint to know that they might want to `go call()` it instead of just `call()`ing it).

Thus, in summary, it seems like a good idea to default to synchronous bindings with an explicit opt-in via `async`.  I'll update the PR in a bit, and happy to hear more thoughts.

lukewagner:
I also like the idea of "async by default", but the monkey wrench for me was cases like `receive`, which return a `stream` or `future`, and where it feels silly to have the generated `async` bindings force you to asynchronously wait to get that stream/future (which you then asynchronously wait on again).  But then writing `non-blocking` (or whatever we want to call it) on `listen` looks confusing too (b/c the overall *operation* is blocking, even if the initial *call* isn't).  And my workaround (of special-casing functions-returning-futures/streams) seemed too subtle and also brittle.

Also, while I also worry about `async` making people think we have the "what color is your function?" problem, I expect it won't be such a problem as to cause people to bounce off it entirely (before they learn that it's just a hint and thus not a problem).

cpetig:
Given that the new opt-in keyword (`blocking` or `async`) would change behavior at the ABI level, and in this case the no-keyword case would remain ABI identical to 0.2 while `async` introduces more call overhead, I feel that making non-async the default results in less surprises.

I was enthusiastic about an async ABI default as well, but it clearly seems the less frequent choice once `future` and `stream` return types are available.

sunfishcode:
> Given that the new opt-in keyword (`blocking` or `async`) would change behavior at the ABI level, and in this case the no-keyword case would remain ABI identical to 0.2 while `async` introduces more call overhead, I feel that making non-async the default results in less surprises.

Ah, that's true.

Initially I was thinking that the alternative would be to just have bindings generators default to sync bindings. That would also eliminate the same surprises. But the place where this breaks is if you have a mix of p2 and p3 APIs, and you want to use async bindings for the p3 APIs, but that would make all your p2 APIs use async bindings too, which they likely weren't designed for.

In other words, we have this table:

| behavior | p2 | p3
| --- | ---- | ---- |
| blocking | use a pollable etc. | `blocking` |
| non-blocking|default | `non-blocking` |

Which shows how if we made `blocking` the default in p3, it would conceptually represent a change in the default.

> I was enthusiastic about an async ABI default as well, but it clearly seems the less frequent choice once `future` and `stream` return types are available.

I don't think `future` reduces the need for `blocking` though. In this code:

```wit
foo: blocking func() -> s32;
foo: non-blocking func() -> futue<s32>;
```

The `blocking` version gives both caller and callee a choice of whether to use sync or async bindings. The `future` version essentially requires both to use async. So between these two I think we're usually going to prefer the `blocking`  version.

sunfishcode:
And following up on my comment [here](https://github.com/WebAssembly/component-model/pull/442#issuecomment-2635449908), since it makes sense for the default to be non-blocking, "shallow" is the wrong sense, and I agree that it makes sense to call the keyword "async" for familiarity.

lukewagner:
Ok, I updated the PR to flip defaults and use `async`, PTAL.

Working on this, I realized that there is a much better place to put a hint that doesn't affect static or dynamic semantics and only informs bindgen: in the import/export name string.  Thus, I updated the PR to encode the `async` hint as `[async]foo`, `[async method]R.foo` or `[async static]R.foo`.  The fact that you can't have an async constructor falls out from the fact that there is no `[async constructor]` allowed.  My impression is that this approach is also slightly easier for the toolchain, but let me know if not and happy to reconsider the "attribute on function type" approach if there are issues with doing it in the name.

alexcrichton (approved):
👍

alexcrichton:
From implementing this:

Today we disallow importing both `[method]foo.bar` and `[static]foo.bar`, and I think we'll want to extend that to async as well? My thinking is:

* For plain labels, you get at most one of `{"a", "[async]a"}`
* For methods/statics, you get at most one of `{"[method]a.b", "[async method]a.b", "[static]a.b", "[async static]a.b"}`

lukewagner:
Ah, great point, I forgot about that, and those restrictions make sense.  What if we took it one small step further and said simply "names must be unique within a scope after stripping any `[...]` annotation prefix"?  The cases this additionally rules out are `{"a", "[constructor]a"}` and `{"[async]a","[constructor]a"}`, which actually seems useful (for languages that use the type-name as the constructor-name).  WDYT?

badeend:
> What if we took it one small step further and said simply "names must be unique within a scope after stripping any [...] annotation prefix"?

That seems like the safest option to me. 👍 That would also automatically work for any new annotation (e.g. [`[get]`/`[set`](https://github.com/WebAssembly/component-model/issues/235)). We can always define exceptions to this rule if a use-case ever pops up.

alexcrichton:
I think that rule might run afoul of making this currently valid component invalid though?

```wasm
(component
  (import "a" (type $a (sub resource)))
  (import "[constructor]a" (func (result (own $a))))
)
```

lukewagner:
Ah right, thanks; I forgot about the type import.  Could we say "names must be unique in a scope after stripping `[...]`, with the exception that `a` and `[constructor]a` are allowed"?

alexcrichton:
Sounds reasonable to me 👍

lukewagner:
Ok, rebased to avoid conflicts, and I added [this commit](https://github.com/WebAssembly/component-model/pull/442/commits/69676201180663865f006c67a89cf4cc899b439e) to cover naming uniqueness conditions.  This commit factors out several scattered bits of uniqueness requirements into a new section in Explainer.md that defines the term "strongly-unique" that attempts to cover what we just discussed.  PTAL!

PR #443, opened 2025-01-21
Describe how start works in the context of async

lukewagner:
This PR fills in a TODO section in Async.md about what happens if an `async`-lifted function is used as a `start` function.  It's pretty short, because it mostly just falls out from how async already works (but let me know if I'm forgetting to cover something).

To summarize, the idea is that `start` acts like a synchronous call, waiting until the `async` callee `task.return`s its value before continuing instantiation.  But since `async` functions can keep running after `task.return`, so can these `start`-created tasks, "in the background".  This appears to be quite useful as a solution for how language toolchains can support "background task"-type use cases without throwing out structured concurrency or the async call tree.

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -23,8 +23,8 @@ summary of the motivation and animated sketch of the design in action.
   * [Backpressure](#backpressure)
   * [Returning](#returning)
 * [Examples](#examples)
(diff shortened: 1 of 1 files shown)


lukewagner:
Merging based on no comments, but we can revisit in issues as this gets implemented.

Issue #441, opened 2025-01-19
How to implement non-blocking POSIX `send` in terms of `stream.write`?


badeend:
Am I correct to say that: when `stream.write` returns `BLOCKED`, it continues to have access to the provided memory buffer until the write either finishes or is canceled?
If so, does this mean wasi-libc has to perform an intermediate copy into a private socket-level buffer first for every `send`/`recv` on non-blocking sockets? (in order to emulate the "readiness"-based POSIX  async model)

dicej:
Great question.  I believe @lukewagner's plan was to allow zero-length `stream.read` and `stream.write` calls, which may be used to query whether the other of the stream is ready.  So a POSIX `send` call would first issue a zero-length `stream.write`, and if it returns `BLOCKED`, cancel it and return `EAGAIN`/`EWOULDBLOCK`.  Otherwise, issue a non-zero length `stream.write`, which _should_ _not_ return `BLOCKED` (but which can be cancelled if it does).

However, this could lead to a livelock-type situation if both the read and write ends are repeatedly probing for readiness using zero-length calls, but failing to line up since both of them are canceling any call that returns `BLOCKED`.  The only way to avoid that is if at least one side is willing to let a `BLOCKED` call complete.  @lukewagner how do you think that would work if both ends are using POSIX `send`/`recv`?

lukewagner:
Yes, relaxing the current length restrictions on `stream.{read,write}` so that zero-length signal "readiness" of the other side is the plan.  I'm mostly just waiting for someone to get enough into the weeds of libc to confirm that the approach makes sense.

To avoid the livelock scenario @dicej mentions, the general rule needs to be that, if you are using a zero-length read/write to probe readiness, once you hear back that "the other side is ready", you must follow that up with a non-zero-length read/write that you allow to complete (either by calling `stream.{read,write}` synchronously, which I think would be the preferred option for libc, or calling it asynchronously, and *not* cancelling if the return value is `BLOCKED`).

If you apply these rules to a vanilla scenario where there is a stream between two components that are both applying this non-blocking I/O technique, then you'd get the following sequence, which I think is what you'd expect:
* The first component, call it A, tries to do a non-blocking `write()` on a file-descriptor which (upon creation) has the "ready" bit unset, and so libc makes a zero-length `stream.write`, which returns `BLOCKED`, so libc returns `EWOULDBLOCK` to the caller.  A does other things until eventually waiting on a libc `select()` which proceeds to block on `task.wait`.
[...]

PR #444, opened 2025-01-26
Allow stream.{read,write}s of length 0 to query/signal readiness

lukewagner:
This PR relaxes the rules for `stream.{read,write}` to accept lengths of `0`.  This can be useful for signalling and querying readiness which in turn can be used to implement non-blocking POSIX operations, as discussed in #441.

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -356,7 +356,11 @@ These built-ins can either return immediately if >0 elements were able to be
 written or read immediately (without blocking) or return a sentinel "blocked"
 value indicating that the read or write will execute concurrently. The readable
 and writable ends of streams and futures can then be [waited](#waiting) on to
-make progress.
+make progress. Notification of progress signals *completion* of a read or write
+(i.e., the bytes have already been copied into the buffer). Additionally,
+*readiness* (to perform a read or write in the future) can be queried and
+signalled by performing a `0`-length read or write (see the [Stream State]
+section in the Canonical ABI explainer for details).
 
 As a temporary limitation, if a `read` and `write` for a single stream or
 future occur from within the same component, there is a trap. In the future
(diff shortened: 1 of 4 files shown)


lukewagner:
Updated the PR based on further discussion in #441 about what happens when a `0`-length read and write rendezvous.  In the previous iteration, both were notified; with this update, only the writer is notified, since this ends up being what you want if there is going to be buffering (as mentioned in the PR and also [this comment](https://github.com/WebAssembly/component-model/issues/441#issuecomment-2622661035)).

alexcrichton:
cc @dicej on this as it relates to our discussion this morning. It looks like the wasip3-prototyping branch doesn't implement the spec as-is as 0-length reads/writes aren't trapped, and it also hasn't been updated to take this PR into account either. Basically I think wasip3-prototyping is an ambiguous state right now as to what 0-length reads/writes mean.

Also question for you @lukewagner. The topic of 0-length reads/writes have come up in the context of figuring out what the result of an async stream operation is. Specifically I was attempting to model the result of a stream operation as a tri-state value of either (a) completed with N items, (b) closed, or (c) cancelled. Given the way the return codes map here though is it correct to say that if you cancel a 0-length operation there's no way to figure out what happened as a result? The canonical ABI code returned will be 0, maybe eventually, and that's ambiguous whether the read/write was cancelled or "completed"?

lukewagner:
You're right, as currently written, after a `BLOCKED` cancellation, the event tuple will be the same as-if the operation completed.  I wasn't sure whether this was a problem due to the fact that the cancellation comes from within the same component that receives the event, but I guess in the `BLOCKED` case there is a race and so this could actually lose useful information.  I already think we need to slightly change the packed result of a stream operation to have a 4-bit "what happened" code separate from the 28-bit optional "payload", and so this could be fit in with that change.

dicej (approved):

lukewagner:
Sounds like this has been implemented in [wasip3-prototyping/#130](https://github.com/bytecodealliance/wasip3-prototyping/pull/130) so I'll merge and we can iterate as necessary in follow-up issues and PRs.

PR #447, opened 2025-02-07
Add `thread.spawn_indirect`

abrown:
This change codifies the conclusions we arrived to in [#89]. It adds a new way to spawn threads, `thread.spawn_indirect`, which retrieves the thread start function from a table. This prompted me to rename `thread.spawn` to `thread.spawn_ref`.

[#89]: https://github.com/WebAssembly/shared-everything-threads/issues/89

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -286,9 +286,10 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift 
            | 0x01 0x00 f:<funcidx> opts:<opts>                   => (canon lower f opts (core func))
            | 0x02 rt:<typeidx>                                   => (canon resource.new rt (core func))
            | 0x03 rt:<typeidx>                                   => (canon resource.drop rt (core func))
-           | 0x07 rt:<typdidx>                                   => (canon resource.drop rt async (core func)) 🔀
+           | 0x07 rt:<typeidx>                                   => (canon resource.drop rt async (core func)) 🔀
            | 0x04 rt:<typeidx>                                   => (canon resource.rep rt (core func))
-           | 0x05 ft:<typeidx>                                   => (canon thread.spawn ft (core func)) 🧵
+           | 0x05 ft:<typeidx>                                   => (canon thread.spawn_ref ft (core func)) 🧵
+           | 0x24 ft:<typeidx> t:<core:tabidx>                   => (canon thread.spawn_indirect ft (table t) (core func)) 🧵
            | 0x06                                                => (canon thread.available_parallelism (core func)) 🧵
            | 0x08                                                => (canon backpressure.set (core func)) 🔀
            | 0x09 rs:<resultlist> opts:<opts>                    => (canon task.return rs opts (core func)) 🔀
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -289,7 +289,7 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift 
            | 0x07 rt:<typeidx>                                   => (canon resource.drop rt async (core func)) 🔀
            | 0x04 rt:<typeidx>                                   => (canon resource.rep rt (core func))
            | 0x05 ft:<typeidx>                                   => (canon thread.spawn_ref ft (core func)) 🧵
-           | 0x24 ft:<typeidx> t:<core:tabidx>                   => (canon thread.spawn_indirect ft (table t) (core func)) 🧵
+           | 0x24 ft:<typeidx> t:<core:tableidx>                 => (canon thread.spawn_indirect ft (table t) (core func)) 🧵
            | 0x06                                                => (canon thread.available_parallelism (core func)) 🧵
            | 0x08                                                => (canon backpressure.set (core func)) 🔀
            | 0x09 rs:<resultlist> opts:<opts>                    => (canon task.return rs opts (core func)) 🔀
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -288,9 +288,6 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift 
            | 0x03 rt:<typeidx>                                   => (canon resource.drop rt (core func))
            | 0x07 rt:<typeidx>                                   => (canon resource.drop rt async (core func)) 🔀
            | 0x04 rt:<typeidx>                                   => (canon resource.rep rt (core func))
-           | 0x05 ft:<typeidx>                                   => (canon thread.spawn_ref ft (core func)) 🧵
-           | 0x24 ft:<typeidx> t:<core:tableidx>                 => (canon thread.spawn_indirect ft (table t) (core func)) 🧵
-           | 0x06                                                => (canon thread.available_parallelism (core func)) 🧵
            | 0x08                                                => (canon backpressure.set (core func)) 🔀
            | 0x09 rs:<resultlist> opts:<opts>                    => (canon task.return rs opts (core func)) 🔀
            | 0x0a 0x7f i:<u32>                                   => (canon context.get i32 i (core func)) 🔀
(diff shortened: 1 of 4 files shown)


design/mvp/CanonicalABI.md
```diff
@@ -3808,6 +3808,48 @@ def canon_thread_spawn(f, c):
...
+### 🧵 `canon thread.spawn_indirect`
+
+For a canonical definition:
+```wat
+(canon thread.spawn_indirect (type $ft) (table $t) (core func $st))
+```
+validation specifies:
+* `$ft` must refer to a `shared` function type; initially, only the type `(func
+  shared (param $c i32))` is allowed (see explanation in `thread.spawn_ref`
+  above)
```
abrown, line 3820: What I can't remember well is whether we truly need `$ft` in the definition. I left it in for now, but it would seem that without it we can still statically verify from the type of `$t` that the table elements are in fact shared function references.

abrown: Implementation-wise it is `$ft` that is problematic currently, so at least temporarily it isn't to useful anyways, seeing as we must fix the start function to only `[i32] -> []`.

lukewagner (commented):
Looks generally good.  One initial suggestion:

design/mvp/CanonicalABI.md
```diff
@@ -3808,6 +3808,48 @@ def canon_thread_spawn(f, c):
...
+* `$t` must refer to a table containing `ft`-typed items
```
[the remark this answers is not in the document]
lukewagner: Or even simpler: since the goal is to implement `pthread_create`, which takes an `i32` index to the same `funcref` table used by `call_indirect`, maybe could just remove `$t`, with validation requiring that `tableidx` must refer to a `(table funcref)`.  With `thread.spawn_ref`, someone can always implement typed function reference table dispatch via `table.get`+`call_ref`, so it doesn't seem like we need the midpoint between these two.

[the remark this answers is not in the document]
abrown: Looking back at the summary of the discussion from the threads subgroup, I see that:

- I promised to try to keep `$ft` or something like it around [here](https://github.com/WebAssembly/shared-everything-threads/issues/89#issuecomment-2532674895):

  > the "spawn indirect" intrinsic will still have a type parameter describing the signature of the start function

- and this is what @rossberg seemed fine with [here](https://github.com/WebAssembly/shared-everything-threads/issues/89):

  > having spawn_indirect with a restricted signature is okay with me, as long as it remains forward-compatible to general type signatures

Since we need to spawn from tables of shared `funcref` we can temporarily restrict `$ft` to just `(shared (func (i32))`.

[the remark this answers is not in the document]
lukewagner: Ah, I see, thanks!  So the `$ft` would serve a similar role as in `call_indirect`: it's for dynamically checking the type of the callee, not typing the table.  In that case, it makes sense to keep `$ft` and require it to be exactly `(shared (func (param i32)))` as you said (in conjunction with specifying that `$t` must have type `(table (ref null (shared func)) shared)`).

design/mvp/CanonicalABI.md
```diff
@@ -3820,6 +3820,48 @@ def canon_thread_spawn(f, c):
...
+  - invokes `$f` with `$c`
+  - executes `$f` until completion or trap in a `shared` context as described by
+    the [shared-everything threads] proposal.
+
+In pseudocode, `$st` looks like:
+
+```python
+def canon_thread_spawn_indirect(t, i, c):
+  trap_if(t[i] is None)
+  f = t[i]
```
abrown, line 3848: What I don't know how to express here in this Python code is "check that `f` is a reference of the right type."

abrown:
@lukewagner, I had to rebase due to some conflicts but the follow-on commits should have the changes we discussed.

lukewagner (approved):
Great, thanks!  This looks good to me with these suggestions:

design/mvp/CanonicalABI.md
```diff
@@ -3820,6 +3820,48 @@ def canon_thread_spawn(f, c):
 ```
 
 
+### 🧵 `canon thread.spawn_indirect`
+
+For a canonical definition:
+```wat
+(canon thread.spawn_indirect (type $ft) (table $t) (core func $st))
```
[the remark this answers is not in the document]
abrown: I agree on the `$ft` part but over in `wasm-tools` it would seem that I need to write out something like the following to show the builtin how to find the index: `(table $libc "start-table")`. Is there some other way to express that?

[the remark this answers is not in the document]
abrown: Ok, makes sense; thanks!

design/mvp/CanonicalABI.md
```diff
@@ -3773,23 +3774,23 @@ async def canon_error_context_drop(task, i):
 ```
 
 
-### 🧵 `canon thread.spawn`
+### 🧵 `canon thread.spawn_ref`
 
 For a canonical definition:
 ```wat
-(canon thread.spawn (type $ft) (core func $st))
+(canon thread.spawn_ref (type $ft) (core func $spawn_ref))
```
abrown, line 3781: Should this syntax also change?

```suggestion
(canon thread.spawn_ref $ft (core func $spawn_ref))
```

abrown: Well, I went ahead and made this change in 1c558c7; let me know if that's unnecessary. Otherwise, I think this is ready for one more look-through.

lukewagner: Yes, you're right, good catch!

design/mvp/Binary.md
```diff
@@ -318,6 +316,9 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift
            | 0x21 async?:<async>? m:<core:memidx>                => (canon waitable-set.poll async? (memory m) (core func)) 🔀
            | 0x22                                                => (canon waitable-set.drop (core func)) 🔀
            | 0x23                                                => (canon waitable.join (core func)) 🔀
+           | 0x40 ft:<typeidx>                                   => (canon thread.spawn_ref ft (core func)) 🧵
+           | 0x41 ft:<typeidx> t:<core:tableidx>                 => (canon thread.spawn_indirect ft (table t) (core func)) 🧵
```
lukewagner, line 320:
```suggestion
           | 0x41 ft:<typeidx> tbl:<core:tableidx>               => (canon thread.spawn_indirect ft tbl (core func)) 🧵
```

design/mvp/Explainer.md
```diff
@@ -1945,19 +1946,35 @@ thread management. These are specified as built-ins and not core WebAssembly
...
 
-(See also [`canon_thread_spawn`] in the Canonical ABI explainer.)
+(See also [`canon_thread_spawn_ref`] in the Canonical ABI explainer.)
+
+
+###### 🧵 `thread.spawn_indirect`
+
+| Synopsis                   |                                                   |
+| -------------------------- | ------------------------------------------------- |
+| Approximate WIT signature  | `func<FuncT>(i: i32, c: FuncT.params[0]) -> bool` |
```
lukewagner, line 1968:
```suggestion
| Approximate WIT signature  | `func<FuncT>(i: u32, c: FuncT.params[0]) -> bool` |
```

lukewagner (commented):
Last tiny nits:

design/mvp/CanonicalABI.md
```diff
@@ -3773,37 +3774,82 @@ async def canon_error_context_drop(task, i):
...
+For a canonical definition:
+```wat
+(canon thread.spawn_indirect $ft $tbl (core func $spawn_indirect))
+```
+validation specifies:
+* `$ft` must refer to a `shared` function type; initially, only the type
+  `(shared (func (param $c i32)))` is allowed (see explanation in
+  `thread.spawn_ref` above)
+* `$tbl` must refer to a table with type `(table (ref null (shared func))
+  shared)`
```
rossberg, line 3835: The table type doesn't seem to follow any syntax defined in the proposal. More importantly, we have subtyping. So I think this ought to say:
```sugggestion
* `$tbl` must refer to a shared table whose element type matches `(ref null (shared func))`
```

lukewagner: Ah, thanks, good point!

abrown: Updated in https://github.com/WebAssembly/component-model/pull/462.

PR #452, opened 2025-02-20
async: Fix typo in binary format of futures/streams

alexcrichton:
This commit fixes (what I believe is) a typo in the binary encoding of futures/streams. Previously they were specified as a `<typeidx>?` but I think these are supposed to be similar to `option<T>` where it's a `<valtype>?` instead.


alexcrichton:
(cc @dicej on this too)

dicej (approved):

lukewagner:
Oops, yes, thanks!

PR #455, opened 2025-02-24
Add a missing 🔀 on `resource.drop async`

alexcrichton:
Adding a missing indicator that this is part of an extension to the component model.


alexcrichton:
While here I wanted to double-check as well: the [validation](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#canon-resourcedrop) currently doesn't mention async beyond the definition itself, is that intentional? E.g. no signature changes or semantic changes in the Python given an `async` definition? (I was a bit surprised...)

alexcrichton:
> the [validation](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#canon-resourcedrop) currently doesn't mention async beyond the definition itself

Answering my own question, I should have been searching for `sync`, not `async`, so disregard me!

Issue #453, opened 2025-02-20
Clarifications/confusion about `task.return`


alexcrichton:
Looking at today's definition of [`task.return` in `CanonicalABI.md`](https://github.com/WebAssembly/component-model/blob/a9197ccf6530e99644ccf0421b5a6bb16f16d822/design/mvp/CanonicalABI.md#-canon-taskreturn) I'm left with a few questions I want to clarify:

* For the type of the function generated it takes `$opts` into account, but that means that the intrinsic can be declared as both async and sync, is that intended? Is there a use case for calling this asynchronously?
* For the runtime validation when the intrinsic is called it ensures that the `$opts` on `task.return` equals the options specified on the original `canon lift`, but is this what's intended when taking into account async/callbacks/etc? My (naive) expectation was that I could have an async/callback lifted export and be able to call this intrinsic synchronously, but as-written that means that the intrinsic would generate a trap I think?

cc @dicej

lukewagner:
Ah, great point!  I think what's missing is that we need to specify that validation only allows a *subset* of the `<opts>` and only this subset is required to equal the `<opts>` on the lift, and this subset would only include `memory`, `string_encoding` and `realloc`.

So we'd have something like:
```python
async def canon_task_return(task, result_type, opts, flat_args):
  ...
  trap_if(opts != only_return_opts(task.opts))
  ...
```
Does that sound right?

alexcrichton:
Ah ok yeah that makes sense to me. To confirm though it's intentional that this intrinsic shouldn't be allowed to have `async` on it?  That definitely simplifies it I think, but wanted to double-check

lukewagner:
Hopefully fixed in #456

PR #456, opened 2025-02-24
Tighten task.return rules

lukewagner:
As brought up and discussed in #453

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -111,7 +111,7 @@ object that is threaded through all the Python functions below as the `cx`
 parameter/field.
 ```python
 class LiftLowerContext:
-  opts: CanonicalOptions
+  opts: LiftLowerOptions
   inst: ComponentInstance
   borrow_scope: Optional[Task|Subtask]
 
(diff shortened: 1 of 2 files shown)


design/mvp/CanonicalABI.md
```diff
@@ -127,14 +127,33 @@ known to not contain `borrow`. The `CanonicalOptions`, `ComponentInstance`,
...
-immediates that can be passed to the `canon` definition being implemented.
+The `LiftLowerContext` class contains the subset of [`canonopt`] which are
+used to lift and lower the individual parameters and results of function
+calls:
 ```python
 @dataclass
-class CanonicalOptions:
+class LiftLowerOptions:
   memory: Optional[bytearray] = None
   string_encoding: Optional[str] = None
```
alexcrichton, line 137: This is completely unrelated to the change here, but I was looking over this again today and I was curious about the `Optional` here.  In the sense that `CanonicalOptions` represents a deserialized `vec(canonopt)` in the binary format this makes sense, but this has the interesting consequence of `None != Some('utf-8')` which seems counter-intuitive.

I also mistakenly thought that we'd validate that if a string were lifted/lowered that `string_encoding` must be present or otherwise it's a validation error, but at least today that's not the case and lift/lowers don't require `string-encoding`. Although that may be a mistake in wasmparser's validator too?

Basically, tl;dr; do you remember if it was intentional to allow `None` here for `string_encoding` when strings are actually lifted/lowered? If so it might be worth updating `__eq__` below to handle that. If not that's definitely something to track on `wasmparser`'s side... (and it'll take a good bit of time to fix I suspect)

alexcrichton (approved):

design/mvp/CanonicalABI.md
lukewagner: That's a good point; I agree we should remove the `Optional` here for `string_encoding` so that there is always a well-defined `string_encoding`.

At the binary format level, I suppose we could go either way in requiring it or saying "if it's not present, it defaults to utf8".  Until now, I had been assuming that the binary format worked like the text-format, where I think we do say that, if `string-encoding` is not explicitly present, the default is utf8.  Either way, it'd be good to document more-explicitly in CanonicalABI.md.

alexcrichton: It makes sense to me yeah to have the binary/text format be close to each other here, but the current state of the default being utf8 if not present feels like a situation we might have accidentally fell into rather than explicitly designed. I basically forget what the reason is, if any.

To me it seems reasonable to have a default, but it also seems equally reasonable to have a validation predicate that strings require a `string-encoding=...` canonical option, similar to how anything with a pointer requires `(memory ..)`. I realize though that memories are a "thing" which can't possibly have a default whereas string-encoding can reasonably have a default so not quite apples-to-apples but sort of close I figure.

All that being said if the reason for having a default string-encoding was "well we made a mistake early on and didn't add a validation predicate" that also seems reasonable. Possibly something we could track in a "future 1.0 breakage wishlist" though? (do we have one of those?)

lukewagner: I suppose one argument in favor of the current "have a default" option is that: let's say that in some post-1.0 future we add a new `canonopt` to backward-compatibly opt-in to a different ABI than 1.0.  In that case, we'd say that the absence of the `canonopt` defaulted to the 1.0 semantics.  So it's as-if we started with only utf8 and then added opt-in non-default choices.  And the reason why not for `realloc`/`memory` is that they're non-defaultable.

alexcrichton: Sounds reasonable to me 👍

PR #462, opened 2025-03-05
Fix language for `thread.spawn_indirect` table type

abrown:
This uses the language suggested [here] to document the table type needed for the `thread.spawn_indirect` canonical builtin.

[here]: https://github.com/WebAssembly/component-model/pull/447#discussion_r1980767772


abrown:
cc: @rossberg

rossberg (approved):

Issue #464, opened 2025-03-07
task.return: Clarifying `self.memory is other.memory`


alexcrichton:
The runtime validation of `task.return` checks that the canonical options are the same as the options used during the original lift (or at least the relevant subset). This includes the `memory` option and this is specified currently in Python as `self.memory is other.memory`. I wanted to clarify what we should do here:

* Is this a syntactic limitation? For example should the exact same memory index be used?
* Is this a runtime limitation? For example are the base pointers/lengths of the memories compared to ensure that they're the same?

For example:

```wasm
(component
  (core module $libc (memory (export "memory") 1))
  (core instance $libc (instantiate $libc))

  (alias core export $libc "memory" (core memory $m1))
  (alias core export $libc "memory" (core memory $m2))

  (core module $reexport
    (import "" "" (memory $m 1))
    (export "memory" (memory $m))
  )
  (core instance $reexport (instantiate $reexport
    (with "" (instance (export "" (memory $m1))))))
  (alias core export $reexport "memory" (core memory $m3))
)
```

If lift and `task.return` both use `$m1`, that seems naturally fine. If one uses `$m1` and one uses `$m2` is that ok? What about `$m1` and `$m3`?

In Wasmtime for example during compliation I think we understand, via interning, that `$m1` and `$m2` are the same. We consider core modules black boxes though so during compilation `$m1` and `$m3` are "different". It's only at runtime we'll realize that all the memories are actually the same. The question here is how are we allowed to implement the check in `task.return`. Will we be required to defer it all the way to runtime to check pointer/length? Or are we instead required to aovid interning `$m1` and `$m2`? Or are we required to intern `$m1` and `$m2` and are allowed to have equality at that level (e.g. fold the `task.return` check into a single compare at runtime).

Personally I'd lean towards "all of `$m1`, `$m2`, and `$m3` are equal", requiring a runtime check during `task.return` that all the pointers are the same post-instantiation. That seems like it'll be easy to implement in any possible implementation scheme. It's "slightly more expensive" in terms of maybe two compares instead of one (one for the type one for memory) but I don't think the scale of that cost should sway our decision too much here.

lukewagner:
[...]

PR #465, opened 2025-03-10
Clarify memory equality check in task.return

lukewagner:
This PR clarifies the wording to answer #464.  It also removes `realloc` from the set of options considered which @dicej pointed out are not relevant when lifting.

### DIFF
--- a/design/mvp/canonical-abi/definitions.py
+++ b/design/mvp/canonical-abi/definitions.py
@@ -192,18 +192,17 @@ class LiftLowerContext:
 ### Canonical ABI Options
 
 @dataclass
-class LiftLowerOptions:
+class LiftOptions:
   string_encoding: str = 'utf8'
   memory: Optional[bytearray] = None
+
+  def equal(lhs, rhs):
+    return lhs.string_encoding == rhs.string_encoding and \
+           lhs.memory is rhs.memory
+
+@dataclass
+class LiftLowerOptions(LiftOptions):
   realloc: Optional[Callable] = None
-
-  def __eq__(self, other):
-    return self.string_encoding == other.string_encoding and \
-           self.memory is other.memory and \
-           self.realloc is other.realloc
-
-  def copy(opts):
-    return LiftLowerOptions(opts.string_encoding, opts.memory, opts.realloc)
 
 @dataclass
 class CanonicalOptions(LiftLowerOptions):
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):

design/mvp/canonical-abi/definitions.py
```diff
@@ -192,18 +192,17 @@ def __init__(self, opts, inst, borrow_scope = None):
...
   string_encoding: str = 'utf8'
   memory: Optional[bytearray] = None
-  realloc: Optional[Callable] = None
 
-  def __eq__(self, other):
-    return self.string_encoding == other.string_encoding and \
-           self.memory is other.memory and \
-           self.realloc is other.realloc
+  def equal(lhs, rhs):
+    return lhs.string_encoding == rhs.string_encoding and \
```
dicej, line 200: As I mentioned [here](https://github.com/bytecodealliance/wasip3-prototyping/pull/70#issuecomment-2707890054), the string encoding comparison can lead to false positives (i.e. trapping when we shouldn't).

Consider a lifted function with one or more parameter types involving the `string` type, but with a result type that does _not_ involve `string`, and imagine it is lifted with the UTF-16 encoding.  If the corresponding `task.return` intrinsic does not specify a string encoding (and indeed it shouldn't since the return type doesn't need one), it will default to UTF-8, leading to a mismatch at runtime when we do this check.

Nothing new with this PR, of course, but something we'll need to address.

dicej (approved):

design/mvp/canonical-abi/definitions.py
lukewagner: You're right, but I wonder if it's worth the complexity to extend the trapping condition to be predicated on the types involved (rather than just a static subset of the set of options)?

dicej: To be clear: I'm not proposing any particular solution here; just pointing out the problem.  If we want to say e.g. "the `task.return` intrinsic must be declared with same memory and string encoding used for the corresponding lift regardless of whether the memory or string encoding is needed for the return type", that's fine with me.

lukewagner: Ok, that seems a bit simpler for now, so I think I'll keep it as is unless another reasons comes up to tighten it up.

PR #466, opened 2025-03-10
Clarify validation rules of canonical options

alexcrichton:
* Refactor the spec a bit to have a dedicated section for general-purpose validation of canonical options with additional rules per-use-site.
* Each option is present at most once in a list of options.
* Validation is required if an option is present, for example `memory` must always be a subtype of `(memory 1)` even if the lifting/lowering doesn't require a `memory`.
* At most one string encoding can be specified.
* The `callback` option is specified what type the function must have.
* The `callback` option requires the `async` option.
* The `async` option is explicitly disallowed on `error-context.*` builtins.
* The `error-context.*` builtins require `memory` and `realloc` as appropriate.
* The requirement of `memory` and `realloc` on `{stream,future}.{read,write}` is documented (although "required by" is currently a bit vague, that's left to a future refactoring).

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -33,6 +33,7 @@ being specified here.
 * [Canonical definitions](#canonical-definitions)
   * [`canon lift`](#canon-lift)
   * [`canon lower`](#canon-lower)
+  * [`canon $opts`](#canon-opts)
   * [`canon resource.new`](#canon-resourcenew)
   * [`canon resource.drop`](#canon-resourcedrop)
   * [`canon resource.rep`](#canon-resourcerep)
@@ -2727,12 +2728,14 @@ For a canonical definition:
 ```wat
 (canon lift $callee:<funcidx> $opts:<canonopt>* (func $f (type $ft)))
 ```
-validation specifies:
+
+In addition to [general validation of `$opts`](#canon-opts) the additional
+validation is performed:
+
 * `$callee` must have type `flatten_functype($opts, $ft, 'lift')`
 * `$f` is given type `$ft`
-* a `memory` is present if required by lifting and is a subtype of `(memory 1)`
-* a `realloc` is present if required by lifting and has type `(func (param i32 i32 i32 i32) (result i32))`
-* if `async` is set, a `post-return` function may not be set
+* a `memory` is present if required by lifting
+* a `realloc` is present if required by lifting
 * if a `post-return` is present, it has type `(func (param flatten_functype({}, $ft, 'lift').results))`
 
 When instantiating component instance `$inst`:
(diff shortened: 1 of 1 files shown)


alexcrichton:
The main motivation behind this was I wasn't able to find the one-liner of "what is the type of `callback`" (although I probably missed it) and that kind of snowballed into more formally specifying what wasmparser does today in terms of validation.

design/mvp/CanonicalABI.md
```diff
@@ -2890,11 +2893,13 @@ For a canonical definition:
...
+In addition to [general validation of `$opts`](#canon-opts) the additional
+validation is performed where `$callee` has type `$ft`:
+
 * `$f` is given type `flatten_functype($opts, $ft, 'lower')`
-* a `memory` is present if required by lifting and is a subtype of `(memory 1)`
-* a `realloc` is present if required by lifting and has type `(func (param i32 i32 i32 i32) (result i32))`
-* there is no `post-return` in `$opts`
+* a `memory` is present if required by lowering
+* a `realloc` is present if required by lowering
 * if `contains_async_value($ft)`, then `$opts.async` must be set
```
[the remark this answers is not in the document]
lukewagner: Good catch; I think we can delete this rule now.

design/mvp/CanonicalABI.md
```diff
@@ -3034,6 +3039,26 @@ elimination of string operations on the labels of records and variants) as well
 as post-MVP [adapter functions].
 
 
+### `canon $opts`
```
lukewagner, line 3042: Nit: could you move this section to be before `canon lift` and call it "`canonopt` validation"?

design/mvp/CanonicalABI.md
```diff
@@ -3034,6 +3039,26 @@ elimination of string operations on the labels of records and variants) as well
...
+
+Canonical options, specified here as `$opts` in a number of locations
+throughout this document, can be specified at most once per `$opts`. For example
+specifying `string-encoding=utf8` twice is an error. Each individual option, if
+present, is validated as such:
+
+* `string-encoding=utf8` - cannot be combined with `utf16` or `latin1+utf16`
+* `string-encoding=utf16` - cannot be combined with `utf8` or `latin1+utf16`
+* `string-encoding=latin1+utf16` - cannot be combined with `utf8` or `utf16`
```
lukewagner, line 3051: Could we collapse this sentence and 3 bullets down to:
* At most one `string-encoding=X` option can be set

design/mvp/CanonicalABI.md
```diff
@@ -3461,8 +3489,12 @@ For canonical definitions:
 (canon stream.read $t $opts (core func $f))
 (canon stream.write $t $opts (core func $f))
 ```
-validation specifies:
+In addition to [general validation of `$opts`](#canon-opts) validation
+specifies:
 * `$f` is given type `(func (param i32 i32 i32) (result i32))`
+* `memory` is required for `stream.write` if required by lowering
+* `memory` is required for `stream.read` if required by lifting
+* `realloc` is required for `stream.read` if required by lifting
```
lukewagner, line 3497: Would a `string-encoding` bullet also be required along the same lines?  Instead of listing `memory`, `realloc` and `string-encoding` out for each built-in, could we factor out these bullets into the new `canonopt` section?

lukewagner (commented):
Looks great, thanks for doing this.  A few quick questions:

design/mvp/CanonicalABI.md
alexcrichton: I like the idea of collapsing things yeah, I'll work on adding that.

Otherwise though I don't think we'd have a bullet for `string-encoding` because it's implicitly always set even when not mentioned, so at least currently in wasmparser there's no validation of `string-encoding` other than at most one is present.

alexcrichton:
Sorry that was a push gone wrong. I've reverted the changes on `main` but I can't seem to reopen this PR any more. I've opened a new one at https://github.com/WebAssembly/component-model/pull/467 addressing your comments Luke

PR #468, opened 2025-03-11
Gate some spec features separately from async

alexcrichton:
This commit comes out of some discussion with folks today where a few features of the async specification are going to be gated behind separate feature gates distinct from the current one. This is to help reduce scope slightly for the 0.3.0 release of WASI and focus on the other features to ship. Here two new emoji-features are added:

* 🚝: marking some builtins as `async`
* 🚟: using `async` with `canon lift` without `callback` (stackful abi)

I've tried to update `Async.md` and `CanonicalABI.md` where appropriate with these new emoji.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -3275,7 +3275,7 @@ For a canonical definition:
 ```
 validation specifies:
 * `$f` is given type `(func)`
-* 🚝 - `async` is allowed (otherwise it must be `false`)
+* 🚟 - `async` is allowed (otherwise it must be `false`)
 
 Calling `$f` calls `Task.yield_` to allow other tasks to execute:
 ```python
@@ -3322,6 +3322,7 @@ For a canonical definition:
 ```
 validation specifies:
 * `$f` is given type `(func (param $si) (param $ptr i32) (result i32))`
+* 🚟 - `async` is allowed (otherwise it must be `false`)
 
 Calling `$f` invokes the following function which waits for progress to be made
 on a waitable in the given waitable set (indicated by index `$si`) and then
@@ -3366,7 +3367,7 @@ For a canonical definition:
 ```
 validation specifies:
 * `$f` is given type `(func (param $si i32) (param $ptr i32) (result i32))`
-* 🚝 - `async` is allowed (otherwise it must be `false`)
+* 🚟 - `async` is allowed (otherwise it must be `false`)
 
 Calling `$f` invokes the following function, which returns `NONE` (`0`) instead
 of blocking if there is no event available, and otherwise returns the event the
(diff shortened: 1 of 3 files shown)


design/mvp/CanonicalABI.md
```diff
@@ -3272,6 +3275,7 @@ For a canonical definition:
 ```
 validation specifies:
 * `$f` is given type `(func)`
+* 🚝 - `async` is allowed (otherwise it must be `false`)
```
lukewagner, line 3278:
```suggestion
* 🚟 - `async` is allowed (otherwise it must be `false`)
```
The reason is that, because this function contains `trap_if(task.opts.callback and not sync)`, `canon yield` can only be called by sync-lifted functions and so there's little point in asynchronously yielding (which allows other tasks to run).

design/mvp/CanonicalABI.md
```diff
@@ -3362,6 +3366,7 @@ For a canonical definition:
 ```
 validation specifies:
 * `$f` is given type `(func (param $si i32) (param $ptr i32) (result i32))`
+* 🚝 - `async` is allowed (otherwise it must be `false`)
```
lukewagner, line 3369:
```suggestion
* 🚟 - `async` is allowed (otherwise it must be `false`)
```
Same argument as above for `yield`.  Also for `waitable-set.wait` (which GH won't let me suggest).

lukewagner (approved):
Makes sense, thanks for filing.  3 suggested tweaks:

design/mvp/CanonicalABI.md
alexcrichton: Perhaps unrelated, but is there utility in an async yield with async callbacks? I thought that would be modeled by returning a status code from the callback?

alexcrichton: Or perhaps answering myself:

| | async yield | sync yield |
|--|---------------|----------|
| callback abi | not allowed | yield to tasks in other components |
| stackful abi | yield to anything | yield to tasks in other components |

is that right? I suspect "callback abi + sync yield" may not be bound in Rust at least for now (it's easy enough to enable yielding to anything, including tasks in the same component) but that's ok

design/mvp/CanonicalABI.md
alexcrichton: Sounds good, updated! Also unrelated to this, is my understanding in https://github.com/WebAssembly/component-model/pull/468#discussion_r1991663664 basically just as applicable here as well? Basically the implicit yield here is inheriting the semantics of `yield`, and for `waitable-set.wait` you have the choice of a blocking wait with the callback abi to let other components run but letting yourself run is done by returning from the callback

design/mvp/CanonicalABI.md
lukewagner: Exactly, yes.

design/mvp/CanonicalABI.md
lukewagner: Yep!

lukewagner:
Looks great; I'll leave it open a few more days for any other comments then merge.

PR #481, opened 2025-03-26
fix(docs): use correct signed integer types in WIT mvp

mkatychev:
https://component-model.bytecodealliance.org/design/wit.html#primitive-types

The WIT primitive types documentation references an `s32` rather than `i32`, while `i32` is present in WAT/WASM types specs I was unable to find one in WIT specs.


lukewagner:
Thanks!

Issue #482, opened 2025-03-26
Interface MVP bnf implies `async` can be used twice in an item


mkatychev:
[MVP docs](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#item-interface) currently state two places where the `async` keyword can be used:
```bnf
func-item ::= id ':' func-type ';'

func-type ::= 'async'? 'func' param-list result-list
resource-item ::= 'resource' id ';'
                | 'resource' id '{' resource-method* '}'
resource-method ::= func-item
                  | id ':' 'static' 'async'? func-type ';'
                  | 'constructor' param-list ';'
```

Does that mean you could have a `resource-method` that's something like the one below?
```wit
foo: static async async func() -> future;
```

alexcrichton:
This is a bug, it shouldn't be valid to use `async async`

PR #484, opened 2025-03-27
fix(docs): remove extra async from resource-method

mkatychev:
Resolves https://github.com/WebAssembly/component-model/issues/482 by removing extra `'async'?` from `resource-emthod` BNF definition


mkatychev:
@alexcrichton 
I'm guessing this is the right way to prune the `async async func() -> ...` scenario.

alexcrichton:
Thanks!

PR #487, opened 2025-03-31
Update table max length

dicej:
The code says `2**28` but the prose says `2**30`.  I believe the former is the correct value?


lukewagner:
D'oh.  Thanks!

PR #489, opened 2025-04-01
Separately gate `error-context` from async

### LINKED ISSUE #469: Should `{future,stream}.close-{readable,writable}` take ownership of the error-context?
[Currently in the canonical ABI](https://github.com/WebAssembly/component-model/blob/f048984d391c43c5fb3c9f8e45a4979a69601980/design/mvp/CanonicalABI.md#-canon-streamfutureclose-readablewritable) these intrinsics only acces the error-context rather than consuming it, like a `borrow<_>` instead of `own<_>`. My (possibly naive) expectation though would be that in the majority of cases that an error context is passed in and then discarded, so should this perhaps both take an `error-context` and remove it from the component's table?

### LINKED ISSUE #472: Require-error-on-close may be a composability hazard
Currently as-specified it's required that for an open stream/future that an error is provided when dropping it, I believe for both the reader and the writer end. (please correct me if I'm wrong about this though). In other words this test currently passes:

```python
async def test_dropping_future_ends():
  inst = ComponentInstance()
  lower_opts = mk_opts()

  host_ft1 = FuncType([FutureType(U8Type()), BoolType()],[])
  async def host_func(task, on_start, on_return, on_block):
    [future, close_now] = on_start()
    if close_now:
        future.close(None)
    on_return([])

  lift_opts = mk_opts()
  async def core_func(task, args):
    assert(not args)

    [wfi] = await canon_future_new(U8Type(), task)
    [] = await canon_lower(lower_opts, host_ft1, host_func, task, [wfi, 1])
    [] = await canon_future_close_writable(U8Type(), task, wfi, 0)

    [wfi] = await canon_future_new(U8Type(), task)
    [] = await canon_lower(lower_opts, host_ft1, host_func, task, [wfi, 0])
    try:
        [] = await canon_future_close_writable(U8Type(), task, wfi, 0)
    except Trap:
      return []

    assert(False)

  await canon_lift(lift_opts, inst, FuncType([],[]), core_func, None, lambda:[], lambda _:(), Task.sync_on_block)
```

Here `core_func` created two futures and closed two futures. In both cases it provides no error context but it wasn't required to in the first place because `host_func` closed the future, but it's required to in the second case bcause `host_func` didn't close the future.

This to me seems like it can create a hazard for composing components together. Whether or not you're allowed to close a future/stream without an error-context is dependent on what the other end has done in some other component. If another end closed prematurely you're allowed to close without an error, but if one day that other end is refactored to keep the end open a little longer than all of a sudden the original caller will break since it should have provided an error-context but forgot to.

I was talking a bit about this with @vados-cosmonic and @dicej today and it was brought up that whether or not an error-context is required at all has design implications on WASI APIs (e.g. `future<option<T>>` if you're required to close-with-error vs `future<T>` if close-with-no-error is an option).

---

Overall, assuming my understanding here is right in that close-with-error is dependent on what the other side has done, I'm curious about futures/streams/error-context in a few different ways:

* One possible fix to this is to maintain state for each half of the future/stream so a components requirement to close-with-error is purely local to the component, not dependent on what the other end is doing. I'm personally not a fan of adding more infrastructure in this regard.
* Ideally I would prefer to drop the close-requires-error requirement altogether. I'm not familiar with the original motivation, however, so I'm basically curious to learn up more on that.
* Is it in theory possible to remove `error-context` entirely? Or defer it to a future WIT extension to a new type? If the above "ideal bullet" for me is done then if it's optional to close-on-error, then it seems like all `future<T>` is roughly equivalent to `future<result<T, error-context>>` except in a slightly more complicated manner. In such a situation I'd prefer to drop the built-in integration of futures/streams with `error-context` entirely and leave that up to WIT-level APIs to define `future<result<T, error-context>>`. (or something like that)

### LINKED ISSUE #474: Defer `error-context` to after 0.3.0
I talked today with a few folks and I wanted to write up our thoughts/conclusions into an issue to get some wider feedback before sending a PR to commit to anything. The basic premise is:

* Emoji-gate `error-context` in the explainer document.
* Remove `error-context` arguments/integration with futures/streams.

In effect this would defer integration with `error-context` until after WASI 0.3.0 has shipped and make `error-context` a future addition to WIT and the component model.

---

Ok, but why? The original motivation for considering this comes from me considering current issues such as https://github.com/WebAssembly/component-model/issues/469 and https://github.com/WebAssembly/component-model/issues/472. Overall, in my opinion at least, *some* changes are needed in the spec with respect to error-context. While these can be fixed and resolved in isolation I wanted to bring up a possible broader context of: can we defer it entirely until after 0.3.0?

In talking with @lukewagner one of the chief desires of `error-context` is eventual future integeration with the WIT `result` type. For example when constructing one you could supply an error-context or when receiving one you could optionally acquire an error-context. Currently `result` does not have this integration so it'll have to be designed in a backwards-compatible fashion, and the theory is that such backwards-compatible infrastructure could be equally applied to streams/errors as well.

---

There's a lot of other various inter-related things that were discussed around this, but that's the crux of the issue effectively. I wanted to open this up before making a PR with the actual changes to get some broader discussion. I know that @vados-cosmonic is probably likely interested in this, and various WASI folks as well.

alexcrichton:
This commit adds a new emoji-gate of 📝 to gate the `error-context` type separately from async. This additionally removes integration of `error-context` with built-in future/stream types where they can still be closed but are no longer closed with an optional error.

Closes #474
Closes #472
Closes #469

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -190,7 +190,7 @@ primvaltype   ::= 0x7f                                    => bool
                 | 0x75                                    => f64
                 | 0x74                                    => char
                 | 0x73                                    => string
-                | 0x64                                    => error-context 🔀
+                | 0x64                                    => error-context 📝
 defvaltype    ::= pvt:<primvaltype>                       => pvt
                 | 0x72 lt*:vec(<labelvaltype>)            => (record (field lt)*)    (if |lt*| > 0)
                 | 0x71 case*:vec(<case>)                  => (variant case+) (if |case*| > 0)
@@ -307,9 +307,9 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift 
            | 0x19 t:<typeidx> async?:<async?>                    => (canon future.cancel-write async? (core func)) 🔀
            | 0x1a t:<typeidx>                                    => (canon future.close-readable t (core func)) 🔀
            | 0x1b t:<typeidx>                                    => (canon future.close-writable t (core func)) 🔀
-           | 0x1c opts:<opts>                                    => (canon error-context.new opts (core func)) 🔀
-           | 0x1d opts:<opts>                                    => (canon error-context.debug-message opts (core func)) 🔀
-           | 0x1e                                                => (canon error-context.drop (core func)) 🔀
+           | 0x1c opts:<opts>                                    => (canon error-context.new opts (core func)) 📝
+           | 0x1d opts:<opts>                                    => (canon error-context.debug-message opts (core func)) 📝
+           | 0x1e                                                => (canon error-context.drop (core func)) 📝
            | 0x1f                                                => (canon waitable-set.new (core func)) 🔀
            | 0x20 async?:<async>? m:<core:memidx>                => (canon waitable-set.wait async? (memory m) (core func)) 🔀
            | 0x21 async?:<async>? m:<core:memidx>                => (canon waitable-set.poll async? (memory m) (core func)) 🔀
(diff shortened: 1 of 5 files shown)


lukewagner:
Thanks!

PR #491, opened 2025-04-02
Document some conventions for `context.{get,set}`

### LINKED ISSUE #485: Toolchain conventions and `context.get`
Currently in `wit-bindgen` it uses `context.get` to store a per-exported-task data structure which manages waitable sets and and child tasks from Rust's perspectives (not component-model tasks, Rust tasks). This is not going to work, however, once there are multiple versions of `wit-bindgen` in the ecosystem with async bindings. For example within a component there could be multiple versions of `wit-bindgen`:

* `0.A.0`: used for one export
* `0.B.0`: used for one export
* `0.C.0`: used for some imports

In these situations the per-task data structure may not be the same across versions `A`, `B`, and `C`, and naively using `context.get` is going to lead everyone to sharing the same pointer but silently corrupting data at runtime as the data structures change.

I don't have a great idea of a solution yet, but it feels like this is going to lie somewhere in the realm of tooling conventions rather than the component model itself. What exactly such a convention is, though, I'm not sure.

alexcrichton:
Closes #485

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -251,6 +251,35 @@ reason why "context-local" storage is not called "task-local" storage (where a
 For details, see [`context.get`] in the AST explainer and [`canon_context_get`]
 in the Canonical ABI explainer.
 
+#### Context-Local Storage Conventions
+
+While this is not required it's recommended that toolchains follow a set of
+conventions around using context-local storage to ensure that different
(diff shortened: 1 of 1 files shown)


alexcrichton:
Closing in favor of https://github.com/WebAssembly/component-model/pull/501#pullrequestreview-2769662069

Issue #480, opened 2025-03-25
Gate MVP example conflicts with BNF


mkatychev:
The feature gate example lists a gate as having both a version and feature field: ` @since(version = 0.2.2, feature = fancy-foo)`
https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#feature-gates

However the BNF for gate syntax lists `feature-field` as only being used by an `@unstable` gate:
```bnf
gate ::= gate-item*
gate-item ::= unstable-gate
            | since-gate
            | deprecated-gate

unstable-gate ::= '@unstable' '(' feature-field ')'
since-gate ::= '@since' '(' version-field ')'
deprecated-gate ::= '@deprecated' '(' version-field ')'

feature-field ::= 'feature' '=' id
version-field ::= 'version' '=' <valid semver>
```

lukewagner:
(Sorry for the slow reply!  I was out the last 2 weeks.). Thanks for the careful read and good catch!  If I'm recalling history correctly, we started by allowing `@since` to have both a version and feature-name, and then removed the latter in #387, but we forgot to update this example.  I'll fix the example shortly.

PR #493, opened 2025-04-07
Fix WIT.md example

lukewagner:
As found by and resolving #480, this PR fixes a WIT.md example to not use a feature-name inside a `@since` gate.


alexcrichton (approved):

mkatychev (approved):
LGTM, thanks for the clarification

Issue #478, opened 2025-03-24
Resources with a function name that is the same as the resource name does not work in all languages


jsturtevant:
We got a report in [wit-bindgen for c#](https://github.com/bytecodealliance/wit-bindgen/issues/1232) that this currently valid wit would cause an error:

```wit
interface resources {
  resource foo {
    foo: func();
  }
}
```

This is currently translated to 

```csharp
public interface IResources {

    public class Foo: IDisposable {
       ....other boiler plate ommited...

        public   unsafe void Foo()
        {
            var handle = this.Handle;
            FooWasmInterop.wasmImportFoo(handle);

        }

    }

}
```

In C# it fails with https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0542?f1url=%3FappId%3Droslyn%26k%3Dk(CS0542) since you cannot have functions of the same name as the class unless it is a constructor.

We've discussed some type of remapping in wit-bindgen but this puts a burden on the c# developer when they run [into the issue ](https://github.com/bytecodealliance/wit-bindgen/pull/1238#issuecomment-2745307437). 

One suggestion was that this type of rule might be applicable more broadly at the wit level. Raising it here for more awareness as I think other languages might also run into it.

lukewagner:
That makes sense.  Coincidentally, we just added this [strongly-unique](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#name-uniqueness) validation rule that was intended to rule our bindings-generator-hazards just like this, but I see we don't rule out this specific case.  I think it'd make sense to add a new case to the rule that considered the names `r` and `[method]r.r` to *not* be strongly-unique (and thus not allowed in the same scope).

MendyBerger:
@jsturtevant this came up in [browser.wit](https://github.com/MendyBerger/browser.wit), a project to translate all web APIs into wit and make them usable in components.

The specific case is here:
https://github.com/MendyBerger/browser.wit/blob/9a4249b347d0a1bb70d8ea3b969598b5a857b28d/wit/web.wit#L10397

Don't think you have to be concerned about backwards compatibility, it's not really used anywhere yet.

lukewagner:
(I'm back, sorry for the delay) Ok, PR up in #494, PTAL!

PR #494, opened 2025-04-07
Restrict 'strongly-unique' predicate to prevent further bindgen conflicts

lukewagner:
This adds the change discussed in #478 to avoid conflicts in the bindings generators for OOP-like languages between methods and constructors.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -2464,7 +2464,9 @@ names that appear within the same scope.
 To determine whether two names (defined as sequences of [Unicode Scalar
 Values]) are **strongly-unique**:
 * If one name is `l` and the other name is `[constructor]l` (for the same
-  `label` `l`), they are strongly-unique.
+  `label` `l`), they *are* strongly-unique.
+* If one name is `l` and the other name is `[method]l.l` (for the same
+  `label` `l`), they *are not* strongly-unique.
 * Otherwise:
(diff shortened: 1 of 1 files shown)


jsturtevant:
LGTM

alexcrichton (approved):

design/mvp/Explainer.md
```diff
@@ -2464,7 +2464,9 @@ names that appear within the same scope.
 To determine whether two names (defined as sequences of [Unicode Scalar
 Values]) are **strongly-unique**:
 * If one name is `l` and the other name is `[constructor]l` (for the same
-  `label` `l`), they are strongly-unique.
+  `label` `l`), they *are* strongly-unique.
+* If one name is `l` and the other name is `[method]l.l` (for the same
+  `label` `l`), they *are not* strongly-unique.
```
alexcrichton, line 2469: Should this rule be extended to other method-like things like `[async method]`, `[static]`, and `[async static]`? I think `[async method]` is a given, but I'm basically curious if `[static]l.l` conflicts with `l`

lukewagner: Oops, right, I totally forgot about these.  Given the root motivation of avoiding conflicts with languages' constructor rules, I'd think that `[static]l.l` would conflict with `l`.  I'll file a new PR...

Issue #490, opened 2025-04-01
Ambiguity between "operation cancelled" and "0-length operation completed"


alexcrichton:
Assuming #444 with respect to 0-length reads writes there's an ambiguity when cancelling an operation between whether the operation was cancelled or wheter the 0-length operation completed. This was discussed briefly at https://github.com/WebAssembly/component-model/pull/444#issuecomment-2766811302 and it sounds like there are encoding changes planned to address this. Regardless though I wanted to open an issue here to write a FIXME in some code, so this is basically a tracking issue for said encoding changes.

PR #498, opened 2025-04-10
Change stream results to indicate cancellation

lukewagner:
As first discussed in #444 and resolving #490, this PR explicitly propagates whether an async stream/future read/write completed or was cancelled.  The `Buffer.MAX_LENGTH` is lowered to 2<sup>28</sup>-1 to match `Table.MAX_LENGTH` and leave a nice 4-bit space for any future additions.

(Originally, I was going to put the 2-bit "result" field in the MSBs such that `Buffer.MAX_LENGTH` could be increased if we wanted in the future, but I noticed this is the reverse of the `canon lower` result, which seemed gratuitously inconsistent.  Also, I realized that once we change the CABI to take advantage of multi-return, this issue will go away, since we'll just return two `i32`s.  I'm not wed to this detail, though.)

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -354,7 +354,7 @@ values *into* the buffer's memory. Buffers are represented by the following 3
 abstract Python classes:
 ```python
 class Buffer:
-  MAX_LENGTH = 2**30 - 1
+  MAX_LENGTH = 2**28 - 1
   t: ValType
   remain: Callable[[], int]
 
@@ -1055,7 +1055,7 @@ stream.)
 ```python
 RevokeBuffer = Callable[[], None]
 OnPartialCopy = Callable[[RevokeBuffer], None]
-OnCopyDone = Callable[[], None]
+OnCopyDone = Callable[[Literal['completed','cancelled']], None]
 
 class ReadableStream:
   t: ValType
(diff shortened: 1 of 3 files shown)


design/mvp/CanonicalABI.md
```diff
@@ -3573,36 +3577,46 @@ allowing multiple partial copies to complete in the interim, reducing overall
...
+  assert(buffer.progress <= Buffer.MAX_LENGTH < 2**28)
+  packed = result | (buffer.progress << 4)
+  assert(packed != BLOCKED)
+  return packed
+```
+The `result` indicates whether the stream was closed by the other end, the
+copy was cancelled by this end (via `{stream,future}.cancel-{read,write}`) or,
+otherwise, completed successfully. In all cases, any number of elements (from
+`0` to `n`) may have *first* been copied into or out of the buffer passed to
+the `read` or `write` and so this number is packed into the `i32` result.
```
alexcrichton, line 3619: Would it be possible to provide a guarantee that if `CANCELLED` is returned then the number of elements transferred is always 0?

alexcrichton:
> Originally, I was going to put the 2-bit "result" field in the MSBs such

I was sort of having second thoughts about this too so I think lining things up seems reasonable. As you say getting multi-value means that we can lift all these restrictions anyway

design/mvp/CanonicalABI.md
lukewagner: I was thinking that it was indeed possible to receive `CANCELLED` and >0 elements.  Hypothetically this could occur even with component-to-component streaming if a few partial writes succeed but then, when the reader tries to cancel, the writer is (in a hypothetical future with threads) on a different thread and the runtime is holding a lock while doing a large copy between buffers, and maybe we allow in this case the cancellation to return `BLOCKED`, and then it seems like we'd expect `CANCELLED` with >0 elements (as opposed to `COMPLETED`; the idea being that `COMPLETED` means "on its own", which is perhaps a distinction without a difference; I dunno?)

alexcrichton: Ok makes sense to me, and I don't think this should be too too hard to bind in various languages. There's going to be some subtelty to interpreting the return values but I don't think that'll be too bad.

alexcrichton (approved):

PR #504, opened 2025-04-22
Document fixed-size-lists must be non-empty

### LINKED ISSUE #503: Reject zero-length fixed-length lists
We've intentionally avoided zero-sized types and would like to continue to avoid them.

alexcrichton:
Currently the component model and WIT generally avoid zero-sized types since they're not always handled well in all languages. Given the niche use case of a zero-sized list add a validation rule that fixed-size-lists are non-zero in length.

Closes #503


alexcrichton:
cc @ChristofPetig

lukewagner (approved):
Thanks!

PR #505, opened 2025-04-25
Improve binary EBNF consistency

lann:
Add inline non-zero length constraint to fixed-length list matching surrounding rules.

Followup to https://github.com/WebAssembly/component-model/pull/504


lukewagner (approved):
Good idea!

PR #507, opened 2025-04-28
Document some filesystem conventions of WIT

alexcrichton:
I've been meaning to write these down for awhile, and now I've gotten around to them! The goal is to outline some preexisting conventions for specifying WIT to bindings generators in a location that isn't buried in any implementation but instead in a shared location that can be iterated on as well.


morenol:
Thanks for this

lukewagner (approved):
Thanks for writing this up!

Issue #459, opened 2025-02-28
async: How does cross-task communication work?


alexcrichton:
I've talked with @lukewagner and @dicej in various degrees about this problem before, but I wanted to try to write up an issue with my thoughts about something concrete at least. At a high level I'm curious how an "async mutex" would work within a component, but I'm going to attempt to make this more specific as well.

Let's say I have a hypothetical setup like this:

* Component A has an export, `run`.
* An external entity (host or other component), calls `run` to create task T1.
* Task T1 acquires an async mutex, but then decides it has taken too long and performs an async yield.
* An external entity then calls `run` again to create task T2.
* Task T2 attempts to acquire the mutex, sees that it's locked, and needs to block.
  * I'm assuming the `callback` ABI here, so this would return. Is this allowed? Returning while blocking on nothing?
* The yield for T1 finishes and then it drops the lock.
  * How does this "wake up" T2? How is the "current task" context switched?

My vague undersatnding is that T2 traps when it returns waiting for the lock, but I'm also still booting up on all the canonical ABI pieces.

alexcrichton:
I wrote up some other somewhat related thoughs at https://github.com/bytecodealliance/wit-bindgen/issues/1182

alexcrichton:
At least for the use case I was thinking of Rust's support we'd definitely want to support the `callback` ABI. While `memory.atomics.wait-async` could work I'd also want to clarify that the blocking operation of an async mutex isn't actually blocking the current thread, it's more that it did some synchronization to figure out it needs to block but after that it's going to suspend. In Rust-parlance the future returns `Poll::Pending` and is arranged to receive a notification when the lock is unlocked. Later when another task unlocks the mutex it'll send a signal to the blocked task (via Rust APIs) and that's the part that needs to interact with the component model.

That's why `memory.atomics.wait-async` could work in theory but it's also not a strict parallel to `memory.atomics.wait`. In that sense there's other possible designs, too, which aren't dependent on memory. I haven't fully thought this through but something could be along the lines of:

* `canon task.id : [] -> [i32]`
* `canon task.wake_sibling : [i32] -> []`

Where when a task blocks on an async mutex it's "to be notified" handle knows the original task's ID (by calling `task.id`). When the lock is unlocked that notification will see that it's unlocking from a separate task via `task.wake_sibling`. Although as I write this down I realize that this still doesn't give the CM knowledge of the original task blocking, it just called `task.id` and then didn't block on anything and also didn't return...

[...]

PR #508, opened 2025-04-28
Handle same-instance stream/future writes when element is empty

lukewagner:
As discussed in #475, as a temporary expediency, the CABI currently traps when trying to lift and lower in the same component instance to avoid the otherwise subtle ordering issues that arise with memory aliasing.  However, if the element type is empty (as enabled by #440), there is no such memory aliasing issue.  Moreover, such futures/streams can be used as a simple way to support #459 use cases without having to add anything fancier like the `wait-async` built-in suggested in #459.  Thus, this tiny PR loosens the same-instance trapping condition to only trap if there is a non-empty element type.

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -363,8 +363,8 @@ signalled by performing a `0`-length read or write (see the [Stream State]
 section in the Canonical ABI explainer for details).
 
 As a temporary limitation, if a `read` and `write` for a single stream or
-future occur from within the same component, there is a trap. In the future
-this limitation will be removed.
+future occur from within the same component and the element type is non-empty,
+there is a trap. In the future this limitation will be removed.
 
 The `T` element type of streams and futures is optional, such that `future` and
 `stream` can be written in WIT without a trailing `<T>`. In this case, the
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1362,7 +1362,8 @@ but in the opposite direction. Both are implemented by a single underlying
       self.set_pending(inst, buffer, on_partial_copy, on_copy_done)
       return 'blocked'
     else:
-      trap_if(inst is self.pending_inst) # temporary
+      assert(self.t == src.t == dst.t)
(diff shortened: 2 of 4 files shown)


alexcrichton (approved):

dicej (approved):

lukewagner:
Trying to actually test this, I just noticed that #440 forgot to update Binary.md to provide an encoding for the "none" case in all the stream/future built-ins (they currently unconditionally take a `typeidx`).

In the type section, which was updated, the `future`/`stream` type constructors currently take a `<valtype>?`.  Updating the stream built-ins to do likewise would thus be a binary breaking change.  I was wondering why the stream built-ins took a `typeidx` and not a `valtype` (given that the element type is often a `primvaltype`), and I think what happened is that the `typeidx` initially referred to the whole `stream<T>`/`future<T>` (which can never be a `primvaltype`) and then we changed it without changing the encoding.  Thinking about how stream/future built-ins would work with `stream<T,U,E>`/`future<T,E>`, though, I think that several of the stream built-ins need to see the whole compound `stream`/`future` type, not just the `T` (since e.g. a future read would be able to produce a `T` or an `E`).

Based on this, what I think the fix is here is to keep the binary format taking a `typeidx`, but require it to refer to a `stream`/`future`.  `wasip3-prototyping` could keep allowing `typeidx` to refer to non-`stream`/`future` types to avoid breaking producers during the transition, although if any pre-transition producer used a `stream<stream<T>>`, it would still break.

WDYT?

alexcrichton:
I think the prototyping repo [already implements the behavior you describe](https://github.com/bytecodealliance/wasm-tools/blob/35e3e4933463b0cb85d0a45b5df78b45b07806ab/crates/wasmparser/src/validator/component.rs#L1704-L1736) where the typeidx is required to be a `(future ...)` or `(stream ...)` depending on the intrinsinc. I suspect that "violated" the spec at the time, but may prove useful in aligning to this direction nonetheless?

lukewagner:
Sweet; that looks great.  I expect it matched spec before the spec changed.  I'll file a new PR to switch back.

Issue #434, opened 2025-01-07
Potential optimization for imported async functions


cpetig:
Exported async functions have been changed in November to receive their arguments on the value stack, so these function calls now impose less overhead.

Imported async functions on the other hand still use a heap allocated list to pass arguments. While this is motivated by minimal overhead for back-pressure it penalizes the common case of async function calls with few parameters, e.g. a string or list. The overhead to store the arguments on the host in the (less likely?) back-pressure case could be acceptable in comparison to this unconditional heap allocation on the client side.

When I brought this up in https://github.com/bytecodealliance/wit-bindgen/pull/1082#issuecomment-2576377929 Joel proposed to open discussion here.

lukewagner:
(Sorry for the slow reply; getting back from holidays) Currently, if there is only one flattened parameter, it is passed as a scalar instead of going through linear memory (see [`flatten_functype`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#flattening)).  So, one option we have is to relax increase this max-async-flat-parameters limit from 1 to something a bit bigger.

We do have to be a little careful since increasing this limit will increase the static size of the engine-internal structure used to represent async calls to have space for the maximum number of async flattened parameters.  But maybe a modest increase to, say, "4" would be reasonable?  It'd be nice to have performance data to motivate this, though.

lukewagner:
Proposed in #520.

PR #520, opened 2025-05-20
Extend number of flat parameters in async lower from 1 to 4

lukewagner:
As suggested in #434, this PR increases the number of flattened parameters allow in async-lowered calls from 1 to 4.  (This is different than the `MAX_FLAT_ARGS` of 16 for sync-lowered calls because runtimes will likely need to reserve space for the flat argument tuple in an inline fixed-size "subtask" structure.)

For consistency, it would also make sense to change `future.write` to take flattened arguments, but this is a larger change (at least in the spec text) since futures are defined as specialized streams which work in terms of buffers, so I was thinking of breaking that out into a separate PR.

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -620,58 +620,68 @@ which to use by default.
 
 ### Async Import ABI
 
-Given an imported WIT function:
+Given these imported WIT functions (using the fixed-length-list feature 🔧):
 ```wit
 world w {
-  import foo: func(s: string) -> string;
+  import foo: func(s: string) -> u32;
+  import bar: func(s: string) -> string;
+  import baz: func(t: tuple<u64; 5>) -> string;
+  import quux: func(t: tuple<u32; 17>) -> string;
 }
 ```
-the default sync import function signature is:
+the default/synchronous lowered import function signatures are:
 ```wat
 ;; sync
-(func (param $s-ptr i32) (param $s-len i32) (param $out i32))
+(func $foo (param $s-ptr i32) (param $s-len i32) (result i32))
+(func $bar (param $s-ptr i32) (param $s-len i32) (param $out-ptr i32))
+(func $baz (param i64 i64 i64 i64 i64) (param $out-ptr i32))
+(func $quux (param $in-ptr i32) (param $out-ptr i32))
 ```
-where `$out` must be a 4-byte-aligned pointer into linear memory into which the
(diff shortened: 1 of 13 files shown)


design/mvp/Async.md
```diff
@@ -620,58 +620,68 @@ which to use by default.
...
 
-Given an imported WIT function:
+Given these imported WIT functions (using the fixed-length-list feature 🔧):
 ```wit
 world w {
-  import foo: func(s: string) -> string;
+  import foo: func(s: string) -> u32;
+  import bar: func(s: string) -> string;
+  import baz: func(t: tuple<u64; 5>) -> string;
+  import quux: func(t: tuple<u32; 17>) -> string;
```
alexcrichton, line 629: s/tuple/list/

design/mvp/CanonicalABI.md
```diff
@@ -2470,7 +2471,7 @@ def flatten_functype(opts, ft, context):
         else:
           flat_results = []
```
alexcrichton, line 2472: For the lift case (sorry can't comment on the lines above this), this looks like it's not limiting `flat_params` at all? I think that this wants the same logic for handling the maximum length, right?

alexcrichton (approved):

design/mvp/CanonicalABI.md
```diff
@@ -2466,6 +2466,8 @@ def flatten_functype(opts, ft, context):
   else:
     match context:
       case 'lift':
+        if len(flat_params) > MAX_FLAT_PARAMS:
```
alexcrichton, line 2469: To confirm, should this be `MAX_FLAT_ASYNC_PARAMS`?

lukewagner: Since we're talking about the incoming direction, the motivation for limiting outgoing to `4` doesn't apply, so I hadn't thought to change it, but does it simplify adapter fusion in cases where the ABI options cause the core lifted/lowered signatures to line up?

alexcrichton: It looks like wasip3-prototyping already supports the limit here as 16, and already has adapters for memory->flat, so I think it's reasonable to leave this as-is (and I agree the motivation for 4 is not relevant here).

I'll assume this'll stick to 16 and see how far I get

lukewagner:
(Updated tests/async to succeed with just-merged wasip3-protoyping support.)

PR #522, opened 2025-06-02
fix: async examples

vados-cosmonic:
Fix some WAT and typos related ot the stackless/stackful examples and introduce sections that correspond to either approach.


lukewagner:
Thanks!

PR #528, opened 2025-06-06
fix: async indicator in canon yield

vados-cosmonic:
This commit fixes the variable name in the python code that seems to be referencing whether it's an *async* yield rather than a *sync* yield, but is written as 'sync'.


lukewagner:
As a convention throughout the CABI (commented in `CanonicalOptions`), to avoid writing `async_` everywhere, we use `sync`, so I'd prefer to do that consistently.

vados-cosmonic:
Ah thanks for clarifying this -- will close his then!

PR #530, opened 2025-06-09
fix: output types of Task.yield_

vados-cosmonic:
Small change for the python type bindings -- it seems `Task.yield_` is written such that it *has* to return an actual event tuple (not even an option of one).


lukewagner:
Thanks!

PR #531, opened 2025-06-16
Disallow `post-return` from calling imports, allow blocking import calls even when `callback` is used

lukewagner:
This PR disallows calling imports during `post-return` which means that sync-to-sync calls never require a fiber (noting that, if `post-return` can call imports, it can block after returning-to-and-unblocking the synchronous caller, thereby requiring a fiber to continue asynchronous execution in the callee), preserving the intended cost model of synchronous calls.

This implies that `post-return` will stick around instead of getting deprecated and removed in the future, thereby removing the reason for having the (as yet unimplemented) `always-task-return` `canonopt`.  Note that, with only a bit more work, a synchronous export that wants to use `task.return` could always lift `async` (no `callback`) and use `backpressure.set` to achieve serialized execution.

The PR also relaxes a few of the traps that prevent certain blocking calls from being used in conjunction with `callback`.  These traps were intended to avoid the need for fibers in some cases, but based on implementation experience, we think they'll be too strict and prevent various realistic scenarios where an unaware transitive dependency does a little unexpected blocking.  Notes and TODOs are added to Async.md explaining this more.

### DIFF
--- a/design/mvp/Async.md
+++ b/design/mvp/Async.md
@@ -67,12 +67,10 @@ by allowing components to import and export "async" functions which abstract
 over, and can be implemented by, idiomatic concurrency in a variety of
 programming languages:
 * `async` functions in languages like C#, JS, Python, Rust and Swift
-  (implemented using [`callback` functions](#waiting))
 * stackful coroutines in languages like Kotlin, Perl, PHP and (recently) C++
 * green threads as-if running on a single OS thread in languages like Go and
   (initially and recently again) Java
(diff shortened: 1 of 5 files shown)


alexcrichton (approved):

dicej:
@lukewagner when you get a chance, would you mind adding a WAT test for this?  That will help ensure implementations get it right.

lukewagner:
You bet; good idea: #535

PR #535, opened 2025-06-24
Add test/values/trap-in-post-return.wast

lukewagner:
This PR adds a .wast calling all the `canon` function that contain a `may_leave` trap from under a `post-return` function testing that they trap.


dicej (approved):
That's a big file!  Thanks for adding it.

Alex and I are deep in the process of upstreaming `wasip3-prototyping` into the main `wasmtime` repo, so I might not have a chance to get this test passing this week, but it's on my to-do list.

dicej:
FYI, I've got the test passing in [this branch](https://github.com/dicej/wasmtime/tree/disallow-import-calls-in-post-return), but I'm going to hold off PRing that until we're done upstreaming async support to Wasmtime.  We'll probably add a config option to get the old behavior (i.e. allowing import calls in post-return functions) in case this breaks real-world pre-async code.

lukewagner:
Awesome, thanks!

Issue #534, opened 2025-06-24
Interface version / compatibilty changes


lann:
Wasmtime currently performs a [slightly complicated dance](https://docs.rs/wasmtime/34.0.1/wasmtime/component/struct.Linker.html#names-and-semver) to allow a guest import of an interface to be linked with a different but semver-compatible version of that interface. Based on a discussion with @lukewagner and @alexcrichton we think this behavior can be brought into the spec itself with some relatively straightforward changes:

- Define "interface version canonicalization", meaning roughly "trimming a version down to the part that matters for compatibility"
  - e.g. `0.2.5` &rarr; `0.2`, `1.2.3` &rarr; `1`
  - See https://github.com/WebAssembly/component-model/pull/378; specifically https://github.com/WebAssembly/component-model/blob/3c6c4a9f1228676a342d2f971ced73b0e0ba0d2b/design/mvp/BuildTargets.md#interface-name-canonicalization
    > This definition preserves semver prerelease tags but we think that may not be necessary

- Update `<version>` to allow canonical versions in addition to full semver

- To allow round-tripping WIT and nicer error messages, add an optional field to imports/exports that preserves the original full version
  - This could contain the entire version (`1.2.3`) or just the trimmed suffix (`.2.3` / `2.3`)

- (pre-1.0) start _enforcing_ (instead of just allowing) canonical versions

With these spec changes, bindings generators would start emitting canonicalized versions in import/export names. This should effectively give behavior similar to what the Wasmtime linker does today, and additionally should solve the same underlying version compatibility problems for other tools like wac and wasi-virt.

lukewagner:
This makes sense to me since it allows simple equality-based name matching to provide the semver semantics that it seems like everyone needs in practice, minimizing the number of places that have to think about it while still preserving the precise version information for diagnostic purposes.

I'm happy to write up a PR for this repo (after waiting to see if there's any more thoughts in this issue) or happy to let you do it @lann if you're inclined.

lann:
I can take a crack at it.

lann:
This draft is ready for initial review: https://github.com/WebAssembly/component-model/pull/536

_Moved comment about binary format [to the PR](https://github.com/WebAssembly/component-model/pull/536#issuecomment-3006199206)._

PR #536, opened 2025-06-25
Interface version canonicalization

lann:
See #534

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -2269,54 +2269,53 @@ interpreted as a *lexical* grammar defining a single token and thus whitespace
 is not automatically inserted, all terminals are single-quoted, and everything
 unquoted is a meta-character.
 ```ebnf
-exportname    ::= <plainname>
-                | <interfacename>
-importname    ::= <exportname>
-                | <depname>
-                | <urlname>
-                | <hashname>
-plainname     ::= <label>
-                | '[async]' <label> 🔀
-                | '[constructor]' <label>
-                | '[method]' <label> '.' <label>
-                | '[async method]' <label> '.' <label> 🔀
-                | '[static]' <label> '.' <label>
-                | '[async static]' <label> '.' <label> 🔀
-label         ::= <fragment>
-                | <label> '-' <fragment>
-fragment      ::= <word>
-                | <acronym>
-word          ::= [a-z] [0-9a-z]*
-acronym       ::= [A-Z] [0-9A-Z]*
-interfacename ::= <namespace> <label> <projection> <interfaceversion>?
-                | <namespace>+ <label> <projection>+ <interfaceversion>? 🪺
-namespace     ::= <words> ':'
-words         ::= <word>
-                | <words> '-' <word>
-projection    ::= '/' <label>
-# FIXME: surrounding alignment
+exportname        ::= <plainname>
+                    | <interfacename>
+importname        ::= <exportname>
+                    | <depname>
+                    | <urlname>
+                    | <hashname>
+plainname         ::= <label>
+                    | '[async]' <label> 🔀
+                    | '[constructor]' <label>
+                    | '[method]' <label> '.' <label>
+                    | '[async method]' <label> '.' <label> 🔀
+                    | '[static]' <label> '.' <label>
+                    | '[async static]' <label> '.' <label> 🔀
+label             ::= <fragment>
+                    | <label> '-' <fragment>
+fragment          ::= <word>
+                    | <acronym>
+word              ::= [a-z] [0-9a-z]*
+acronym           ::= [A-Z] [0-9A-Z]*
+interfacename     ::= <namespace> <label> <projection> <interfaceversion>?
+                    | <namespace>+ <label> <projection>+ <interfaceversion>? 🪺
+namespace         ::= <words> ':'
+words             ::= <word>
+                    | <words> '-' <word>
+projection        ::= '/' <label>
 interfaceversion  ::= '@' <valid semver>
                     | '@' <canonversion> 🔗
 canonversion      ::= [1-9] [0-9]* 🔗
                     | '0.' [1-9] [0-9]* 🔗
                     | '0.0.' [1-9] [0-9]* 🔗
 semversuffix      ::= [0-9A-Za-z.+-]* 🔗
-depname       ::= 'unlocked-dep=<' <pkgnamequery> '>'
-                | 'locked-dep=<' <pkgname> '>' ( ',' <hashname> )?
-pkgnamequery  ::= <pkgpath> <verrange>?
-pkgname       ::= <pkgpath> <pkgversion>?
-pkgversion    ::= '@' <valid semver>
-pkgpath       ::= <namespace> <words>
-                | <namespace>+ <words> <projection>* 🪺
-verrange      ::= '@*'
-                | '@{' <verlower> '}'
-                | '@{' <verupper> '}'
-                | '@{' <verlower> ' ' <verupper> '}'
-verlower      ::= '>=' <valid semver>
-verupper      ::= '<' <valid semver>
-urlname       ::= 'url=<' <nonbrackets> '>' (',' <hashname>)?
-nonbrackets   ::= [^<>]*
-hashname      ::= 'integrity=<' <integrity-metadata> '>'
+depname           ::= 'unlocked-dep=<' <pkgnamequery> '>'
+                    | 'locked-dep=<' <pkgname> '>' ( ',' <hashname> )?
+pkgnamequery      ::= <pkgpath> <verrange>?
+pkgname           ::= <pkgpath> <pkgversion>?
+pkgversion        ::= '@' <valid semver>
+pkgpath           ::= <namespace> <words>
+                    | <namespace>+ <words> <projection>* 🪺
+verrange          ::= '@*'
+                    | '@{' <verlower> '}'
+                    | '@{' <verupper> '}'
+                    | '@{' <verlower> ' ' <verupper> '}'
+verlower          ::= '>=' <valid semver>
+verupper          ::= '<' <valid semver>
+urlname           ::= 'url=<' <nonbrackets> '>' (',' <hashname>)?
+nonbrackets       ::= [^<>]*
+hashname          ::= 'integrity=<' <integrity-metadata> '>'
 ```
 Components provide six options for naming imports:
 * a **plain name** that leaves it up to the developer to "read the docs"


design/mvp/Explainer.md
```diff
@@ -294,7 +294,7 @@ sort           ::= core <core:sort>
                  | type
                  | component
                  | instance
-inlineexport   ::= (export <exportname> <sortidx>)
+inlineexport   ::= (export <exportname> <fullversion>? <sortidx>)
```
lann, line 297: Is `<fullversion>` useful here?

lann:
For the binary encoding the most straightforward option from a quick review would seem to be adding variants of [`importname'` / `exportname'`](https://github.com/WebAssembly/component-model/blob/2f8eda87c0029695e48c33e5ac3ef039eced1556/design/mvp/Binary.md?plain=1#L376-L377) along the lines of:

```
importname' ::= 0x00 len:<u32> in:<importname>                       => in  (if len = |in|)
              | 0x01 len:<u32> in:<importname> fullverlen:<u16> fullver:<valid semver>
```

I suppose if we wanted to optimize the binary a bit this extra field could contain _just_ the part of the original version that got lopped off by canonicalization.

On this field width:

> `fullverlen:<u16>`

https://semver.org/#does-semver-have-a-size-limit-on-the-version-string

> No, but use good judgment. A 255 character version string is probably overkill, for example. Also, specific systems may impose their own limits on the size of the string.

🤷

design/mvp/Explainer.md
```diff
@@ -2242,8 +2242,9 @@ the identifier `$x`). In the case of exports, the `<id>?` right after the
 preceding definition being exported (e.g., `(export $x "x" (func $f))` binds a
 new identifier `$x`).
 ```ebnf
-import ::= (import "<importname>" bind-id(<externdesc>))
-export ::= (export <id>? "<exportname>" <sortidx> <externdesc>?)
+import      ::= (import "<importname>" <fullversion>? bind-id(<externdesc>))
+export      ::= (export <id>? "<exportname>" <fullversion>? <sortidx> <externdesc>?)
+fullversion ::= (fullversion <valid semver>)
```
```suggestion
fullversion ::= (fullversion "<valid semver>")
```

design/mvp/Explainer.md
```suggestion
inlineexport   ::= (export "<exportname>" <fullversion>? <sortidx>)
```
(pre-existing, but since we're touching this line)

design/mvp/Explainer.md
```diff
@@ -574,8 +574,8 @@ instancedecl  ::= core-prefix(<core:type>)
                 | <alias>
                 | <exportdecl>
                 | <value> 🪙
-importdecl    ::= (import <importname> bind-id(<externdesc>))
-exportdecl    ::= (export <exportname> bind-id(<externdesc>))
+importdecl    ::= (import <importname> <fullversion>? bind-id(<externdesc>))
+exportdecl    ::= (export <exportname> <fullversion>? bind-id(<externdesc>))
```
```suggestion
importdecl    ::= (import "<importname>" <fullversion>? bind-id(<externdesc>))
exportdecl    ::= (export "<exportname>" <fullversion>? bind-id(<externdesc>))
```

lukewagner: (pre-existing)

lukewagner:
@lann Thanks for starting this!  For the binary encoding question: yes, taking over the `0x00` byte and using it as a discriminant is a nice coincidence we can take advantage of (and could you update the corresponding bullet in the "Warts" section at the end)?

> I suppose if we wanted to optimize the binary a bit this extra field could contain just the part of the original version that got lopped off by canonicalization.

Is there a simplicity argument to be made that requiring the concatenation of the `version` and the `fullversion` to match `<valid semver>` is simpler than allowing the `fullversion` to be `<valid semver>` and then adding the additional validation requirement (which I assume we want) that the `fullversion` has to "match" the `version`?  If so, that could be a second argument in favor in addition to size.

design/mvp/Explainer.md
```diff
@@ -2379,6 +2383,33 @@ interpreted with the same [semantics][SemVerRange]. (Mostly this
 interpretation is the usual SemVer-spec-defined ordering, but note the
 particular behavior of pre-release tags.)
 
+The `version` production used in `interfacename`s accepts both `valid semver`
```
lukewagner: Could you frame "canonicalized interface version" as a validation rule factored out into a new "Canonical Interface Name" section alongside and symmetric to "Name Uniquness" and say that it is temporarily not enforced but will start issuing warnings and be enforced post-Preview-3?

design/mvp/Explainer.md
```diff
@@ -2283,10 +2284,13 @@ words         ::= <word>
                 | <words> '-' <word>
 projection    ::= '/' <label>
 version       ::= '@' <valid semver>
```
lukewagner: Probably good to rename this `interfaceversion` since it's specific to `interfacename` (and to be symmetric to `pkgversion`).  (I know that'll mess up the column alignment and fixing bloats the diff obscuring the change; maybe leave it unaligned and fix it right before merging.)

alexcrichton:
For the binary encoding, here's another possible encoding:

```
importname' ::= 0x00 len:<u32> in:<importname>                       => in  (if len = |in|)
              | 0x01 len:<u32> in:<importname>                       => "${in.name}@N"  (if len = |in|,  in.version = N.*)
              | 0x02 len:<u32> in:<importname>                       => "${in.name}@0.N"  (if len = |in|,  in.version = 0.N.*)
              | 0x03 len:<u32> in:<importname>                       => "${in.name}@0.0.N"  (if len = |in|,  in.version = 0.0.N.*)
```

maybe with affordances for rc/etc unsure. The basic idea though is that the actual import name would always be `foo:bar/baz@0.1.2` in the binary format but the semantic meaning (e.g. the text format) would be a subslice of such a string. This codifies that in the binary format it's always a valid semver and the discriminant byte says basically how to shorten it. The goal here would be to make the binary format still pretty clear what it can be without changing the meaning of the meaning at a parsed layer.

>> fullverlen:<u16>
>
> https://semver.org/#does-semver-have-a-size-limit-on-the-version-string
>
>>    No, but use good judgment. A 255 character version string is probably overkill, for example. Also, specific systems may impose their own limits on the size of the string.

For this I'd recommend using `<u32>` regardless. We already limit many strings far below the theoretical 4G limit with a 32-bit length and keeping `<u32>` makes it more consistent with the rest of the decoding process. Otherwise when implementing a decoder you'd have to implement a specific function for decoding a 16-bit LEB which is otherwise not required when parsing WebAssembly today. Basically while I agree that >255 characters for a version is silly, I'd say that for consistency with the rest of the binary format this'd want to be `<u32>` if we go with this variant.

lukewagner:
@alexcrichton Good idea; that cleanly answers some of the questions above.  My only light concern is that tools might just treat the `<importname>` as the name and miss the nuance of chopping off parts of the versions.  I suppose tests and common low-level tools could catch/factor-out most of this though.  But if we go this direction: I suppose technically we don't even need the {`0`, `1`, `2`, `3`} opcode; it could just be derived from the full `<valid semver>` string, making version canonicalization a binary encoding detail.  Thoughts?

alexcrichton:
I agree yeah there's risk since the name in the binary format is "so simple", but yeah that's also where I'd hope that tests could weed things out. It'd be pretty simple in parser libraries I'd imagine to avoid exposing the full name as the import name if the discriminant was present.

My thinking though was that the name always has a full and valid semver, as defined by semver itself. That way the discriminant says what the "real" import name is (e.g. chopping off other stuff) for linking/semantic purposes. Although I may be misunderstanding what you're thinking about how to drop the discriminant?

lann:
I think @lukewagner is suggesting that the differences between 1/2/3 can be derived from the string itself. The algo would be something like:

- starting at `@`:
- if the string between `@` and the first `.` isn't `0`, trim before the first `.`
- if the string between `@` and the second `.` isn't `0.0`, trim before the second `.`
- otherwise, trim immediately after any digits after the second `.` (which should only be a `-` or `+`)

alexcrichton:
Ah I see! So something like (as a transition to the future):

```
importname' ::= 0x00 len:<u32> in:<importname>  => in  (if len = |in|)
              | 0x01 len:<u32> in:<importname>  => "${in.name}@${in.canonver}"  (if len = |in|)
```

where in the future we'd drop 0x00 entirely (and possibly rename 0x01 to 0x00). The `<importname>` is always required to have a full and valid semver too?

lann:
I think of the "semver-aware" options I prefer @lukewagner's 1 (extra) discriminant option; if you are parsing semver anyway then the logic is only marginally more complex than the 3 discriminant option.

I'm more ambivalent on whether the parser should be semver-aware. I like the conceptual simplicity of "the name is the name" but this _is_ a binary encoding and if we're going to require validation of semver then we're probably already committing to most of that _code_ complexity anyway.

lann:
We discussed this in a meeting today and decided to simplify a bit:

- In the text format `fullversion` will change to `versionsuffix` and hold just the part of the full version that is removed by canonicalization
- The binary format will use two strings: the canonicalized import name and the `versionsuffix`

lann:
After spending way too much time staring at SemVer and SemVer accessories I have a new draft of the explainer changes. I ran out of time to edit so hopefully it's still coherent...

I should be able to get to the binary format changes tomorrow.

lann:
A draft of the binary format changes is in. Again I didn't have time to edit today but I think its probably close enough for another review.

design/mvp/Binary.md
```diff
@@ -399,7 +402,10 @@ Notes:
   `(result (own $R))`, where `$R` is the resource labeled `r`.
 * Validation of `[method]` names requires the first parameter of the function
   to be `(param "self" (borrow $R))`, where `$R` is the resource labeled `r`.
-* `<valid semver>` is as defined by [https://semver.org](https://semver.org/)
```
lann, line 402: `<valid semver>` wasn't referenced in this file.

design/mvp/Explainer.md
```diff
@@ -2276,17 +2278,23 @@ fragment      ::= <word>
...
 words         ::= <word>
                 | <words> '-' <word>
 projection    ::= '/' <label>
-version       ::= '@' <valid semver>
+# FIXME: surrounding alignment
+interfaceversion ::= '@' <valid semver>
+                   | '@' <canonversion>
+canonversion     ::= [1-9] [0-9]*
+                   | '0.' [1-9] [0-9]*
+                   | '0.0.' [1-9] [0-9]*
```
lann: There is some ambiguity here; `'0.0.' [1-9] [0-9]*` matches a subset of `<valid semver>`. I think this makes sense in the context of non-canonical versions getting phased out but could add more to clarify if needed.

design/mvp/Explainer.md
```diff
@@ -2276,17 +2278,23 @@ fragment      ::= <word>
...
-interfacename ::= <namespace> <label> <projection> <version>?
-                | <namespace>+ <label> <projection>+ <version>? 🪺
+interfacename ::= <namespace> <label> <projection> <interfaceversion>?
+                | <namespace>+ <label> <projection>+ <interfaceversion>? 🪺
 namespace     ::= <words> ':'
 words         ::= <word>
                 | <words> '-' <word>
 projection    ::= '/' <label>
-version       ::= '@' <valid semver>
+# FIXME: surrounding alignment
```
lann: TODO after final review

design/mvp/Explainer.md
```diff
@@ -2242,8 +2243,9 @@ the identifier `$x`). In the case of exports, the `<id>?` right after the
 preceding definition being exported (e.g., `(export $x "x" (func $f))` binds a
 new identifier `$x`).
 ```ebnf
-import ::= (import "<importname>" bind-id(<externdesc>))
-export ::= (export <id>? "<exportname>" <sortidx> <externdesc>?)
+import        ::= (import "<importname>" <versionsuffix>? bind-id(<externdesc>))
+export        ::= (export <id>? "<exportname>" <versionsuffix>? <sortidx> <externdesc>?)
+versionsuffix ::= (versionsuffix "<semver suffix>")
```
lukewagner: Should `<semver suffix>` be defined below (it doesn't seem to be on semver.org)?  If so, then could it be `semversuffix` (since we otherwise don't use spaces for production names defined by this doc).

design/mvp/Explainer.md
lukewagner: Good point and I think you're right; it's fine if we word the validation rules in terms of "matching `canonversion`" as suggested above.

design/mvp/Explainer.md
```diff
@@ -2539,6 +2547,46 @@ annotations. For example, the validation rules for `[constructor]foo` require
 for details.
 
 
+### Canonical Interface Name
```
lukewagner: Can you find the place in the preceding "Import and export definitions" section where `valid semver` is mentioned and update that paragraph to link down to this new section?

design/mvp/Explainer.md
```diff
@@ -2539,6 +2547,46 @@ annotations. For example, the validation rules for `[constructor]foo` require
...
+
+- has no `interfaceversion`
+- has an `interfaceversion` matching the `canonversion` production
+
+The purpose of `canonversion` is to simplify the matching of compatible import
+and export versions. For example, if a guest imports some interface from
+`wasi:http/types@0.2.1` and a host provides the (subtype-compatible) interface
+`wasi:http/types@0.2.6`, we'd like to make it easy for the host to link with the
+guest. The `canonversion` for both of these interfaces would be `0.2`, so this
+linking could be done by matching canonical interface names literally.
```
lukewagner, line 2574: Could you also mention the more-nuanced "forward compatibility" case (a component imports the new version, but uses only the functions present in the hold version with compatible types)?

design/mvp/Binary.md
```diff
@@ -399,7 +402,10 @@ Notes:
...
+* Validation requires that `versionsuffix` is preceded by an `interfaceversion`
+  with a `canonversion` and that the concatenation of the `canonversion` and the
```
```suggestion
  matches `canonversion` and that the concatenation of the `canonversion` and the
```

lukewagner (commented):
Great, thanks!  Just a few small comments:

design/mvp/Explainer.md
lann: Sure. I was trying to avoid pulling in semver syntax but I guess this can at least give the valid character set and leave full validation to prose.

design/mvp/Explainer.md
```diff
@@ -2276,17 +2278,24 @@ fragment      ::= <word>
...
                 | <words> '-' <word>
 projection    ::= '/' <label>
-version       ::= '@' <valid semver>
+# FIXME: surrounding alignment
+interfaceversion  ::= '@' <valid semver>
+                    | '@' <canonversion>
+canonversion      ::= [1-9] [0-9]*
+                    | '0.' [1-9] [0-9]*
+                    | '0.0.' [1-9] [0-9]*
+semversuffix      ::= [0-9A-Za-z.+-]*
```
lann: This just checks for valid semver characters. The "real" validation is covered in prose below, i.e. `<canonversion><semversuffix>` must match `valid semver`.

I had a version of this as `[.+-][0-9A-Za-z.+-]+` which was less ambiguous but also implied that `versionsuffix` and the binary `0x01` variant couldn't be used with an empty suffix like you'd get with full version `0.0.1`. Given that we're saying we'll remove non-canonical interface versions I think allowing the empty suffix will ultimately be simpler.

design/mvp/Explainer.md
```diff
@@ -2539,6 +2550,49 @@ annotations. For example, the validation rules for `[constructor]foo` require
...
+
+The purpose of `canonversion` is to simplify the matching of compatible import
+and export versions. For example, if a guest imports some interface from
+`wasi:http/types@0.2.1` and a host provides the (subtype-compatible) interface
+`wasi:http/types@0.2.6`, we'd like to make it easy for the host to link with the
+guest. The `canonversion` for both of these interfaces would be `0.2`, so this
+linking could be done by matching canonical interface names literally.
+Symmetrically, if a host provides `wasi:http/types@0.2.1` and a guest imports
+`wasi:http/types@0.2.6`, so long as the guest's imported `instancetype` is
+subtype-compatible with the host implementation linking should be trivial.
```
```suggestion
`wasi:http/types@0.2.6`, so long as the guest only uses the subset of
functions defined in `wasi:http/types@0.2.1` (which is checked by normal
component type validation), linking succeeds. Thus, including only the
canonicalized version in the name allows both backwards and (limited) 
forwards compatibility using only trivial string equality (as well as the
type checking already required).
```

lann:
> validation ignores "versionsuffix"

In the current iteration `versionsuffix` is validated: https://github.com/lann/component-model/blob/fcb2152b85816e923b7121530818bfe8f627dc60/design/mvp/Binary.md?plain=1#L405-L408

> a short note in the "Type Checking" section of Explainer.md

Done: https://github.com/lann/component-model/blob/fcb2152b85816e923b7121530818bfe8f627dc60/design/mvp/Explainer.md?plain=1#L992-L994

lukewagner:
> In the current iteration versionsuffix is validated

Oops, you're right; I should have said "ignored when judging type compatibility of `instance` and `component` types" or something like that.

lann:
From out-of-band discussion, we're going to emoji-gate and merge this before the implementation is done.

lann:
Final two commits [emoji-gate this feature](https://github.com/WebAssembly/component-model/pull/536/commits/1fe35e736b46ef05d607a4a72e6d233ff2274214) and [fix ebnf alignment](https://github.com/WebAssembly/component-model/pull/536/commits/55f8d970e40adac380736fec400e7ac24ec6b6e9).

lukewagner:
Thanks again for the idea and all the work writing it up!

PR #537, opened 2025-06-26
fix: add ReadableStream write fn

vados-cosmonic:
`ReadableStream.write` appeared to be missing from the docs -- it should work near identically to `read` (except for internal implementation of course), but wasn't called out.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1279,11 +1279,12 @@ OnCopyDone = Callable[[CopyResult], None]
 class ReadableStream:
   t: ValType
   read: Callable[[ComponentInstance, WritableBuffer, OnCopy, OnCopyDone], None]
+  write: Callable[[ComponentInstance, WritableBuffer, OnCopy, OnCopyDone], None]
   cancel: Callable[[], None]
   drop: Callable[[], None]
 ```
 The key operation is `read` which works as follows:
-* `read` never blocks and returns its values by either synchronously or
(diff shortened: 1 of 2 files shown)


lukewagner:
Thanks for the careful read.  The reason `write` isn't in `ReadableStream` is that a `ReadableStream` that comes from the host really doesn't have a `write` -- you can only `read` it (see [this comment](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md?plain=1#L1406-L1410)).  Now I'm guessing you're looking at the fact that the `shared` field of `StreamEnd` has type `ReadableStream` while the `WritableStreamEnd` subclass calls `write` on it.  This weirdness is addressed in [this comment](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md?plain=1#L1496-L1500).  If there were a `WritableStream` interface (symmetric to `ReadableStream`), we could say the type of `shared` is `ReadableStream|WritableStream` (which would still be a little imprecise, for lack of covariance, but better).  I didn't do that before b/c it didn't seem worth it just for this one type annotation, but thinking about it again, probably it's a good idea and it could be added to the base-class list of `SharedStreamImpl` alongside `ReadableStream` which would be useful too.  I'll do that in a separate PR (b/c it'll probably involve a bit of prose refactoring).

vados-cosmonic:
Ah thank you for the explanation -- I certainly added this because of seeing the `ReadableStream` being in there! I'll let you get to updating that when you're ready :)

PR #538, opened 2025-06-26
chore: rename revoke_buffer to reclaim_buffer

vados-cosmonic:
This commit renames `revoke_buffer` which is used during stream copying to `reclaim_buffer` which hopefully better conveys (if only slightly) the handoff/change in ownership of the buffer in use.


lukewagner:
Thanks!  Could you also update the `RevokeBuffer` function type to match?

vados-cosmonic:
Ah thanks @lukewagner -- done!

lukewagner:
Thanks again!

PR #541, opened 2025-07-07
Add more documentation for `stream.{read,write}`.

sunfishcode:
Notably:

 - Document what stream.{read,write} do after a `copy-result.dropped`, similar to what is documented for future.{read,write}.

 - Document that `copy-result.dropped` does not signify an error.

 - Reword "the only valid next operation is" to "anything subsequent operation [...] traps" to avoid confusion about what "valid" means in this context.

 - Say "single-element" instead of "length-1" so that it doesn't risk scanning as "length minus one".


lukewagner (approved):
Looks great, thanks!

PR #545, opened 2025-07-14
Fix the name of `trap_if_on_the_stack`.

sunfishcode:
Fix a reference to `trap_if_on_the_stack` to use the correct name.


lukewagner:
Thanks!

PR #551, opened 2025-08-11
Fix typo in use of task.return

wingo:
If I understand correctly, `task.return` for a string should take two parameters and return no values.


lukewagner:
d'oh, thanks!

Issue #125, opened 2022-11-08
Dictionaries


guybedford:
Are dictionaries on the roadmap for the component model? At the moment it seems this requires representation via `list<tuple<string, ...>>`, where a dictionary could be a relatively straightforward host-level representation over such a datatype.

lukewagner:
That's a good question and indeed this is one of those types on the edge.  Some reasons why it seems like we should hold off for now:
* We have `record` types which subsume a lot of what would otherwise be dictionary use cases, particularly with `option` fields, while providing much more declarative interface information and a more-efficient representation. To wit, the Web IDL [`dictionary`](https://webidl.spec.whatwg.org/#idl-dictionaries) type is effectively a `record` and used extensively in Web APIs, while the Web IDL [`record`](https://webidl.spec.whatwg.org/#idl-record) type is effectively a dictionary and used relatively infrequently.
* It would be a bunch of tricky design and implementation work, since there's a wide design and implementation space of options (esp. around sortedness/uniqueness).

Thus, as with #56, while I can see the use case, I'd like to see if we can get by without these in an MVP release.

guybedford:
I would of course tend towards JS semantics here, so records having unique strings and ordering supported (ie exactly a `list<tuple<string, ...>>` representation underneath). In theory uniqueness could be managed by the host, in the same way valid USV strings are managed by the host without requiring a check everytime, since it's a strong property of the nature of the data it returns.

I can certainly get by with a `list`, but my concern is that it's a wide enough pattern that there's a natural temptation to want to decorate the interface with a JS object to give a nice representation - which then forces me back outside of the component model to a space of component model + sprinkles of JS, which seems non-ideal.

lukewagner:
Agreed.  And just explaining a bit more about the design challenges:

It makes sense that when passing in a host value, the host can already have guarantees of uniqueness; the complexity comes when the Canonical ABI lifts a dictionary from linear memory since the host can't trust anything and thus must verify things dynamically every time.

For ordering: requiring ordered keys (in the `list<tuple<string, ...>>`) is great for lifting performance (each string can be compared with its predecessor, fused with the copy loop) but JS objects and maps (and, in general, any hash-table-based dictionary) aren't lexicographically ordered, so this would add a non-trivial sorting operation when lowering these into wasm (negating any win from knowing uniqueness).  But if we drop the ordering requirement, detecting dupes when lifting linear memory is rather more expensive.

[...]

Issue #383, opened 2024-07-25
Lazy value lowering


lukewagner:
This isn't a fully-fleshed out idea, but it's coming up in #369 and so I thought it might be nice to split out here.  I'm tentatively excited about it, though (in that "this is probably what we should've done in the first place... sigh" sort of way).

## Motivation

When lowering a dynamically-sized value (`list` or `string`) or lowering more than `MAX_FLAT_PARAMS`/`RESULTS`, the Canonical ABI specifies calling a wasm-defined allocation function (indicated per import/export via the `realloc` immediate on `canon lift`/`lower`, but usually there's just one exported with the name `cabi_realloc`) to get the linear memory to lower the `list`/`string` into.  This works decently well, but sometimes it's constraining, leading to [various](https://github.com/WebAssembly/component-model/issues/148#issuecomment-1378072049) [ideas](https://github.com/WebAssembly/component-model/issues/369#issuecomment-2247088561) for further customizations, with more ideas brewing.  But the root problem seems to be the control flow inversion of the adapter controlling the calls to `realloc`, which limits wasm's control of the order and manner of allocation.

## Idea

In all the places where the CABI currently says to call `realloc` and store the resulting `i32` pointer, instead we could say that the CABI stores a fresh `i32` index that refers to the abstract value-to-be-lowered.  The `i32` length (of `list`s and `string`s) is stored eagerly as usual.  Control flow is then handed to wasm, passing it this shallowly-lowered tuple of params/results.  Wasm can now look at the `i32` length values to see how much to allocate and then acquire that memory however it wants; if there are multiple values-to-be-lowered, wasm can do smart things like batching allocations.

Once the destination address is selected, wasm then calls a new built-in to lower the value, e.g.:
* `canon value.lower $t: [validx:i32 dstp:i32] -> []`

For a `list<list<u8>>`, the first call to `value.lower` will generate N more `validx`s for each `list<u8>` element, which the calling wasm again gets to control the allocation and lowering of (and so on, recursively through the type structure).

We could also allow multiple partial reads of the same `validx` so that a single logical `list` value could be read into multiple non-contiguous segments (think `readv()`).  Once fully lowered, a `validx` will trap if lowered again.

[...]

PR #554, opened 2025-08-14
Add a 'map<K,V>' type

lukewagner:
This PR adds a `map<K,V>` to the set of Component Model and WIT value types as discussed in #125, specifically based on [this design](https://github.com/WebAssembly/component-model/issues/125#issuecomment-3129960033) by @yordis.  With the approach of treating `map<K,V>` as just a specialization of `list<tuple<K,V>>`, the addition is pretty small and most of the work will be in the bindings generators.  I'm not in a particular rush to add this, but given the interest in #125, it seems useful to have a detailed proposal that can be prototyped and then potentially merged.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -792,12 +792,14 @@ this can sometimes allow values to be represented differently. For example,
 of boolean fields uses a sequence of boolean-valued bytes.
 
 Since a `map` is a specialization of a list of (key, value) pairs without any
-additional semantic guarantee of key uniqueness, the Component Model does not
-forcibly prevent duplicate keys from appearing in the list. In the case of
-duplicate keys, the expectation for bindings generators is that for any given
-key, the *last* (key, value) pair in the list defines the value of the key in
-the map. To simplify bindings generation, `<keytype>`s is a conservative subset
-of `<valtype>`, but this subset could be expanded over time based on use cases.
+additional semantic guarantee of key uniqueness or ordering, the Component Model
+does not prevent duplicate keys from appearing in the list. Bindings generators
+*may* deduplicate and reorder keys as long as the *last* (key, value) pair in the
+original list defines the final value of the key. However, bindings generators
+are not *required* to deduplicate keys and may instead simply present the
+original list to guest code. To simplify bindings generation, `<keytype>` is a
+conservative subset of `<valtype>`, but this subset could be expanded over time
+based on use cases.
 
 Note that, at least initially, variants must have a non-empty list of
 cases. This could be relaxed in the future to allow an empty list of cases, with
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -58,6 +58,7 @@ implemented, considered stable and included in a future milestone:
 * 📝: the `error-context` type
 * 🔗: canonical interface names
 * 🐘: [memory64]
+* 🗺️: the `map` type
 
 (Based on the previous [scoping and layering] proposal to the WebAssembly CG,
 this repo merges and supersedes the [module-linking] and [interface-types]
@@ -563,12 +564,14 @@ defvaltype    ::= bool
                 | (enum "<label>"+)
                 | (option <valtype>)
                 | (result <valtype>? (error <valtype>)?)
+                | (map <keytype> <valtype>) 🗺️
                 | (own <typeidx>)
                 | (borrow <typeidx>)
                 | (stream <valtype>?) 🔀
                 | (future <valtype>?) 🔀
 valtype       ::= <typeidx>
                 | <defvaltype>
+keytype       ::= bool | s8 | u8 | s16 | u16 | s32 | u32 | s64 | u64 | char | string 🗺️
 resourcetype  ::= (resource (rep i32) (dtor <core:funcidx>)?)
                 | (resource (rep i32) (dtor async <core:funcidx> (callback <core:funcidx>)?)?) 🚝
                 | (resource (rep i64) (dtor <core:funcidx>)?) 🐘
@@ -772,6 +775,7 @@ defined by the following mapping:
                     (option <valtype>) ↦ (variant (case "none") (case "some" <valtype>))
 (result <valtype>? (error <valtype>)?) ↦ (variant (case "ok" <valtype>?) (case "error" <valtype>?))
                                 string ↦ (list char)
+             (map <keytype> <valtype>) ↦ (list (tuple <keytype> <valtype>))
 ```
 
 Specialized value types have the same set of semantic values as their
@@ -786,6 +790,14 @@ this can sometimes allow values to be represented differently. For example,
 `list<char>` uses a sequence of 4-byte `char` code points.  Similarly,
 `flags` in the Canonical ABI uses a bit-vector while an equivalent record
 of boolean fields uses a sequence of boolean-valued bytes.
+
+Since a `map` is a specialization of a list of (key, value) pairs without any
+additional semantic guarantee of key uniqueness, the Component Model does not
+forcibly prevent duplicate keys from appearing in the list. In the case of
+duplicate keys, the expectation for bindings generators is that for any given
+key, the *last* (key, value) pair in the list defines the value of the key in
+the map. To simplify bindings generation, `<keytype>`s is a conservative subset
+of `<valtype>`, but this subset could be expanded over time based on use cases.
 
 Note that, at least initially, variants must have a non-empty list of
 cases. This could be relaxed in the future to allow an empty list of cases, with
@@ -3036,6 +3048,7 @@ At a high level, the additional coercions would be:
 | `enum` | same as [`enum`] | same as [`enum`] |
 | `option` | same as [`T?`] | same as [`T?`] |
 | `result` | same as `variant`, but coerce a top-level `error` return value to a thrown exception | same as `variant`, but coerce uncaught exceptions to top-level `error` return values |
+| `map` | `new Map(_)` | `Map`s directly or other objects via `Object.entries(_)` |
 | `own`, `borrow` | see below | see below |
 | `future` | to a `Promise` | from a `Promise` |
 | `stream` | to a `ReadableStream` | from a `ReadableStream` |
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -1021,6 +1021,7 @@ keyword ::= 'as'
           | 'include'
           | 'interface'
           | 'list'
+          | 'map'
           | 'option'
           | 'own'
           | 'package'
@@ -1715,6 +1716,7 @@ ty ::= 'u8' | 'u16' | 'u32' | 'u64'
      | list
      | option
      | result
+     | map
      | handle
      | future
      | stream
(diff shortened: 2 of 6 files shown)


design/mvp/Explainer.md
```diff
@@ -2793,6 +2805,7 @@ At a high level, the additional coercions would be:
 | `enum` | same as [`enum`] | same as [`enum`] |
 | `option` | same as [`T?`] | same as [`T?`] |
 | `result` | same as `variant`, but coerce a top-level `error` return value to a thrown exception | same as `variant`, but coerce uncaught exceptions to top-level `error` return values |
+| `map` | `new Map(_)` | `Map`s directly or other objects via `Object.entries(_)` |
```
yordis, line 3053: `ToJSValue` could a plain JS `object` as well, not sure if matters here

lukewagner: That's worth discussing with JS folks, but the hazard I was remembering with plain JS objects is that, when used for arbitrary keys, there can be conflicts with built-in properties (e.g., `__proto__`), which was one of the original motivations for adding `Map`.  (Also, objects only work with the `keytype` is `string`.)

design/mvp/WIT.md
```diff
@@ -1772,6 +1779,12 @@ variant result {
 These types are so frequently used and frequently have language-specific
 meanings though so they're also provided as first-class types.
 
+🗺️ The `map` type is semantically equivalent to a list of pairs of keys and
+values but is meant to be represented by bindings generators in the source
+language as a mapping from keys to values (e.g., as an object, map, dictionary
```
ouillie: I don't think we should be encouraging the "as an object" interpretation. This is mainly relevant to JS, where an object can only have string-typed keys. I know that's an initial restriction to only support strings, but I think it should be clear that that's subject to change and the intent is to implement general maps.

See also https://github.com/WebAssembly/component-model/issues/125#issuecomment-2635803031

lukewagner: Yeah, that's a good point; [updated](https://github.com/WebAssembly/component-model/pull/554/commits/6661b85a2882191300e34b4d5fb343a6ea5c5c92).

design/mvp/Explainer.md
```diff
@@ -780,6 +784,14 @@ this can sometimes allow values to be represented differently. For example,
 `flags` in the Canonical ABI uses a bit-vector while an equivalent record
 of boolean fields uses a sequence of boolean-valued bytes.
 
+Since a `map` is a specialization of a list of (key, value) pairs without any
+additional semantic guarantee of key uniqueness, the Component Model does not
+forcibly prevent duplicate keys from appearing in the list. In the case of
+duplicate keys, the expectation for bindings generators is that for any given
+key, the *last* (key, value) pair in the list defines the value of the key in
+the map. To simplify bindings generation, `<keytype>`s is a conservative subset
```
lann: > the expectation for bindings generators

Can we find a way to make this language stronger? Imagine some "authorization middleware" component sits on an interface and inspects a call with a map param. If the value `{"action": "view", "action": "delete"}` comes through it is going to be very important that the middleware treats the map exactly the same as the next component.

lukewagner: Yeah, I suppose "expectation" is too weak; would it make sense to say "Although the Component Model cannot enforce this property, bindings generators **MUST** ..."?

primoly: I see two things about the current `map` specification that can lead to problems.

First, allowing duplicate keys and expect bindings to just ignore the last occurrence. I think uniqueness of keys should be mandated by the spec and enforced at the boundary.

The second is that the order of entries is propagated, but some languages’ (standard) `Map` type(s) will  not preserve this. In my opinion, it should be expected that bindings will map `map` to a type that retains order. It would then also be a good idea to rename to something else (e.g. `dict`, `ordered-map`) so there is less chance of confusion with a conventional (non order preserving) `Map` type.

lukewagner: Yeah, it's a tradeoff to be sure.  However, if `map` forces the runtime to build a temporary hash set (to enforce uniqueness) that is pure overhead (and potentially a pretty non-trivial runtime-internal memory allocation, which we otherwise avoid in the CABI), interface designers will have to ask whether they can "afford" to use `map` or whether they should use `list<tuple<K,V>>` instead for performance reasons, which seems net worse.

ouillie: 1. I have no issue with mandating uniqueness in the spec. It's always easier to relax constraints than to tighten them, anyway, so might as well be more constrained up front. However, when that constraint is violated by a misbehaving lifting component, I think the behavior should be the same for the lowering component; just ignore all but the final value. I don't think we should do the C thing and call it undefined behavior (unless that's normal for the component model?) and I don't think the lowering should trap. So this is just a question of semantics rather than behavior.

2. I don't think the basic map type should be ordered. Most languages do not use an ordered basic map type because 9 times out of 10 you don't need an ordering for your maps. In the cases where you do need an ordered map, you could always fall back in `list<tuple<key, value>>`. I don't think it's super important to create a specialization for an ordered map as well, but that could be revisited later.

lann: Trying to balance a useful requirement with potential guest language constraints (does `wit-bindgen-c` _really_ need a hashtable impl?), maybe something along the lines of:

_Generated map interfaces MUST behave as if duplicate entries have been filtered out (last duplicate key value wins). Map values with existing duplicate entries MAY be passed to other components._

yordis: In most programming languages (Rust, Go, Elixir, Ruby, JavaScript, etc), map/hashmap structures overwrite the existing value on a put/insert operation. Key order is typically not guaranteed; when it is, it is more around the compiled-time declaration of the keys than actual order of insert operations.

I found that the 80% of cases when uniqueness or ordering is required, it’s left to the domain to define. Or very specialized data structures are created (uniq combination of `plain` + `order` + `uniqueness` 😭)

At such a low level, the performance overhead should be close to zero.

As other pointed out `list<tuple<key, value>>` is a safe fallback to create your own domain specific data structure enforcing such constraints (uniq combination of `plain` + `order` + `uniqueness` 😭)

Lastly, call it `map` or any other name, what I want is the behavior described here (no order, no uniqueness checks) althou, I think `map` is what most people in the industry would expect.

badeend: Couple of notes:

---

Agree that keys of `map`s should not guarantee any order. This aligns with [Rust's HashMap](https://doc.rust-lang.org/std/collections/struct.HashMap.html), [.NET's Dictionary](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?view=net-9.0), [Java's HashMap](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html), [Go's map](https://go.dev/blog/maps#iteration-order), and probably more.

---

Agree that the C-M host should not be _required_ to deduplicate keys. Though, I suggest that they are _allowed_ to. I like the prior suggestion in [#125](https://github.com/WebAssembly/component-model/issues/125#issuecomment-2641508050):
> One thing we should probably say, spec-wise, is that, just like how we non-deterministically allow but do not require NaN payloads to be scrambled (allowing, but not requiring, hosts to canonicalize NaNs at cross-component boundaries), we could likewise non-deterministically allow-but-not-require duplicate keys to be merged (with well-defined last-value-overwrites semantics when such merging occurs).

With that, `map<k, v>` and `list<tuple<k, v>>` are indeed identical at the binary level. But semantically they are different in that:
fully deterministic / security-sensitive C-M hosts MAY reorder and deduplicate keys in `map`s, but not on `list`s of `tuple`s.

---

In the case of duplicate keys, the PR currently requires a "last key wins" strategy. This mean that guests must first read through the entire collection into memory before being able to process the entries _safely_. If we go with a "first key wins" strategy, guests would be able to process entries as they come in, and skip duplicates later on in the list. The difference between the two strategies is probably negligible today, but could become more relevant with [lazy lowering](https://github.com/WebAssembly/component-model/issues/383)

yordis: about "last" vs. "first" write wins; I am not attached to any implementation; other than "last" is what I experienced pretty much my entire career

I leave that decision to people who are more capable than I am

lukewagner: @badeend Good points!  So if we had the semantics of lifting and lowering nondeterministically permute and discard duplicate keys and the deterministic profile subsets this by sorting and removing all duplicates (with deterministic selection of which value wins), then:
1. guest code wouldn't be allowed to depend on order or duplicates or uniqueness
2. impls could reorder/dedupe or not as they saw fit
3. the determinstic semantics is unordered

The CABI would still accept/produce an ordered list-of-pairs that might contain duplicates and a production implementation of component-to-component linking would likely still preserve order and duplicates, but guest code "shouldn't" depend on this, and perhaps hosts could having a testing/chaos mode that reordered/deduped in-between components to catch bugs.

Does that sound right?

lann: > deterministic profile subsets this by sorting 

Could you say more about why this would be helpful? It isn't immediately clear to me that it would be worth the runtime cost.

badeend: > Does that sound right?

Pretty much, yeah. 👍

lukewagner: @lann Since the deterministic profile can't randomly permute, if we don't normalize order in the deterministic profile, then that effectively makes order an observable part of the semantics of `map` values.  But yeah, I suppose in some cases the performance might be a problem, even for the deterministic profile, so it's a tradeoff worth discussing.

lann: Ah I see, you're talking about baking unorderedness (and presumably dedupedness?) into the map lowering semantics while I was only thinking of making the bindings generation guidelines language stronger.

I see the benefit of formalizing it but yeah, requiring sorting on every map lowering seems quite a different order of tradeoff than NaN canonicalization. :slightly_smiling_face:

Do you have any references on high-level motivations / use-cases for deterministic profiles? It seems difficult to evaluate this kind of tradeoff without that context.

lukewagner: The use case for the deterministic profile (introduced in the 3.0 draft [here](https://webassembly.github.io/spec/versions/core/WebAssembly-3.0-draft.pdf#subsubsection*.797) as part of the relaxed SIMD instructions) is just to define, if you want to run wasm deterministically, here's how to do it.  If we don't specify sort+dedupe, that means that, even when running deterministically, a component can produce different outputs for the input `{a:1, b:2}` vs `{b:2, a:1}` which means that these two values must be considered unequal if you're, e.g., caching outputs keyed by inputs.  That's a corollary, but I don't know how much of a problem it is.

lukewagner: Finally getting back to this: [this commit](https://github.com/WebAssembly/component-model/pull/554/commits/3bf6a04f993a02187790e8e47770051d98f8205a) tries to clarify the wording about what bindings generators are allowed/required to do, speaking to Lann's original concern.

I also played with the idea above of nondeterministically reordering/deduping the list-of-key-value-pairs in the lifting/lowering rules.  However, it seemed like more trouble than it was worth and forcing the deterministic profile to *always* sort/dedupe seemed like overkill, so for now I just left it as is.

lukewagner:
A bunch of great implementation and bindings generator work has happened in the meantime (nice work @yordis!), and since this features is properly gated, it seems reasonable to merge this PR soon (if there's no further feedback on the open thread above).  As always, we can iterate in new issues/PRs as `map` makes its way to being released.

PR #557, opened 2025-08-28
Add built-ins for creating/switching cooperative threads

lukewagner:
This PR extends the concurrency support in Preview 3 with the ability for core wasm to create and switch between "cooperative threads", providing fiber-like functionality that is exposed to core wasm as plain core function imports that are integrated with and built on the async machinery that is already part of Preview 3.

Cooperative threads are meant to be released in some 0.3.x minor version after 0.3.0, so they are given a separate emoji-gate than 0.3.0's 🔀.  Since cooperative threads pave the way for preemptive threads (which depend on [shared-everything-threads](https://github.com/webAssembly/shared-everything-threads)), the emoji-gate for cooperative threads is 🧵 and preemptive threads are switched to 🧵②.  The current emoji-gates 🚝 and 🚟 are folded into 🧵 since they're also post-0.3.0 and make sense only with cooperative threads but if there's still a reason to keep them separate, we can split them back out.

This feature is intended to cover both "host threads" (e.g., pthreads.h) and "green threads" (e.g. goroutines and Java virtual threads) use cases.  The former use case mostly only needs the ability to spawn a thread (with, in a cooperative setting, the choice between whether to run the new thread immediately or "later"). But the green-threads use case needs to take explicit control of switching between threads (instead of leaving it up to the wasm runtime), so several additional built-ins are added to do this such as `thread.switch-to` (which works like the [`switch`](https://github.com/WebAssembly/stack-switching/blob/main/proposals/stack-switching/Explainer.md#invoking-continuations) instruction in the stack-switching proposal).  For the summary of the built-ins, see [this section](https://github.com/WebAssembly/component-model/blob/threads/design/mvp/Concurrency.md#waiting).  Also see the rewritten [goals](https://github.com/WebAssembly/component-model/blob/threads/design/mvp/Concurrency.md#goals) and [summary](https://github.com/WebAssembly/component-model/blob/threads/design/mvp/Concurrency.md#summary) in `Concurrency.md` (formerly `Async.md`).

With the above new built-ins, `thread.spawn_indirect` becomes an optimized fusion of `thread.new_indirect`+`thread.resume-later`.  For simplicity, `thread.spawn_indirect` is kept gated to 🧵②, so that it's "the one that depends on `shared`" and `thread.new_indirect` is added by 🧵 and has no `shared` option for now.  As noted in the [Binary format TODO](https://github.com/WebAssembly/component-model/blob/threads/design/mvp/Binary.md#binary-format-warts-to-fix-in-a-10-release), as part of 1.0-rc, we'll need to add a `shared?` option to *every* existing built-in (so that they can all be called from preemptive threads) anyhow.  But this PR does change the binary format for `thread.spawn_indirect` (to include an optional `shared` immediate) so CC @abrown on these changes in [Binary.md](https://github.com/WebAssembly/component-model/blob/threads/design/mvp/Binary.md).  (It's fine to not implement them any time soon, but I think it's useful to see the sketch of how they'd eventually look.)

For consistency, the `yield` built-in is backwards-compatibly renamed `thread.yield` (keeping the opcode the same and keeping `yield` in the text format but marked "deprecated" until the transition is done).

What was previously called "context-local storage" in the explainer is now renamed to "thread-local storage" (since it's now literally stored per thread), but the `context.{get,set}` built-ins keep the same name since they still seem to make sense as names.  As planned earlier, the static length of the thread-local storage array is bumped from "1" to "2" b/c now there's a good reason to store the "linear memory shadow stack pointer" alongside the "general TLS struct pointer".

Although a bunch of the explanatory prose changes in this PR, the actual functional addition to what was already necessary for 0.3.0 is surprisingly little; there's mostly just this new ability to directly `thread.switch-to`.  Hopefully that remains true in the implementation.

I expect to keep this PR open for a while until someone with an implementor's hat on has a chance to look carefully, so no rush for folks to review.  I'd suggest reading the diff of `Concurrency.md` first then `Explainer.md` then `CanonicalABI.md`.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -214,7 +214,7 @@ label'        ::= len:<u32> l:<label>                     => l    (if len = |l|)
 valtype       ::= i:<typeidx>                             => i
                 | pvt:<primvaltype>                       => pvt
 resourcetype  ::= 0x3f 0x7f f?:<funcidx>?                 => (resource (rep i32) (dtor f)?)
-                | 0x3e 0x7f f:<funcidx> cb?:<funcidx>?    => (resource (rep i32) (dtor async f (callback cb)?)) 🚝
+                | 0x3e 0x7f f:<funcidx> cb?:<funcidx>?    => (resource (rep i32) (dtor async f (callback cb)?)) 🧵
 functype      ::= 0x40 ps:<paramlist> rs:<resultlist>     => (func ps rs)
 paramlist     ::= lt*:vec(<labelvaltype>)                 => (param lt)*
 resultlist    ::= 0x00 t:<valtype>                        => (result t)
@@ -288,14 +288,14 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift 
            | 0x01 0x00 f:<funcidx> opts:<opts>                   => (canon lower f opts (core func))
            | 0x02 rt:<typeidx>                                   => (canon resource.new rt (core func))
            | 0x03 rt:<typeidx>                                   => (canon resource.drop rt (core func))
-           | 0x07 rt:<typeidx>                                   => (canon resource.drop rt async (core func)) 🚝
+           | 0x07 rt:<typeidx>                                   => (canon resource.drop rt async (core func)) 🧵
            | 0x04 rt:<typeidx>                                   => (canon resource.rep rt (core func))
            | 0x08                                                => (canon backpressure.set (core func)) 🔀
            | 0x09 rs:<resultlist> opts:<opts>                    => (canon task.return rs opts (core func)) 🔀
            | 0x05                                                => (canon task.cancel (core func)) 🔀
            | 0x0a 0x7f i:<u32>                                   => (canon context.get i32 i (core func)) 🔀
            | 0x0b 0x7f i:<u32>                                   => (canon context.set i32 i (core func)) 🔀
-           | 0x0c cancel?:<cancel?>                              => (canon yield cancel? (core func)) 🔀
+           | 0x0c cancel?:<cancel?>                              => (canon thread.yield cancel? (core func)) 🔀
            | 0x06 async?:<async?>                                => (canon subtask.cancel async? (core func)) 🔀
            | 0x0d                                                => (canon subtask.drop (core func)) 🔀
            | 0x0e t:<typeidx>                                    => (canon stream.new t (core func)) 🔀
@@ -320,13 +320,21 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift 
            | 0x21 cancel?:<cancel?> m:<core:memidx>              => (canon waitable-set.poll cancel? (memory m) (core func)) 🔀
            | 0x22                                                => (canon waitable-set.drop (core func)) 🔀
            | 0x23                                                => (canon waitable.join (core func)) 🔀
-           | 0x40 ft:<typeidx>                                   => (canon thread.spawn_ref ft (core func)) 🧵
-           | 0x41 ft:<typeidx> tbl:<core:tableidx>               => (canon thread.spawn_indirect ft tbl (core func)) 🧵
-           | 0x42                                                => (canon thread.available_parallelism (core func)) 🧵
+           | 0x26                                                => (canon thread.index (core func)) 🧵
+           | 0x27 ft:<typeidx> tbl:<core:tableidx>               => (canon thread.new_indirect ft tbl (core func)) 🧵
+           | 0x28 cancel?:<cancel?>                              => (canon thread.switch-to cancel? (core func)) 🧵
+           | 0x29 cancel?:<cancel?>                              => (canon thread.suspend cancel? (core func)) 🧵
+           | 0x2a                                                => (canon thread.resume-later (core func)) 🧵
+           | 0x2b cancel?:<cancel?>                              => (canon thread.yield-to cancel? (core func)) 🧵
+           | 0x40 shared?:<sh?> ft:<typeidx>                     => (canon thread.spawn_ref shared? ft (core func)) 🧵②
+           | 0x41 shared?:<sh?> ft:<typeidx> tbl:<core:tableidx> => (canon thread.spawn_indirect shared? ft tbl (core func)) 🧵②
+           | 0x42 shared?:<sh?>                                  => (canon thread.available-parallelism shared? (core func)) 🧵②
 async?   ::= 0x00                                                =>
            | 0x01                                                => async
 cancel?  ::= 0x00                                                =>
-           | 0x01                                                => cancellable 🚟
+           | 0x01                                                => cancellable 🧵
+sh?      ::= 0x00                                                =>
+           | 0x01                                                => shared 🧵②
 opts     ::= opt*:vec(<canonopt>)                                => opt*
 canonopt ::= 0x00                                                => string-encoding=utf8
            | 0x01                                                => string-encoding=utf16
@@ -510,6 +518,8 @@ named once.
   repurposed.
 * Most built-ins should have a `<canonopt>*` immediate instead of an ad hoc
   subset of `canonopt`s.
+* Add optional `shared` immediate to all canonical definitions (explicitly or
+  via `<canonopt>`) when shared-everything-threads (🧵②) is added.
 
 
 [`core:byte`]: https://webassembly.github.io/spec/core/binary/values.html#binary-byte
--- a/design/mvp/Concurrency.md
+++ b/design/mvp/Concurrency.md
@@ -1,21 +1,18 @@
-# 🔀 Async Explainer
+# Concurrency Explainer
 
-*This explainer describes a feature that is part of the forthcoming "Preview 3"
-release of the Component Model. The relevant parts of the [AST explainer],
-[binary format] and [Canonical ABI explainer] are gated by the 🔀 emoji.*
-
-This explainer provides a high-level summary of the native async support in the
-Component Model. For a detailed presentation of the runtime semantics, see the
-[Canonical ABI explainer]. See also the [Wasm I/O 2024 presentation] for a
-summary of the motivation and animated sketch of the design in action.
+This document contains a high-level summary of the native concurrency support
+added as part of [WASI Preview 3], providing background for understanding the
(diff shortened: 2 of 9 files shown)


design/mvp/Concurrency.md
```diff
@@ -501,6 +543,12 @@ loop interleaving `stream.read`s (of the readable end passed for `in`) and
 `stream.write`s (of the writable end it `stream.new`ed) before exiting the
 task.
 
+*Any* of the threads contained by a task can call `task.return`; there is no
```
TartanLlama, line 613: What happens if multiple threads call `task.return`? Trap?

lukewagner: Yep (which is also what currently happens if a single thread calls `task.return` twice in succession).  I should probably explicitly mention the trap-on-multiple-calls above.

lukewagner: [Added](https://github.com/WebAssembly/component-model/pull/557/commits/5fcd4c1a90b3a5dac8a53c1fabc4b17e86f1294b)

design/mvp/Explainer.md
```diff
@@ -51,9 +50,8 @@ implemented, considered stable and included in a future milestone:
 * 🪙: value imports/exports and component-level start function
 * 🪺: nested namespaces and packages in import/export names
 * 🔀: async
-  * 🚝: marking some builtins as `async`
-  * 🚟: using `async` with `canon lift` without `callback` (stackful lift)
```
alexcrichton, line 55: Could you detail a bit more about folding these together? Reading over https://github.com/bytecodealliance/wasm-tools/pull/2298 it's not clear to me we'll want to lump them all together. For example async-stackful feels to me best left split out still. Also these seem all more async-related rather than threading-related (e.g. may wish to keep the separate gate)

* `resource.drop async` 
* `subtask.cancel async` 
* `{stream,future}.cancel-{read,write} async`

however changing these to threading seems fine:
 
* `thread.yield cancellable` (used to be `yield async`)
* `waitable-set.poll cancellable` 
* `waitable-set.wait cancellable`

lukewagner: Mostly I was just trying to minimize the combinatorial state space and I'm definitely happy to split things out if we get close to releasing a 0.3.x and some of these built-ins are still unimplemented or unused by language toolchains.  But to provide the rationale for why I think they will be:
* while there is the pthread.h coop threads use case which is only needs `sync` or `async callback`, for the "green thread" use cases (goroutines, Java Virtual Threads, Kotlin coroutines) that need `thread.switch`, I believe they'll naturally want stackful `async`
* when using stackful `async`, I expect it's natural to use `async` on `resource.drop`, `subtask.cancel` and `{stream,future}.cancel-{read,write}` by default

alexcrichton: Personally I'd prefer to keep things separate where possible because we already suffer from under-testing major features. I find that lumping things together under one feature umbrella encourages very large changes which are rarely thoroughly tested. With split features I find it encourages smaller changes which are much easier to verify are tested for example.

Put another way I agree with your motivations and I agree with the eventual conclusion, but given our development so far I've found large features to be detrimental to the end quality of the final implementation. One example is https://github.com/bytecodealliance/wasmtime/issues/11611 where it's just too easy to forget about something given the huge matrix of possible interactions with all the async features. We very frequently will discuss changes to a non-existent implementation (historically in the implementation of p3) which results in spec changes but we never write down tests for these changes to ensure the old behavior works plus the new cases.

Basically I'd like to keep these all separate for development concerns even if in the end we end up lumping everything under one umbrella. I also think they should be separate if the end result is unusable without everything implemented because keeping things separate doesn't change the net amount of work done in total but I think produces a higher quality result as implementations are validated/verified in chunks rather than all-at-once.

lukewagner: Making sure things are tested when added is certainly important.  My expectation is that both the stackful-async and `async` built-ins will need to be tested b/c they would be used by the essential green thread use cases motivating cooperative threads.  But if that turns out not to be the case or if we want to roll out cooperative threads itself in chunks, then happy to revisit in that context.

alexcrichton: I guess what I'm saying is that empirically testing has not happend historically. There's a whole matrix of possibilities of just the async bits before this PR and only some are tested. We're still running into missing tests and blind spots in the implementation today. Personally I think this would have greatly benefitted from not frontloading the full matrix of all options for all our use cases and instead splitting things up piecemeal but that's all history now at this point.

What I'd like to do going forward is that we've already got buckets for some features and I don't want to proactively start making those buckets all one big bucket. We'd have to proactively take a bunch of existing feature-gated code and update it all to be under a new feature gate. Today it's easy enough to grep for a single feature gate and see what needs testing when that wants to be "stabilized" but this becomes much more difficult when it's one large feature. At that point everyone asks if it's ready and no one can keep it all in their heads and so we point at a green set of tests and shrug.

Again though I don't dispute that in the long run that we'll probably want all these features implemented. My point is that I think it's valuable for developing a quality implementation that these are kept separate. It's not the same person designing, implementing, testing, and reviewing all in one got, but lots of different people, and in that situation keep things in smaller chunks is easier to keep in your head.

Basically, personally, I don't want to wait to revisit this, I want to change this right now to keep features separate.

lukewagner: Ok, reverted then.

lukewagner:
I realized an interesting corner case (introduced by #548 removing the "suspend-the-whole-instance" behavior of sync imports) that isn't observable in 0.3.0 that becomes observable with either cooperative threads or stackful `async`: if one thread blocks on a synchronous `{stream,future}.{read,write}`, another thread in the same component instance can resume and attempt to `{stream,future}.cancel-{read,write}` it.  I believe the desired behavior is to trap (since the sync read/write isn't expecting cancellation) like we do with other sync non-`cancellable` built-ins, and so that's what [this commit](https://github.com/WebAssembly/component-model/pull/557/commits/0c0e6ea3101f44b7f2941b379b02ee75f683addd) adds.

TartanLlama:
Merged into `wasm-tools`: https://github.com/bytecodealliance/wasm-tools/pull/2298

TartanLlama:
Implemented in Wasmtime here: https://github.com/bytecodealliance/wasmtime/pull/11751
I think this is good to merge!

design/mvp/Concurrency.md
```diff
@@ -36,68 +32,117 @@ summary of the motivation and animated sketch of the design in action.
...
+  * host threads that are scheduled outside the language's own runtime in
+    languages like C, C++, C#, Python, Rust and many more that expose pthreads
+    or other OS threads
+  * promises, futures, streams and channels
+  * callbacks, in languages with no other built-in concurrency mechanisms
+* Provide [fiber]-like stack-switching capabilities via Core WebAssembly
+  import calls in a way that complements, but doesn't depend on, new Core
+  WebAssembly proposals including [stack-switching] and
+  [shared-everything-threads].
+* Allow polyfilling in browsers via JavaScript Promise Integration ([JSPI])
```
dicej, line 53: Now that we have a complete set of cooperative threading intrinsics, it would be great to expand on this, possibly in its own document.  In particular, I'd like to see a rough sketch of how the various `thread.*` intrinsics might be implemented in terms of JSPI.  It wouldn't have to go into a lot of detail -- just enough to demonstrate that the intrinsics can be implemented in the browsers and standalone JS engines we have today (rather than needing to wait for core stack switching).

No need to include that in this PR, but would be worth a follow-up PR.

lukewagner: Yeah, great point; will do.

design/mvp/Binary.md
```diff
@@ -322,13 +322,21 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift
            | 0x21 cancel?:<cancel?> m:<core:memidx>              => (canon waitable-set.poll cancel? (memory m) (core func)) 🔀
            | 0x22                                                => (canon waitable-set.drop (core func)) 🔀
            | 0x23                                                => (canon waitable.join (core func)) 🔀
-           | 0x40 ft:<typeidx>                                   => (canon thread.spawn_ref ft (core func)) 🧵
-           | 0x41 ft:<typeidx> tbl:<core:tableidx>               => (canon thread.spawn_indirect ft tbl (core func)) 🧵
-           | 0x42                                                => (canon thread.available_parallelism (core func)) 🧵
+           | 0x26                                                => (canon thread.index (core func)) 🧵
+           | 0x27 ft:<typeidx> tbl:<core:tableidx>               => (canon thread.new_indirect ft tbl (core func)) 🧵
```
dicej: Nit: What's the reason for sometimes using underscores and sometimes using dashes in these names?  (I realize we were already doing that before this PR, but I only now noticed it.)

dicej: BTW, I just noticed you switched `thread.available_parallelism` to `thread.available-parallelism` in this PR;  maybe we should do the same for the others that still use underscores?

design/mvp/Explainer.md
```diff
@@ -2045,6 +2047,209 @@ value has not already been written.
...
+| Canonical ABI signature    | `[fi:i32 c:i32] -> [i32]`                                     |
+
+The `thread.new_indirect` built-in adds a new thread to the current component
+instance's table, returning the index of the new thread. The function table
+supplied via [`core:tableidx`] is indexed by the `fi` operand and then
+dynamically checked to match the type `FuncT` (in the same manner as
+`call_indirect`). Lastly, the indexed function is called in the new thread
+with `c` as its first and only parameter.
+
+Currently, `FuncT` must be `(func (param i32))` and thus `c` must always an
```
dicej:
```suggestion
Currently, `FuncT` must be `(func (param i32))` and thus `c` must always be an
```

design/mvp/Binary.md
lukewagner: I mostly was just trying to keep `thread.new_indirect` symmetric with `thread.spawn_indirect` as named/linked [here](https://github.com/WebAssembly/shared-everything-threads/blob/main/proposals/shared-everything-threads/Overview.md#thread-management-builtins), but it does sortof stick out and maybe we should switch them all over to hyphens?

lukewagner: Yeah, regularity seems like a good idea, and perhaps the principle here is that built-ins have a different (kebab-cased) naming scheme than core wasm instructions (which are snake-cased).  [Changed](https://github.com/WebAssembly/component-model/pull/557/commits/ed97e7bd7fa91632b82a256325520e14c5d36207)

PR #562, opened 2025-09-10
rename `async` option to `cancellable` for `waitable-set.poll`

dicej:
Looks like this got missed when it was renamed for `waitable-set.wait` and `yield` (and in the prose for `waitable-set.poll` itself).


lukewagner:
Oops, thanks!

PR #569, opened 2025-10-20
Port Wasmtime's `async-closed-stream` test

yoshuawuyts:
Hi! This is a port of Wasmtime's [async-closed-stream.rs](https://github.com/bytecodealliance/wasmtime/blob/bd7b59dad0923097760eb73103582252fcd2f408/crates/test-programs/src/bin/async_closed_stream.rs) test to WAST. The core of this test include bindings to the following WIT:

```wit
interface closed-stream {
  get: func() -> stream<u8>;
}
```

With the core of the test being the following (note that it's never actually invoked):

```rust
impl Guest for Component {
    fn get() -> StreamReader<u8> {
        bindings::wit_stream::new().1
    }
}
fn main() {}
```

This is the first time I'm writing WAST, Wasm Bytecode, and Lisp. The test locally seems to pass, but since it doesn't actually invoke anything all I can tell is that it's syntactically well-formed. I'd like to port more of the Wastime async tests to WAST in the future, so I figured I'd start with the simplest one. That said though: I would really appreciate a review on this, as I'm at best 60% confident this does what it's supposed to. Thanks!

### DIFF
--- a/test/async/closed-stream.wast
+++ b/test/async/closed-stream.wast
@@ -1,25 +1,102 @@
-;; This test contains a single component $A which imports a closed stream
-;; whose instantiation shouldn't trap.
-
+;; This test contains two components $C and $D that test that if the writable side
+;; of a stream is dropped, the other side registers a STREAM DROPPED status
+;; when attempting to read from the stream.
 (component definition $Tester
-  (component $A
+  ;; Creates a stream and keeps a handle to the writable end of it.
+  (component $C
     (core module $Memory (memory (export "mem") 1))
     (core instance $memory (instantiate $Memory))
     (core module $CM
       (import "" "mem" (memory 1))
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Oops, I've been informed I was missing the second half of this test from: https://github.com/bytecodealliance/wasmtime/blob/dcd654469e78ccaa25c08ee63f303d623c108a6c/crates/misc/component-async-tests/tests/scenario/streams.rs#L262

yoshuawuyts:
Okay, test should be all update now. What the code actually did was test that if we attempt to read from a stream whose writer side has been dropped, the reader should return the `Dropped` status when we try to read from it.

The opposite side of this test should probably be written at some point too, but for now I'm going to settle on just porting the existing tests as-is. Thanks!

lukewagner:
Thanks!

PR #577, opened 2025-11-25
add `async` to funcs that need to block in WAST tests

dicej:
This anticipates the forthcoming changes to WIT/CM async and helps us get CI green for https://github.com/bytecodealliance/wasmtime/pull/12043


lukewagner:
These test fixes are now folded into #578, so I think we can close this.

PR #565, opened 2025-09-15
Add section to Async.md discussing readiness/non-blocking I/O

lukewagner:
This PR adds some explanatory prose in a new section in Async.md to discuss zero-length reads/writes and capture some of the discussion in #561 and #441.

PR #578, opened 2025-11-29
Make 'async' part of the function type, not a hint

lukewagner:
This PR moves `async` from being a hint that is textually mangled into function names to an optional effect type that is part of a function type.  WIT is not changed; just how it's encoded as a component import/export definition.  For example,
```wit
foo: async func(s: string) -> string;
```
goes from being encoded (when imported) as:
```wat
(import "[async]foo" (func (param "s" string) (result string)))
```
to
```wat
(import "foo" (func async (param "s" string) (result string)))
```
Along with (and motivating) this change is a runtime behavioral change: if core wasm "blocks" (viz.: calls `waitable-set.{wait,poll}`, `thread.suspend`, or calls something that might block synchronously) *and* the current task has not returned *and* the current task's function type does not contain `async`, there is a trap.  Thus, `async` moves from being a "hint" that can be violated to an invariant that a client (host or component) can depend on.

The motivation for this change is working through the browser/JS embedding (both in a near-term `jco transpile` context but also considering a longer-term native browser/JS embedding).  If the lack of `async` doesn't *guarantee* the lack of blocking, then a browser/JS embedding would *never* be able to call a non-`async` component export in a synchronous context (with important use cases being: an `addEventListener` callback, constructors, getters and setters).  This is certainly the case for `jco transpile` implemented in terms of JSPI (where "blocking" manifests as the JS glue code calling into wasm that returns a `Promise` that cannot be `await`ed in a synchronous JS context).  But even in a native browser implementation, blocking in a synchronous context basically requires spinning a nested event loop which browsers are generally trying to kill off and prevent any new occurrences of.  This change would also be beneficial for other, non-browser embeddings which have the same underlying implementation constraints as browsers.

This change also reduces the cognitive overhead for bindings generators and power users of the hint.  As a hint, since a non-`async` import *might* block, the bindings generator had to ask: "do I, and how do I, expose a power-user option for overriding the hint to handle the blocking case".  (This is especially tricky when considering native binding support, as we'd eventually like for all embeddings of components into language runtimes, where there's not the option of "build flags"). Conversely, if I'm a very careful programmer who wants to achieve maximum concurrency: if `async` is just a hint and callees "might" always block: should I override the hint and when? Do I just determine when to override experimentally or conservatively always use async?  Having a checked effect avoids developers wasting their time with these obscure-but-otherwise-necessary questions.

Although this change would seem to give functions a "color", for the reasons described in this PR in [Concurrency.md#summary](https://github.com/WebAssembly/component-model/blob/01ae148b9afabe2225c858e64959023212a94d59/design/mvp/Concurrency.md?plain=1#L105-L128), `async` at the WIT/Component level does not have the same infective quality as source-language `async`.  This change is also backwards compatible with existing Preview 2 components (which are all non-`async` and have no way to block) and existing Preview 3 WASI worlds (which already mark all the exports as `async` and can thus always block, as you'd expect).

### DIFF
--- a/design/mvp/Concurrency.md
+++ b/design/mvp/Concurrency.md
@@ -841,42 +841,22 @@ Despite the above, the following scenarios do behave deterministically:
 
 ## Interaction with the start function
 
-Since any component-level function with an empty signature can be used as a
-[`start`] function, there's nothing to stop an `async`-lifted function from
-being used as a `start` function. Async start functions are useful when
-executing general-purpose code at initialization time, e.g.:
-* If the top-level scripts of a scripting language are executed by the `start`
-  function, asychrony arises from regular use of the language's concurrency
-  features. For example, in JS, this takes the form of [top-level `await`].
-* If C++ or other OOPLs global object constructors are executed by the `start`
-  function, these can execute general-purpose code which may use concurrent
-  I/O APIs.
+All start functions (both component-level and Core WebAssembly start functions
+called via `core instance` definition) implicitly have the component-level
+function type `func()`, i.e., they are synchronous and take and return no
+arguments. Based on the above description of synchronous functions, this means
+that start functions may not block before returning. However, if a
+component-level start function is lifted using the async ABI, it *may* block
+after calling `task.return`, and may thus serve as a long-running "background
+task" to which work can be dispatched (e.g., via the `setInterval()` or
+`requestIdleCallback()` JavaScript APIs). From the perspective of [structured
+concurrency], these background tasks are new task tree roots (siblings to the
+roots created when component exports are called by the host).
 
-Since component `start` functions are already defined to be executed
-synchronously before the component is considered initialized and ready for its
-exports to be called, the natural thing for `start` to do when calling an
-`async`-lifted function is wait for the callee to reach the ["returned"
-state](#returning). This gives `async` `start` functions a simple way to do
-concurrent initialization and signal completion using the same language
-bindings as regular `async` `export` functions.
-
-However, as explained above, an async task can always continue executing after
-reaching the "returned" state and thus an async task spawned by `start` may
-continue executing even after the component instance is initialized and
-receiving export calls. These post-return `start`-tasks can be used by the
-language toolchain to implement traditional "background tasks" (e.g., the
-`setInterval()` or `requestIdleCallback()` JavaScript APIs). From the
-perspective of [structured concurrency], these background tasks are new task
-tree roots (siblings to the roots created when component exports are
-called by the host). Thus, subtasks and threads spawned by the background task
-will have proper async callstacks as used to define reentrancy and support
-debugging/profiling/tracing.
-
-In future, when [runtime instantiation] is added to the Component Model, the
-component-level function used to create a component instance could be lowered
-with `async` to allow a parent component to instantiate child components
-concurrently, relaxing the fully synchronous model of instantiation supported
-by declarative instantiation and `start` above.
+As a Preview 3 follow-up [TODO](#TODO), component type definitions should be
+extended to allow an `async` effect that declares that component instantiation
+is allowed to [block](#blocking). This would be necessary to implement, e.g.,
+JS [top-level `await`] or I/O in C++ constructors executing during `start`.
 
 
 ## Async ABI
@@ -1280,6 +1260,10 @@ comes after:
   `write` of a stream/future happen from within the same component instance
 * zero-copy forwarding/splicing
 * some way to say "no more elements are coming for a while"
+* add an `async` effect on `component` type definitions allowing a component
+  type to block during instantiation
+* add an `async` effect on `resource` type definitions allowing a resource
+  type to block during its destructor
 * `recursive` function type attribute: allow a function to opt in to
   recursive [reentrance], extending the ABI to link the inner and
   outer activations


design/mvp/CanonicalABI.md
```diff
@@ -3810,6 +3853,10 @@ def unpack_event(mem, thread, ptr, e: EventTuple):
   store(cx, p2, U32Type(), ptr + 4)
   return [event]
 ```
+A non-`async`-typed function export that has not yet returned a value
+unconditionally traps if it transitively attempts to call `wait` (regardless of
```
badeend, line 3842: > A non-`async`-typed function export that has not yet returned a value unconditionally traps if (...)

I see this definition in a couple places. Does this mean that a non-`async`-typed function export _may_ block _after_ it has returned a value?

design/mvp/Concurrency.md
```diff
@@ -374,6 +455,69 @@ allowing control flow to escape into the wild.
...
+all of which are described above or below in more detail:
+* calling an `async`-typed function import using the sync ABI
+* suspending the current thread via the
+  [`thread.suspend`](#thread-built-ins) built-in
+* cooperatively yielding (e.g., during a long-running computation) via the
+  [`thread.yield`](#thread-built-ins) built-in
+* waiting for one of a set of concurrent operations to complete via the
+  [`waitable-set.{wait,poll}`](#waitables-and-waitable-sets) built-ins
+* waiting for a stream or future operation to complete via the
+  [`{stream,future}.{,cancel-}{read,write}`](#streams-and-futures) built-ins
```
badeend, line 479: How about [readiness checks/non-blocking I/O](https://github.com/WebAssembly/component-model/pull/565)? Are those considered "blocking"?

design/mvp/Concurrency.md
```diff
@@ -72,34 +74,58 @@ as `select`, `epoll`, `io_uring`, `kqueue` and Overlapped I/O) making the
...
+functions so that the contained Core WebAssembly code is free to block.
+Implementing a non-`async` function will primarily only arise when a component
+is *virtualizing* the non-`async` *imports* of a `world` (e.g., the getters and
+setters of `wasi:http/types.headers`). In this virtualization scenario (once
+functions are allowed to be [recursive](#TODO)), the Canonical ABI and/or Core
+WebAssembly [stack-switching] proposal will allow a parent component to
+implement a child's non-`async` imports in terms of the parent's `async`
+imports in the same manner as [JSPI]. Thus, overall, `async` in WIT and the
+Component Model does not behave like a "color" in the sense described by the
+popular [What Color Is Your Function?] essay.
```
badeend, line 128: In what sense does it not behave like coloring? Irrespective of how they're lowered/lifted:
Async imports may only be called from within async exports, right?

To me, that sounds exactly like the virality described in "What Color Is Your Function?".

design/mvp/CanonicalABI.md
alexcrichton: My understanding (Luke correct me if I'm wrong), is that theoretically yes. However ABI-wise this isn't possible today because if you're non-`async` then returning a value is only possibly by returning from the function itself. The only wasm code that gets to run after returning is post-return which (I believe) can't block either.

In a hypothetical future with a different ABI, however, such as always using `task.return` to communciate the return value, then this is allowing blocking behavior after returning. The consequence (which is my own personal inference from this) is that the host must run such code on a fiber unconditionally because it doesn't know if the guest will block after calling `task.return`. All the host knows is that the guest won't block before `task.return` is called.

design/mvp/Concurrency.md
alexcrichton: My understanding (Luke correct me if I'm wrong), is "no" so long as the intrinsics are declared with the `async` option. Omitting the `async` option means that they block for the operation, but this won't be what non-blocking I/O does anyway. For nonblocking I/O all the intrinsics used will be declared `async` which means that a "BLOCKED" status code will be returned as opposed to trapping, and that can be handled as per usual.

design/mvp/CanonicalABI.md
lukewagner: Yep!

design/mvp/Concurrency.md
alexcrichton: The way I think about this is that you're correct that functions have colors but it's not as extreme as the essay. You're correct that a sync export cannot call an async import in a blocking fashion, but the reason it's not as extreme as the essay is that it's expected to be conventional that WIT worlds have async exports. In such a situation it doesn't matter if your guest language is doing sync or async things -- it'll "all work out". 

So colored in one sense, not colored in a different sense.

design/mvp/Concurrency.md
lukewagner: There's no special case for 0-length reads writes and so the general rules apply: calling `stream.{read,write}` with `async` is always fine but calling without `async` traps (if called from a non-`async` function that has not yet returned a value).

design/mvp/CanonicalABI.md
dicej: To be clear, this _is_ very much possible today -- no hypothetical ABI additions needed.  A guest is able to async-lift a non-`async`-typed function, call `task.return`, and then block all it wants.

alexcrichton: Oh oops, sorry, yes! So for example a host would have to run an async-lifted function into a non-async function type in a fiber.

design/mvp/Concurrency.md
lukewagner: I tried to capture the reasoning for how this isn't coloring [here](https://github.com/WebAssembly/component-model/blob/01ae148b9afabe2225c858e64959023212a94d59/design/mvp/Concurrency.md?plain=1#L105-L128) in the PR, but the summary is:
* Just because code is called through an `async` export doesn't mean the code has the be rewritten from a synchronous style to asynchronous style (unlike with source-language `async`/`await`), because you can still use the sync ABI for *all* imported/exported functions.
* Practically speaking, most code will be called through an `async` export so it can block all it wants.  Mostly only components *virtualizing* worlds containing sync imports will need to implement sync functions and in this advanced virtualization scenario, you usually need higher-order ("donut") composition (where the parent component *both* implements the imports of the child *and* calls the exports of the child).  Once we allow recursive reentrance (it's currently a [TODO](https://github.com/WebAssembly/component-model/blob/01ae148b9afabe2225c858e64959023212a94d59/design/mvp/Concurrency.md?plain=1#L1285-L1287)) a component A virtualizing the world of a component B could then have a callstack A--calls-->B--calls-->A, and then A could implement B's sync imports by suspending in the inner A call to the outer A call, just like JSPI allows in browsers today where "A" is "native JS code".  Thus sync *will* actually be able to call async, but only if you do this advanced "donut wrapping" composition.

design/mvp/CanonicalABI.md
lukewagner: (and sorry on my part here and below for the racy interleaved replies; didn't see your comments when leaving my own)

lukewagner:
@dicej Yep!  Is there already a dev release with these traps implemented that I can use to write the tests?

pannous:
Some time ago you promised not to go down the road of function coloring, yet here we are. Async is the goto keyword of our generation. The concept of asynchronicity belongs on the caller side, not on the callee side. There are alternatives; investing the time and handling it properly now will prevent decades of headaches and pain.

dicej:
> @dicej Yep! Is there already a dev release with these traps implemented that I can use to write the tests?

If you can build [this PR branch](https://github.com/bytecodealliance/wasmtime/pull/12043) of Wasmtime, you should be set, e.g. `cargo install --git https://github.com/dicej/wasmtime --branch trap-blocking-in-sync-tasks --features component-model-async --locked wasmtime-cli`.  I don't think we'll have a dev release until that PR is merged, but let me know if you have trouble and we'll figure something out.

pannous:
> exports impose little to no requirements on the guest language's style of concurrency

OK sorry for commenting before reading the full PR. No requirements would be fine but what does "little requirements" mean

and if the caller may or may not call the function in a concurrent fashion (as it should be !) then why add async keyword anyways?

design/mvp/Concurrency.md
badeend: So, if I got it right:
At the component interface boundaries there is coloring. But because we can still mix-and-match ABIs, that coloring doesn't (necessarily) make its way into user code. Assuming the entrypoint export is async, of course.

---

Are non-async exports allowed to start/spawn async import calls, as long as they don't block on the newly created task?

lukewagner:
> OK sorry for commenting before reading the full PR. No requirements would be fine but what does "little requirements" mean

If we're just talking about source code compiled to run in a component with `async` exports using the sync ABI, I don't think there are any requirements; `async` just buys you more optionality for what you can do at runtime.  The "little or" in "little or no requirements" in Concurrency.md#summary refers to cases where you need to call a component export in a synchronous context, so you might opt for non-`async` exports.

> and if the caller may or may not call the function in a concurrent fashion (as it should be !) then why add async keyword anyways?

It's really just about capturing in the calling contract (= function type), as understood by both the caller and the callee (unlike lifting/lowering ABI options, which are encapsulated impl details of their respective component) the (runtime checked) fact that the callee will not [block](https://github.com/WebAssembly/component-model/blob/tweak-async/design/mvp/Concurrency.md#blocking).

design/mvp/Concurrency.md
lukewagner: > Are non-async exports allowed to start/spawn async import calls, as long as they don't block on the newly created task?

Yep!  And once they return their value (via `task.return`) they can even `waitable-set.wait` on them to finish.

badeend: Alright, got it. Makes sense to me then!

lukewagner:
Ok, added a nice beefy `test/async/trap-if-block-and-sync.wast`.

lukewagner:
Two additional tweaks to this change:
* Added a clarification to Concurrency.md that `start` functions (component and core) are to be interpreted as sync functions.  (Later we could add an `async` effect on the component-type to allow components to declare that instantiation could block.)
* Re-allowed `waitable-set.poll` to be called form sync functions by changing its semantics to not yield but instead do the more primitive operation of just returning whether any waitable in the set has a pending event, since this appears to be actually quite useful to do in a synchronous context in some virtualization scenarios.

design/mvp/CanonicalABI.md
```diff
@@ -4521,13 +4566,22 @@ other threads in a cooperative setting.
...
   match event_code:
     case EventCode.NONE:
       return [SuspendResult.NOT_CANCELLED]
     case EventCode.TASK_CANCELLED:
       return [SuspendResult.CANCELLED]
 ```
+If a non-`async`-typed function export that has not yet returned a value
+transitively calls `thread.yield`, it returns immediately without blocking
+(instead of trapping, as with other possibly-blocking operations like
+`waitable-set.poll`). This is because, unlike other built-ins, `thread.yield`
```
badeend:
```suggestion
(instead of trapping, as with other possibly-blocking operations like
`waitable-set.wait`). This is because, unlike other built-ins, `thread.yield`
```

design/mvp/Concurrency.md
```diff
@@ -374,6 +455,69 @@ allowing control flow to escape into the wild.
...
+
+At a high level, there are six ways for a call to a component export to block,
+all of which are described above or below in more detail:
+* calling an `async`-typed function import using the sync ABI
+* suspending the current thread via the
+  [`thread.suspend`](#thread-built-ins) built-in
+* cooperatively yielding (e.g., during a long-running computation) via the
+  [`thread.yield`](#thread-built-ins) built-in
+* waiting for one of a set of concurrent operations to complete via the
+  [`waitable-set.{wait,poll}`](#waitables-and-waitable-sets) built-ins
```
badeend:
```suggestion
* waiting for one of a set of concurrent operations to complete via the
  [`waitable-set.{wait}`](#waitables-and-waitable-sets) built-ins
```

design/mvp/Explainer.md
```diff
@@ -1750,7 +1739,7 @@ For details, see [`canon_waitable_set_wait`] in the Canonical ABI explainer.
 
 where `event` is defined as in [`waitable-set.wait`](#-waitable-setwait).
 
-The `waitable-set.poll` built-in [suspends][waiting] the [current thread] in
+The `waitable-set.poll` built-in suspends the [current thread] in
 the "ready" state (like `thread.yield`). Once nondeterministically resumed,
 `waitable-set.poll` will return either an event from one of the waitables in
```
badeend:
```suggestion
The `waitable-set.poll` built-in returns either an event from one of the waitables in
```

lukewagner:
@badeend Oops, yes, good catch.  I grepped "poll" and found a bunch of places that needed updating.

badeend:
WASI's [stdio](https://github.com/WebAssembly/WASI/blob/f367c465d5032e5fe9c6b960ae2c0a81a50da4e5/proposals/cli/wit-0.3.0-draft/stdio.wit) is fully async. There has been [discussion](https://github.com/WebAssembly/WASI/issues/757) about this before.

The nearest counterpart in browsers (`console.log`) never blocks and can be called in getters/setters/constructors/etc despite technically doing I/O.

I think the WASI decision hinged on the ability to always block. With that now gone, do you think the WASI interface needs to be revised after this PR?

lukewagner:
@badeend That's a great catch!  Based on the earlier discussions, I think there's (still) significant value in having stdin/stdout expose streams (for forwarding/splicing purposes and efficient completion-based bulk byte transfer).  But obviously we also want to be able to `printf()` to `console.log` at any point in time.  I think perhaps the way to square this circle is to add synchronous `read`/`write` functions to the `stdin`/`stdout`/`stderr` interfaces, alongside the existing `read-via-stream`/`write-via-stream` functions.  This would be somewhat analogous to how, in C, `stdout` has both a `FILE*` that you can pass to `fwrite()` and a file descriptor that you can pass to `write()`.

Alternatively, we could say "as a part of the behavioral contract of `std{in,out,err}`, the streams returned by `{read,write}-via-stream` never return "blocked" when the `stream.{read,write}` built-ins are called asynchronously (which you are always allowed to call), but this seems somewhat fragile and hazardous, e.g., if you want to virtualize these interfaces.  But if we were in a time pinch to release 0.3.0, this might be a quick fix while we add the synchronous functions mentioned above in a 0.3.1.

badeend:
> this seems somewhat fragile and hazardous

I agree ;)

> perhaps the way to square this circle is to add synchronous read/write functions to the stdin/stdout/stderr interfaces, alongside the existing read-via-stream/write-via-stream functions.

Sounds fine to me. Let's continue in the [pre-existing issue](https://github.com/WebAssembly/WASI/issues/757) on that topic.

PR #584, opened 2025-12-18
Tweak: add 'bool' and 'char' to the temporary allow-list of same-instance future/stream copies

lukewagner:
This is a small follow-up to #580 to align the spec with the current implementation; allowing all the primitive types instead of just the numeric types avoids making an otherwise-unnecessary distinction.


alexcrichton:
Oh I think this is an implementation bug because spec-wise we need to validate the bool/char contents as they're transferred. That means that a simple `memcpy` isn't possible, so we need to update Wasmtime to avoid special-casing them

lukewagner:
Oh right of course, I totally forgot about the validation constraint; thanks.

dicej:
> Oh I think this is an implementation bug because spec-wise we need to validate the bool/char contents as they're transferred. That means that a simple `memcpy` isn't possible, so we need to update Wasmtime to avoid special-casing them

1. You're supposed to be on vacation still
2. Good point.  Are floats okay?  We did away with the NaN canonicalization, right?

lukewagner:
@dicej Yes, floats are allowed to be copied without canonicalization (canonicalization is only nondeterministically *allowed*)

PR #585, opened 2025-12-19
Disallow trailing commas in tuple types

sno2:
Trailing commas are not allowed in any of the analogous generic types or function parameter lists.


tschneidereit:
Is there any reason not to go the other way and allow them in all productions where doing so is feasible? I don't see why the arguments for allowing them that apply to other languages wouldn't apply to wit as well.

sno2:
Yeah, that makes sense. I made an issue to discuss this further: #586

PR #589, opened 2025-12-30
Refine recursion check to handle child import/export forwarding

lukewagner:
The current `trap_if_on_stack` recursion check is meant to trap if a single component instance is called recursively (as this can lead to unpredictable deadlock due to backpressure as well as some other problems enumerated in this PR).  (There's a [TODO](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Concurrency.md#todo) to add a way for components to opt-in to allowing recursive reentrance in a way that avoids both the trap and the deadlock.) However, the current simplistic Python definition misses the case where a single parent component is reentered via its child components (that are re-exported or receive forwarded imports).  The reason is that, when calling into a component instance A, one must also consider the call to be entering any parent of A that is not also a parent of the caller's component.  This PR adds a `parent` field to `ComponentInstance` and then uses it to fix recursion check (which is now called `call_is_recursive`).

A nice side-effect of this change (which actually helped point out the problem) is that it ensures that component-to-component recursion checks can be eliminated at compile time in the vast majority of cases where parent components do not also contain core module instances (which allow recursion via `call_indirect`), leaving only the pre-existing host-to-component recursion checks (which in most cases should be a single "ah, the stack is empty, nothing to do here" branch).

This change also shouldn't break any existing code since it only impacts obscure corner cases that we don't think producer toolchains can even emit atm.

### DIFF
--- a/test/async/trap-on-reenter.wast
+++ b/test/async/trap-on-reenter.wast
@@ -63,3 +63,48 @@
   ))
 )
 (assert_trap (invoke "c") "wasm trap: cannot enter component instance")
+
+;; also, for now, trap on parent-to-child
+(component $Parent
+  (component $Child
+    (core module $CoreChild
+      (func (export "f"))
+    )
+    (core instance $core_child (instantiate $CoreChild))
+    (func (export "f") (canon lift (core func $core_child "f")))
+  )
+  (instance $child (instantiate $Child))
+  (canon lower (func $child "f") (core func $f))
+
+  (core module $CoreOuter
+    (import "" "f" (func $f))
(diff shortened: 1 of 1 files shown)


lukewagner:
I updated the PR with a second commit to make the recursion check more conservative and even easier to optimize (see the new text in CanonicalABI.md as for why).  As the original post says, we're not aware of any producer toolchains that can generate components that hit this case, so the update minimizes runtime complexity until we properly implement recursive reentrance.  PTAL!

dicej (approved):

PR #590, opened 2026-01-05
Add more tests for trap-if-block-and-sync

TartanLlama:
Some threading intrinsics were not tested against for the trap conditions on blocking calls inside sync functions. This PR ensures that calling `thread.yield-to`, `thread.switch-to`, and `thread.resume-later` inside sync calls does not trap.


lukewagner:
D'oh, and thanks!

PR #593, opened 2026-01-06
Clarify {stream,future}.cancel-{read,write}

lukewagner:
I realized that the current Python CABI definition ignores the case where a first `{stream,future}.cancel-{read,write}` blocked and then the component tries cancelling a second time.  I think it should simply trap, so this PR adds a new `CopyState` to catch this case.

(I would have added a test to `test/async/cancel-stream.wast`, but it's currently only possible to hit using host powers (because cancellation never blocks in component-to-component cases).  Once we add [`{stream,future}.read-complete`](https://github.com/WebAssembly/WASI/issues/757#issuecomment-3575955734), we'll be able to cause `{stream,future}.cancel-write` to block from within a pure component WAST, so I added a note to add a test then.)

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -4293,7 +4293,7 @@ unconditionally traps if it transitively attempts to make a synchronous call to
 `cancel-read` or `cancel-write` (regardless of whether the cancellation would
 have completed without blocking). There is also a trap if there is not
 currently an async copy in progress (sync copies do not expect or check for
-cancellation and thus cannot be cancelled and repeatedly cancelling the same
+cancellation and thus cannot be cancelled, and repeatedly cancelling the same
 async copy after the first call blocked is not allowed).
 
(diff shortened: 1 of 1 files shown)


design/mvp/CanonicalABI.md
```diff
@@ -4286,7 +4293,8 @@ unconditionally traps if it transitively attempts to make a synchronous call to
 `cancel-read` or `cancel-write` (regardless of whether the cancellation would
 have completed without blocking). There is also a trap if there is not
 currently an async copy in progress (sync copies do not expect or check for
-cancellation and thus cannot be cancelled).
+cancellation and thus cannot be cancelled and repeatedly cancelling the same
```
dicej: Nit: add a comma to separate two independent clauses.
```suggestion
cancellation and thus cannot be cancelled, and repeatedly cancelling the same
```

dicej (approved):

PR #594, opened 2026-01-07
use `async` type as necessary in `sync-streams.wast`

dicej:
Previously, Wasmtime wasn't catching all cases of blocking-in-sync-contexts, but I'm about to push a PR which fixes that.


lukewagner:
Thanks!

Issue #595, opened 2026-01-11
No license for tests


Maniues:
`test` directory does not have a license file. Could you clarify license for them?

lukewagner:
Oh right, good catch!  We already indicate Apache License 2.0 for the `design/` directory, and I think that's compatible the tests copied from wasm-tools/wasmtime, so maybe we should just upgrade `LICENSE` to apply to the whole repo?

alexcrichton:
Wasmtime is technically Apache 2.0 with the LLVM exception, and if that's a problem it's ok to remove those for now and we can get things sorted out. The wasm-tools tests are triple-licensed and can be Apache 2.0 I believe (they're Apache-2.0/MIT for Rust, and then Apache-2.0 + LLVM for Wasmtime)

lukewagner:
Ah, I see, thanks.  I'm a license novice, so this may be a dumb question, but: since "Apache 2.0 + LLVM Exception" is more permissive than "Apache 2.0" (and we don't need the exceptions since these are WASTs we're copying, not compilers), could we still just use "Apache 2.0"?  If not, would an alternative to deleting the directory be to say "Apache 2.0 everywhere except under `tests/wasmtime`"?

alexcrichton:
for that I'd cc @tschneidereit, as I am also very much a license novice...

tschneidereit:
I'm pretty sure the exception can't be dropped, unfortunately. That would have been a great thing to enable in the exception itself (which I'm pretty sure would've been possible), but that's not the case. As things stand, the license is pretty explicit about needing to be retained exactly.

I should note that I can't find a trustworthy (or any, really) source explicitly saying removal of the exception is not permissible, but given that e.g. the [GPLv3](https://www.gnu.org/licenses/gpl-3.0.en.html) has a whole section dedicated to a mechanism for this kind of exception (section 7, "Additional Terms"), I think it's fairly safe to conclude that such a framework is required in order to make removal of exceptions permissible.

I can ask the board to go to the BA's council to get actual legal guidance on this, but it might be easier to go ahead with a dual license, assuming we don't have too many individual copyright holders to ask for permission to do so.

tschneidereit:
> If not, would an alternative to deleting the directory be to say "Apache 2.0 everywhere except under tests/wasmtime"?

[...]

PR #597, opened 2026-01-15
Add license for tests

lukewagner:
As brought up and discussed in #595.  The default license for this repo is Apache 2.0, but since `test/wasmtime/*` are copied from github.com/bytecodelliance/wasmtime which is Apache 2.0 With LLVM Exceptions, and licenses don't work like subtyping, `LICENSE` special-cases `test/wasmtime/*` and a separate license file is added in that directory.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,2 +1,1 @@
-The tests in test/wasmtime are licensed under Apache 2.0 With LLVM Exceptions.
-All other files are licensed under [Apache 2.0](LICENSE-APACHE).
+Unless otherwise noted in a `LICENSE` in a subdirectory, all files are licensed under [Apache 2.0](LICENSE-APACHE).


LICENSE
```diff
@@ -1,10 +1,2 @@
-Please see the LICENSE file for particular directories for the terms applicable
```
tschneidereit, line 1: I think you could keep these first two lines and not list any exceptional cases specifically. That way if later on additional more specific LICENSE files are added, this file doesn't need to change.

LICENSE
```diff
@@ -1,10 +1,2 @@
...
-
-The relevant directories and licenses are:
-
-design/           - Apache License 2.0
-
-(Other directories will be added when the [module-linking] repository merges.)
-
-[module-linking]: https://github.com/webassembly/module-linking/
+The tests in test/wasmtime are licensed under Apache 2.0 With LLVM Exceptions.
+All other files are licensed under Apache 2.0.
```
tschneidereit:
```suggestion
All other files are licensed under [Apache 2.0](LICENSE-APACHE).
```

tschneidereit (approved):
Two suggestions, but lgtm either way!

LICENSE
lukewagner: Yeah, that's probably the right-er way to do things.  I didn't want to initially b/c that means putting a `LICENSE` file in every `test/` subdir, of which we expect there to be many and growing.  But I just tried a different directory configuration that puts the imported tests in a separate `test-imported` directory in the root, allowing `test/LICENSE` to be Apache 2.0 and `test-imported/{wasm-tools,wasmtime}` to have their own separate `LICENSE`s, which seems altogether better, I think.  PTAL!

lukewagner:
@dicej @alexcrichton ~~heads up on directory reorg, specifically putting the `wasm-tools`/`wasmtime` tests under a new `test-imported` directory (which seems like a good idea anyway to indicate that these are the tests that are periodically copied in/out rather than the source of truth).~~
[EDIT: nevermind, reverted]

LICENSE
tschneidereit: I think I'm probably missing something, because what I tried to suggest was something much simpler, and without the kind of restructuring. I just realize that maybe I should've suggested a slight change to the text in the first two lines, instead of saying it should stay the same:

```
Unless otherwise noted in a `LICENSE` in a subdirectory, all files are licensed under [Apache 2.0](LICENSE-APACHE).
```

And then you add the `LICENSE` file to `test/wasmtime`, and that's it: no explicit listing of directories in this file at all, no other text at all, in fact. But again, I'm probably missing something making this not work.

lukewagner: Ah, I misunderstood, gotcha.  Ok, reverted and made that change.  PTAL (last time?)

tschneidereit: LGTM! Sorry for sending you on a wild goose chase before

lukewagner: No worries, and thanks for the help!

PR #604, opened 2026-02-02
Remove tests related to `refines`

### LINKED ISSUE #603: Question: Is the variant case binary format spec missing the refines field?
### Clarification needed: binary encoding of `variant case (refines n)`

While reading the component-model binary specification, I noticed a mismatch between
the grammar in https://github.com/WebAssembly/component-model/blob/main/design/mvp/Binary.md and existing test cases.

The binary grammar currently defines:

`case ::= l:<label'> t?:<valtype>? 0x00`

However, some tests use refines on variant cases. For example, from
[test/wasm-tools/definedtypes.wast](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/definedtypes.wast):

```
(type (variant
  (case "x")
  (case "y" string (refines 0))))
```

When encoded using wasm-tools, the second case appears to include an additional
immediate after the valtype, consistent with a <u32> refines index rather than
a fixed trailing 0x00.

Questions for clarification:

- Is refines intended to be encoded as an optional <u32> in the binary format?
- If so, should the case production in Binary.md be updated to reflect this?

Clarifying the intended binary encoding would help downstream implementations avoid
guessing the boundary between decoding and validation.

alexcrichton:
This was removed some time ago but wasm-tools forgot to catch up.

Closes #603

### DIFF
--- a/test/wasm-tools/definedtypes.wast
+++ b/test/wasm-tools/definedtypes.wast
@@ -22,8 +22,8 @@
 
   (type $A15a (variant (case "x")))
   (type $A15b (variant (case "x" $A1)))
-  (type $A15c (variant (case $x "x") (case $y "y" string (refines $x)) (case "z" string (refines $y))))
-  (type $A15d (variant (case "x") (case "y" string (refines 0)) (case "z" string (refines 1))))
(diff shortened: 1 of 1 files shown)


lukewagner:
Thanks!

PR #613, opened 2026-02-25
Add `(implements "I")` to plainname imports to allow multiple imports

### LINKED ISSUE #287: WIT: world `extern-type` rule is too restrictive
Currently the `extern-type` rule in the WIT grammar is:

```ebnf
extern-type ::= func-type ';' | 'interface' '{' interface-items* '}'
```

This means that both the `import-item` and `export-item` rules in a world may only have named function or inlined interface items.

For example:

```wit
world foo {
    import a: func();
    export b: interface {
        bar: func();
    }
}
```

However, it may be desirable to have a single interface definition that is imported or exported under different names in a world or to simply import or export a "known" interface with a different kebab name.

The former is currently not possible in the grammar, but it is possible with cut-and-paste:

```wit
world foo {
    import a: interface {
        bar: func();
    }

    import b: interface {
        bar: func();
    }
}
```

This obviously duplicates the type information on every inline interface definition.

What I propose with this issue is to add `use-path` as a case to the `extern-type` rule.

This would allow for the following:

```wit
interface bar {
    bar: func();
}

world foo {
   import a: bar;
   import b: bar;
}
```

and also for:

```wit
world foo {
   import my-http-handler: wasi:http/incoming-handler;
}
```

Note: there may be ambiguity introduced by extending the `extern-type` rule in the current `wit-parser` implementation as right now it allows lexing `foo: bar` as a package name made up of three tokens (`<id>`, `':'`, `<id>`); a hypothetical `import foo: bar:baz/qux` in a world might be difficult for it to parse as it will likely see that as an import of a package path with nested namespaces. The `wac` parser lexes package paths as individual tokens and therefore prohibits whitespace in the path, so that would lex as `<import-keyword>`, `<id>`, `:`, `<package-path>`.

ricochet:
Add `(implements "I")` annotation to be added to plainname imports which enables multiple imports of the same interface

Extend the WIT `extern-type` grammar to allow `use-path` as a third case,
enabling `import id: use-path` and `export id: use-path` to create
plain-named imports/exports whose instance type matches a named interface.

This allows importing the same interface multiple times under different
plain names (e.g., `import primary: store; import secondary: store;`),
encoded using the `(implements "...")` annotation pattern.

Fixes #287


design/mvp/Binary.md
```diff
@@ -408,11 +408,14 @@ Notes:
...
-  export, respectively, in the same scope (component, component type or
-  instance type).
+* Validation requires that `[constructor]`, `[method]` and `[static]` annotated
+  `plainname`s only occur on `func` imports or exports and that the first label
+  of a `[constructor]`, `[method]` or `[static]` matches the `plainname` of a
+  preceding `resource` import or export, respectively, in the same scope
+  (component, component type or instance type).
+* Validation requires that `[implements=<I>]` annotated `plainname`s only
+  occur on `instance` imports or exports and that `I` is a valid
+  `interfacename`.
```
lukewagner: I think the grammatical rules (the EBNF in Explainer.md) force `I` to be a valid `interfacename` without validation having to say so here.

Requiring `[implements=<I>]L` to only annotate `instance`-typed imports/exports makes sense, but if we do it here, then I think we should also require it for plain `interfacename`-named imports/exports too.  This restriction actually makes sense and shouldn't break anyone b/c it matches what you can do in WIT today (thereby removing an expressivity gap from #614), so I'm in favor.

design/mvp/Explainer.md
```diff
@@ -2683,6 +2684,23 @@ annotations trigger additional type-validation rules (listed in
...
+(component
+  (import "[implements=<wasi:keyvalue/store>]primary" (instance ...))
+  (import "[implements=<wasi:keyvalue/store>]secondary" (instance ...))
+)
+```
+
+Here, both imports implement `wasi:keyvalue/store` but have distinct plain
+names `primary` and `secondary`. Bindings generators can use the
+`[implements=<I>]` annotation to know which interface the instance implements,
+enabling them to generate the same typed bindings for both imports.
```
lukewagner: The resource types (e.g. `bucket`) defined directly in `wasi:keyvalue/store` will be treated as different, so "generate the same typed bindings for both imports" isn't quite right.  (And this makes sense: if we want to allow two totally different implementations of these two keyvalue stores, each implementation should be able to have its own implementation of `bucket`.)  However I suppose bindings generators could/should share types for *value* types (like the `key-response` record), since they're not tied to the implementation; they just need predictable names.  So maybe "enabling them to share value type bindings"?

Second, it might also be good to give some examples of how the `interfacename` helps hosts/clients of a component.

design/mvp/Explainer.md
```diff
@@ -2819,13 +2837,19 @@ Values]) are **strongly-unique**:
 * If one name is `l` and the other name is `[*]l.l` (for the same
   `label` `l` and any annotation `*` with a dotted `l.l` name), they *are
   not* strongly-unique.
+* An `[implements=<I>]L` name and a bare `I` `interfacename` are always
+  strongly-unique (they are different name kinds: `plainname` vs
+  `interfacename`).
```
lukewagner: I might totally be forgetting a case, but I *think* we don't need this bullet at all; the "Otherwise" clause gives us the behavior we want.

design/mvp/Explainer.md
```diff
@@ -2819,13 +2837,19 @@ Values]) are **strongly-unique**:
...
   * The names are strongly-unique if the resulting strings are unequal.
 
 Thus, the following names are strongly-unique:
 * `foo`, `foo-bar`, `[constructor]foo`, `[method]foo.bar`, `[method]foo.baz`
+* `[implements=<a:b/c>]one`, `[implements=<a:b/c>]two` (different labels
+  after stripping annotation)
+* `[implements=<a:b/c>]foo` and `a:b/c` (different name kinds)
 
 but attempting to add *any* of the following names would be a validation error:
 * `foo`, `foo-BAR`, `[constructor]foo-BAR`, `[method]foo.foo`, `[method]foo.BAR`
```
```suggestion
Thus, the following set of names are strongly-unique and can thus all be imports (or exports) of the same component (or component type or instance type):
* `foo`, `foo-bar`, `[constructor]foo`, `[method]foo.bar`, `[method]foo.baz`, `foo:bar/baz`, `[implements=<foo:bar/baz>]bar`, `[implements=<foo:bar/baz>]quux`

but attempting to add *any* of the following names would be a validation error:
* `foo`, `foo-BAR`, `[constructor]foo-BAR`, `[method]foo.foo`, `[method]foo.BAR`, `[implements=<a:b/c>]foo`, `foo:bar/baz`, `bar`, `[implements=<x:y/z>]bar`
```

design/mvp/WIT.md
```diff
@@ -1381,9 +1406,32 @@ export-item ::= 'export' id ':' extern-type
...
+```wit
+package local:demo;
+
+interface store {
+    get: func(key: string) -> option<string>;
+}
+
+world my-world {
+    import primary: store;
+    import secondary: store;
```
lukewagner: Since you're using `wasi:http/incoming-handler` for the export, maybe use fully-qualified `wasi:keyvalue/store` for the imports (to avoid making a distinction that is beside the main point)?

design/mvp/WIT.md
```diff
@@ -1381,9 +1406,32 @@ export-item ::= 'export' id ':' extern-type
...
+    export my-handler: wasi:http/incoming-handler;
+}
+```
+
+Here, `primary` and `secondary` are two distinct imports that both have the
+instance type of the `store` interface. The plain name of the import is the
+`id` before the colon (e.g., `primary`), not the interface name. This
+contrasts with `import store;` (without the `id :` prefix), which would create
+a single import using the full interface name `local:demo/store`.
```
lukewagner: Maybe also speak to the `export` which has the name `my-handler` and how this is different than `export wasi:http/incoming-handler` (and perhaps we could switch to `wasi:http/handler` in anticipation of p3 :).

design/mvp/WIT.md
```diff
@@ -2061,6 +2109,53 @@ This duplication is useful in the case of cross-package references or split
 packages, allowing a compiled `world` definition to be fully self-contained and
 able to be used to compile a component without additional type information.
 
+When a world imports or exports a named interface with a custom plain name
+(using the `id: use-path` syntax), the encoding uses the `[implements=<I>]`
+annotation defined in [Binary.md](Binary.md) and
+[Explainer.md](Explainer.md#import-and-export-definitions) to indicate which
```
```suggestion
annotation defined in [Explainer.md](Explainer.md#import-and-export-definitions) to indicate which
```

lukewagner: I'd remove mention of `Binary.md` b/c it doesn't add anything beyond the text explainer for our purposes here.

lukewagner (commented):
Awesome, thanks for writing this up!  The PR looks great; just minor nits.  I think what'll be important for this change is seeing how it "plays out" in the producer and consumer tooling, so I think we'll want to keep this PR open until we get some experience from tool implementations, but so far so good from my perspective.

test/wasm-tools/implements.wast
```diff
@@ -0,0 +1,78 @@
+;; RUN: wast --assert default --snapshot tests/snapshots %
```
lukewagner:
```suggestion
```
Since test is new and not wasm-tools-specific, could you move it to a new directory `test/validation/implements.wast`?

test/wasm-tools/implements.wast
```diff
@@ -0,0 +1,78 @@
...
+    (import "[implements=<a:b/c>]one" (instance
+      (export "get" (func (param "key" string) (result (option string))))
+    ))
+    (import "[implements=<a:b/c>]two" (instance
+      (export "get" (func (param "key" string) (result (option string))))
+    ))
+  ))
+)
+
+;; Invalid: [implements=<...>] on func (must be instance)
```
lukewagner: Could you also add a failing test for attempting to export value type and a resource type?  And second: could you also add copies of these tests for `interfacename`s (given that recent addition to the validation rules)?

lukewagner (approved):
Thanks!

lukewagner:
Oh, I forgot: we probably should have an emoji-gate for this associated with the new rules in the grammar in Explainer.md and the descriptions in WIT.md.

design/mvp/WIT.md
```diff
@@ -366,6 +366,31 @@ world union-my-world-b {
...
+interface store {
+    get: func(key: string) -> option<string>;
+}
+
+world base {
+    import cache: store;
+}
+
+world extended {
+    include base with { cache as my-cache }
```
lukewagner, line 387: Thinking more about this example/rule in the context of duplicate-name-conflicts: could you expand this example to have two `include`s that contain a conflict for the plain-name `cache` and explicitly mention that this would be an error without the `with` (b/c we can't de-duplicate plain names like we can with pure interface names).

ricochet:
> Oh, I forgot: we probably should have an emoji-gate for this associated with the new rules in the grammar in Explainer.md and the descriptions in WIT.md.

Since this is about adding labels, let's use 🏷️  aka the `:label:` emoji

test/wasm-tools/implements.wast
ricochet: We now have:
                                                                                
  [implements] tests:                                                           
  - func import                                  
  - component import
  - value type import
  - resource type import
  - value type export 
  - resource type export

  interfacename tests:
  - func import
  - component import
  - value type import
  - resource type import
  - func export
  - value type export
  - resource type export

Found while iterating in wasm-tools that we'll also want: 
- Empty label — `[implements=<a:b/c>]` with nothing after `]` must be rejected                                   
- Empty interface name — `[implements=<>]name` with nothing between `<>` must be rejected

design/mvp/WIT.md
```diff
@@ -2061,6 +2131,52 @@ This duplication is useful in the case of cross-package references or split
...
+🏷️ When a world imports or exports a named interface with a custom plain name
+(using the `id: use-path` syntax), the encoding uses the `[implements=<I>]`
+annotation defined in [Explainer.md](Explainer.md#import-and-export-definitions) to indicate which
+interface the instance implements. For example, the following WIT:
+
+```wit
+package local:demo;
+
+interface store {
+    get: func(key: string) -> option<string>;
```
lukewagner: I just realized that there's a pretty nuanced technical design point (that iiuc the current impl is getting getting right atm) that this example should explicate via the generated WAT:

Could you expand this example to include a `bucket` resource type (returned by an `open` function, like in the current wasi-keyvalue draft) inside the existing `store` interface.  That should end up with two resource types being imported in the WAT.

Next, after the WAT, could you add a second variation of this example which factors out the `bucket` type into a `types` interface (like we do in wasi-http) that is `use`d by the `store` interface.  And this should result in *one* resource type import that is aliased into the two `store` imports.

design/mvp/WIT.md
```diff
@@ -1381,9 +1430,30 @@ export-item ::= 'export' id ':' extern-type
...
+
+Here, `primary` and `secondary` are two distinct imports that both have the
+instance type of the `wasi:keyvalue/store` interface. The plain name of the
+import is the `id` before the colon (e.g., `primary`), not the interface name.
+This contrasts with `import wasi:keyvalue/store;` (without the `id :` prefix),
+which would create a single import using the full interface name
+`wasi:keyvalue/store`. Similarly, the export `my-handler` has the instance type
+of `wasi:http/handler` but uses the plain name `my-handler` instead of the full
+interface name, which is useful when a component wants to export the same
+interface multiple times or simply use a more descriptive name.
```
lukewagner, line 1460: It's probably a good idea to briefly address the ambiguity hazard that Peter mentions at the end of [this comment](https://github.com/WebAssembly/component-model/issues/287#issue-2049710094).

Specifically, as Peter suggested, `a:b` gets lexed into a single `id` token (not 3 tokens `<id='a'>`, `colon`, `<id='b'>`) and similarly `a:b/c` gets lexed as a single `id` token (not 5 tokens).  Thus, if you write:
```wit
world w {
  import a:b;  // error: can't import a package
  import a: b;  // ok, assuming `b` names an interface in scope
}
```
and this makes us future-compatible with the nested packages feature (🪺) where `import a:b:c/d` would be unambiguously interpreted as an import of an interface named `a:b:c/d`.

design/mvp/Explainer.md
```diff
@@ -2683,6 +2685,31 @@ annotations trigger additional type-validation rules (listed in
...
 
+🏷️ When an instance import or export is annotated with `[implements=<I>]L`, it
+indicates that the instance implements interface `I` but is given the plain
+name `L`. This enables a component to import or export the same interface
+multiple times with different plain names. For example:
+
+```wat
+(component
+  (import "[implements=<wasi:keyvalue/store>]primary" (instance ...))
+  (import "[implements=<wasi:keyvalue/store>]secondary" (instance ...))
```
lukewagner: Thinking about this a bit more in the context of [`@external-id`](https://docs.google.com/presentation/d/1LjqnvH93_y0wpHu3YO9kVHWNFvIds9ooLMoENqm8VOg/edit?slide=id.g3cc415a6a25_0_559#slide=id.g3cc415a6a25_0_559), in which one could write:
```wit
@external-id("a@b:c")
import my-store: wasi:keyvalue/store;
```
when the syntax you need for some external identifier doesn't fit in a `plainname` and it'd be a pain to stuff the external identifier in some custom section or external config file... probably the `@external-id` should *not* go in the name-string; it should go in a separate field of the `import`, like the `versionsuffix`, because you don't need it for name-matching, uniqueness, and if the import name is used in logs, the `@external-id` would probably be unwanted noise.  But then I realized that the same rationale applies to the `implements=<I>`: since we're already excluding it from name uniqueness, it's really just an annotation, like `@external-id`.

Thus, for the above WIT, the WAT would be:
```wat
(import "my-store" (implements "wasi:keyvalue/store") (external-id "a@b:c") (instance ...types...))
```
WDYT?

guest271314: You could probably shorten that `@external-id` to something like `(import "console_log" as console-log ...)`, e.g., `import * as mod` , `export { default as mod }` in JavaScript. Further, in conjuction, create the program that analyzes WAT, etc. and modifies non-kebab names for "kebab" names.

design/mvp/WIT.md
alexcrichton: Done now, if you'd like to review

design/mvp/Explainer.md
alexcrichton: I've now updated the PR with this suggestion, including binary format fixes and such.

alexcrichton:
Ok I've gone through this PR and updated everything to the best of my ability. This update notably takes into account Luke's [previous comment](https://github.com/WebAssembly/component-model/pull/613#discussion_r2996644625) which overhauls how this is encoded in the binary format. I've additionally updated the sibling [wasm-tools PR](https://github.com/bytecodealliance/wasm-tools/pull/2453) through which I've validated and verified some of the wat and tests here. I've additionally implemented all the various bits of wasm-tools such that fuzzing is now clean and such.

My hope/expectation is that integrating this into bindings generators will be pretty easy, assuming the [`generate_nominal_type_ids` method](https://docs.rs/wit-parser/latest/wit_parser/struct.Resolve.html#method.generate_nominal_type_ids) is used. One of the major features that this exposes, which never came up prior, is importing the same interface multiple times and exporting multiple times. I realize that sounds obvious but for any bindings generator which has been using `TypeId` or `InterfaceId` as keys in maps this can wreak havoc without the "nominal" form of a graph. Effectively this, in my opinion at least, really puts a nail in the coffin of bindings generators taking a WIT AST as-is and generating bindings from it. Essentially it's required to post-process into a format more akin to the actual wasm representation (e.g. interfaces are duplicated, as well as types defined in those interfaces). This is all the purpose of `generate_nominal_type_ids`, which notably got an update for this feature as well.

design/mvp/Binary.md
```diff
@@ -391,13 +391,17 @@ in the explainer.)
...
 export         ::= en:<exportname'> si:<sortidx> ed?:<externdesc>?      => (export en si ed?)
-importname'    ::= 0x00 len:<u32> in:<importname>                       => in     (if len = |in|)
-                 | 0x01 len:<u32> in:<importname>                       => in     (if len = |in|)
-                 | 0x02 len:<u32> in:<importname> vs:<versionsuffix>    => in vs  (if len = |in|) 🔗
-exportname'    ::= 0x00 len:<u32> in:<exportname>                       => in     (if len = |in|)
-                 | 0x01 len:<u32> in:<exportname>                       => in     (if len = |in|)
-                 | 0x02 len:<u32> in:<exportname> vs:<versionsuffix>    => in vs  (if len = |in|) 🔗
+importname'    ::= 0x00 len:<u32> in:<importname>                       => in      (if len = |in|)
+                 | 0x01 len:<u32> in:<importname>                       => in      (if len = |in|)
+                 | 0x02 len:<u32> in:<importname> vs:<versionsuffix>    => in vs   (if len = |in|) 🔗
```
lukewagner: Since I don't think anyone has implemented/shipped `versionsuffix`, do you suppose we could delete the previous (now obsolete) binary opcodes for it?  (If it's a pain to have `0x03` change meaning, maybe we could just leave `0x03` as a hole for now until the 1.0 binary format reshuffle.)

lukewagner (approved):
Looks great, thanks!  One minor suggestion:

design/mvp/Binary.md
alexcrichton: Seems reasonable to me, updated 👍

lukewagner:
Ok, this looks good, so I'll merge later this week if there's no additional comments and we can iterate in follow-up PRs if need be.

PR #635, opened 2026-04-09
Add tags to core:sort

bvisness:
This was seemingly an oversight, as I can't think of any reason why it should be impossible to import and export tags during linking.


lukewagner (approved):
Makes sense, thanks!

PR #638, opened 2026-04-16
Make resource `dtor` type explicit. 

michael-weigelt:
Follow-up to https://github.com/WebAssembly/component-model/pull/624. 

In a 64bit component, the `resource rep` argument can be `i32` or `i64`, and the destructor's signature must match it.


lukewagner (approved):
Makes sense, thanks!

PR #639, opened 2026-04-17
Add WAVE format to this repository

alexcrichton:
This commit copies over the specification of the WAVE from the [wasm-tools repository][repo] over to this repository. The intent of this commit is to make the concept of a stringification of component model values a bit more standard that "just something that lives in wasm-tools". In copying over the document I've added a **Note** section explaining the current status to document some limitations and also be clear that there's not necessarily uniform consensus yet. Nevertheless this feels like a better home for this document than the wasm-tools repository.

[repo]: https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasm-wave/README.md

### DIFF
--- a/design/mvp/WAVE.md
+++ b/design/mvp/WAVE.md
@@ -1,0 +1,271 @@
+# WAVE: Web Assembly Value Encoding
+
+WAVE is a human-oriented text encoding of WebAssembly Component Model values. It is designed to be consistent with the
+[WIT IDL format](./WIT.md).
+
+|Type|Example Values
+|---|---
+|Bools|`true`, `false`
+|Integers|`123`, `-9`
+|Floats|`3.14`, `6.022e+23`, `nan`, `-inf`
+|Chars|`'x'`, `'☃︎'`, `'\''`, `'\u{0}'`
+|Strings|`"abc\t123"`
+|Tuples|`("abc", 123)`
+|Lists|`[1, 2, 3]`
+|Records|`{field-a: 1, field-b: "two"}`
+|Variants|`days(30)`, `forever`
+|Enums|`south`, `west`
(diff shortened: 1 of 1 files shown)


lukewagner:
Overall, it seems like a good idea to me to include a human-readable format for Component Model values in the spec if for no other reason that, once we have [value imports/exports 🪙](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#gated-features), we'll want to use this same syntax for value constants, e.g.:
```wit
interface i {
  x: list<string> = [ "one", "two" ];
}
```

One high-level comment is that, in the context of the Component Model standard, "WebAssembly Value" seems a bit too general as a name and sounds like it should include (or *only* include) Core WebAssembly values.  Given the goal of eventual inclusion in WIT, I'd suggest calling it something descriptive like "WIT Values".

A second high-level comment is that I'd feel more comfortable merging this if it contained an EBNF grammar in the same style as WIT.md.  Also, in the same way that the "semantics" of WIT are defined (via example) in WIT.md by translation to Component Model types, it'd be great if there was a section at the end that described how to "encode" a WIT value as a component model `value` 🪙 definition (which is I think what use cases like wrpc assume).  Both of these would help make this new value syntax a natural extension of what we're already doing with WAT/WIT instead of a third new thing.

alexcrichton:
That makes sense to me, yeah. I don't personally have the time right now to invest in this, but I'm happy to leave it here in case someone wants to pick it up.

lann:
> an EBNF grammar

I don't recall how closely I followed the conventions in this repo but it exists! https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasm-wave/wave_ebnf.md

PR #640, opened 2026-04-20
Define bounded lists

### LINKED ISSUE #385: Bounded lists and strings
This extends #181 with the ability to embed bounded strings or lists directly within other structures while maintaining the variable length property.

Some environments require avoiding all memory allocations at runtime. This happens typically in either small embedded environments or safety related products.

To achieve this in a call returning a string or list/vector you typically combine a fixed size buffer with a length indicator into a containing data structure, e.g. see [heapless::String](https://docs.rs/heapless/latest/heapless/struct.String.html) or [boost::static_string](https://www.boost.org/doc/libs/1_85_0/libs/static_string/doc/html/index.html).

Thus I propose to optionally extend the fixed size buffers of #181 as implemented in #384 with a preceding length indicator. This length indicator could be of type u8 for strings/vectors below 256 elements, u16 below 64k and u32 otherwise. Unused elements in the buffer should be considered uninitialized (`MaybeUninit` in Rust).

This addresses the problem described in my recent comment in #383 .

I propose the WIT syntax `string<..32>` or `list<u32, ..16>`, the binary format would need a new type for bounded strings, lists could use a negative len specifier in 0x67 (see #384 ) or a different defvaltype.

Felix-El:
Add bounded lists (`list<T, ..N>`)

Closes #385.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -201,6 +201,7 @@ defvaltype    ::= pvt:<primvaltype>                       => pvt
                 | 0x6f t*:vec(<valtype>)                  => (tuple t+)    (if |t*| > 0)
                 | 0x6e l*:vec(<label'>)                   => (flags l+)    (if 0 < |l*| <= 32)
                 | 0x6d l*:vec(<label'>)                   => (enum l+)     (if |l*| > 0)
+                | 0x6c t:<valtype> maxlen:<u32>           => (list t (le maxlen)) (if maxlen > 0) 🔧
                 | 0x6b t:<valtype>                        => (option t)
                 | 0x6a t?:<valtype>? u?:<valtype>?        => (result t? (error u)?)
                 | 0x69 i:<typeidx>                        => (own i)
@@ -282,8 +283,9 @@ Notes:
   to match the preceding `sort`.
 * (The `0x00` immediate of `case` may be reinterpreted in the future as the
   `none` case of an optional immediate.)
-* 🔧 for fixed-sized lists the length of the list must be larger than 0 to pass
-  validation.
+* 🔧 for fixed-sized lists (`0x67`) the length of the list must be larger than
+  0 to pass validation; for bounded lists (`0x6c`) the maximum length must be
+  larger than 0 to pass validation.
 
 
 ## Canonical Definitions
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -87,12 +87,12 @@ sort                ::= 0x00 cs:<core:sort>                                => co
                       | 0x03                                               => type
                       | 0x04                                               => component
                       | 0x05                                               => instance
-inlineexport        ::= na:<nameattributes> si:<sortidx>                   => (export na si)
+inlineexport        ::= n:<exportname'> si:<sortidx>                       => (export n si)
 ```
 Notes:
 * Reused Core binary rules: [`core:name`], (variable-length encoded) [`core:u32`]
 * The `core:sort` values are chosen to match the discriminant opcodes of
-  [`core:externtype`].
+  [`core:importdesc`].
 * `type` is added to `core:sort` in anticipation of the [type-imports] proposal. Until that
   proposal, core modules won't be able to actually import or export types, however, the
   `type` sort is allowed as part of outer aliases (below).
@@ -101,8 +101,8 @@ Notes:
   for aliases (below).
 * Validation of `core:instantiatearg` initially only allows the `instance`
   sort, but would be extended to accept other sorts as core wasm is extended.
-* Validation of `instantiate` requires each `<externname>` in `c` to match a
-  `name` in a `with` argument (using plain string equality) and for the types to
+* Validation of `instantiate` requires each `<importname>` in `c` to match a
+  `name` in a `with` argument (compared as strings) and for the types to
   match.
 * When validating `instantiate`, after each individual type-import is supplied
   via `with`, the actual type supplied is immediately substituted for all uses
@@ -116,23 +116,24 @@ Notes:
 
 (See [Alias Definitions](Explainer.md#alias-definitions) in the explainer.)
 ```ebnf
-alias ::= s:<sort> 0x00 i:<instanceidx> n:<name>           => (alias export i n (s))
-        | s:<sort> 0x01 i:<core:instanceidx> n:<core:name> => (alias core export i n (s))
-        | s:<sort> 0x02 ct:<u32> idx:<u32>                 => (alias outer ct idx (s))  (if s in outeraliassort)
+alias       ::= s:<sort> t:<aliastarget>                => (alias t (s))
+aliastarget ::= 0x00 i:<instanceidx> n:<name>           => export i n
+              | 0x01 i:<core:instanceidx> n:<core:name> => core export i n
+              | 0x02 ct:<u32> idx:<u32>                 => outer ct idx
 ```
 Notes:
 * Reused Core binary rules: (variable-length encoded) [`core:u32`]
-* For (`core`) `export` aliases, `i` is validated to refer to a (core) instance
-  in the (core) instance index space that exports `n` with the specified `sort`.
-* For `outer` aliases:
-  * `ct` is validated to be *less than or equal to* the number of enclosing
-    "scopes" (where a "scope" is one of: a `component` definition, a `component`
-    type, an `instance` type)
-  * `i` is validated to be a valid index in the `s` index space of the target
-    scope (counting outward, starting with `0` referring to the current scope)
-  * if `ct` is greater than 0 and crosses a component (as opposed to type)
-    boundary and `s` is `type`, the target type may not transitively refer to a
-    resource type.
+* For `export` aliases, `i` is validated to refer to an instance in the
+  instance index space that exports `n` with the specified `sort`.
+* For `outer` aliases, `ct` is validated to be *less or equal than* the number
+  of enclosing `component`s and `type`s and `i` is validated to be a valid index
+  in the `sort` index space of the targeted `component`/`type` (counting
+  outward, starting with `0` referring to the current `component`/`type`).
+* For `outer` aliases, validation restricts the `sort` to one of `type`,
+  `module` or `component`.
+* For `outer` aliases that reach across a `component` boundary (as opposed to
+  a `type` boundary), validation additionally requires that any outer-aliased
+  `type` does not transitively refer to a `resource` type.
 
 
 ## Type Definitions
@@ -148,12 +149,13 @@ core:moduledecl  ::= 0x00 i:<core:import>                                 => i
                    | 0x01 t:<core:type>                                   => t
                    | 0x02 a:<core:alias>                                  => a
                    | 0x03 e:<core:exportdecl>                             => e
-core:alias       ::= 0x10 0x01 ct:<u32> idx:<u32>                         => (alias outer ct idx (type))
+core:alias       ::= s:<core:sort> t:<core:aliastarget>                   => (alias t (s))
+core:aliastarget ::= 0x01 ct:<u32> idx:<u32>                              => outer ct idx
 core:importdecl  ::= i:<core:import>                                      => i
-core:exportdecl  ::= n:<core:name> t:<core:externtype>                    => (export n t)
+core:exportdecl  ::= n:<core:name> d:<core:importdesc>                    => (export n d)
 ```
 Notes:
-* Reused Core binary rules: [`core:import`], [`core:externtype`],
+* Reused Core binary rules: [`core:import`], [`core:importdesc`],
   [`core:rectype`]
 * Unfortunately, the `core:deftype` rule results in an encoding ambiguity: the
   `0x50` opcode is used by both `core:moduletype` and a non-final
@@ -169,9 +171,8 @@ Notes:
   core type index space will not contain any core module types.
 * As described in the explainer, each module type is validated with an
   initially-empty type index space.
-* In `core:alias`, the first `0x10` is the opcode for `type` in `core:sort` and
-  `0x01` is an opcode to distinguish `outer` aliases from potential future
-  `export` aliases.
+* `alias` declarators currently only allow `outer` `type` aliases but
+  would add `export` aliases when core wasm adds type exports.
 
 ```ebnf
 type          ::= dt:<deftype>                            => (type dt)
@@ -200,8 +201,8 @@ defvaltype    ::= pvt:<primvaltype>                       => pvt
                 | 0x70 t:<valtype>                        => (list t)
                 | 0x67 t:<valtype> len:<u32>              => (list t len)  (if len > 0) 🔧
                 | 0x6f t*:vec(<valtype>)                  => (tuple t+)    (if |t*| > 0)
-                | 0x6e l*:vec(<labellit>)                 => (flags l+)    (if 0 < |l*| <= 32)
-                | 0x6d l*:vec(<labellit>)                 => (enum l+)     (if |l*| > 0)
+                | 0x6e l*:vec(<label'>)                   => (flags l+)    (if 0 < |l*| <= 32)
+                | 0x6d l*:vec(<label'>)                   => (enum l+)     (if |l*| > 0)
                 | 0x6b t:<valtype>                        => (option t)
                 | 0x6a t?:<valtype>? u?:<valtype>?        => (result t? (error u)?)
                 | 0x69 i:<typeidx>                        => (own i)
@@ -209,9 +210,9 @@ defvaltype    ::= pvt:<primvaltype>                       => pvt
                 | 0x66 t?:<valtype>?                      => (stream t?) 🔀
                 | 0x65 t?:<valtype>?                      => (future t?) 🔀
                 | 0x63 k:<valtype> v:<valtype>            => (map k v) (if k is in <keytype>) 🗺️
-labelvaltype  ::= l:<labellit> t:<valtype>                => l t
-case          ::= l:<labellit> t?:<valtype>? 0x00         => (case l t?)
-labellit      ::= len:<u32> l:<label>                     => "l"  (if len = |l|)
+labelvaltype  ::= l:<label'> t:<valtype>                  => l t
+case          ::= l:<label'> t?:<valtype>? 0x00           => (case l t?)
+label'        ::= len:<u32> l:<label>                     => l    (if len = |l|)
 <T>?          ::= 0x00                                    =>
                 | 0x01 t:<T>                              => t
 valtype       ::= i:<typeidx>                             => i
@@ -230,9 +231,9 @@ instancedecl  ::= 0x00 t:<core:type>                      => t
                 | 0x01 t:<type>                           => t
                 | 0x02 a:<alias>                          => a
                 | 0x04 ed:<exportdecl>                    => ed
-importdecl    ::= na:<nameattributes> et:<externtype>     => (import na et)
-exportdecl    ::= na:<nameattributes> et:<externtype>     => (export na et)
-externtype    ::= 0x00 0x11 i:<core:typeidx>              => (core module (type i))
+importdecl    ::= in:<importname'> ed:<externdesc>        => (import in ed)
+exportdecl    ::= en:<exportname'> ed:<externdesc>        => (export en ed)
+externdesc    ::= 0x00 0x11 i:<core:typeidx>              => (core module (type i))
                 | 0x01 i:<typeidx>                        => (func (type i))
                 | 0x02 b:<valuebound>                     => (value b) 🪙
                 | 0x03 b:<typebound>                      => (type b)
@@ -259,13 +260,11 @@ Notes:
   [TODO](Concurrency.md#TODO).
 * Validation of `resourcetype` requires the destructor (if present) to have
   type `[i32] -> []`.
-* In addition to the validation rules for `alias` *definitions* mentioned above,
-  validation of `alias` *declarators* also requires:
-  * `export` aliases only alias the `instance` and `type` sorts
-  * `outer` aliases only alias the `core type` and `type` sorts
+* Validation of `instancedecl` (currently) only allows the `type` and
+  `instance` sorts in `alias` declarators.
 * As described in the explainer, each component and instance type is validated
-  with an initially-empty type index space. (Outer aliases can be used to pull
-  in type definitions from containing `component` and `type` scopes.)
+  with an initially-empty type index space. Outer aliases can be used to pull
+  in type definitions from containing components.
 * `exportdecl` introduces a new type index that can be used by subsequent type
   definitions. In the `(eq i)` case, the new type index is effectively an alias
   to type `i`. In the `(sub resource)` case, the new type index refers to a
@@ -273,15 +272,13 @@ Notes:
   index spaces. (Note: *subsequent* aliases can introduce new type indices
   equivalent to this fresh type.)
 * Validation rejects `resourcetype` type definitions inside `componenttype` and
-  `instancetype`. Thus, handle types inside a `componenttype` can only refer
+  `instancettype`. Thus, handle types inside a `componenttype` can only refer
   to resource types that are imported or exported.
-* `<label>` is defined as part of the
-   [text format](Explainer.md#import-and-export-definitions).
 * All parameter labels, result labels, record field labels, variant case
   labels, flag labels, enum case labels, component import names, component
   export names, instance import names and instance export names must be
   [strongly-unique] in their containing scope.
-* Validation of `externtype` requires the various `typeidx` type constructors
+* Validation of `externdesc` requires the various `typeidx` type constructors
   to match the preceding `sort`.
 * (The `0x00` immediate of `case` may be reinterpreted in the future as the
   `none` case of an optional immediate.)
@@ -305,6 +302,7 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift 
            | 0x05                                                => (canon task.cancel (core func)) 🔀
            | 0x0a v:<valtype> i:<u32>                            => (canon context.get v i (core func)) 🔀
            | 0x0b v:<valtype> i:<u32>                            => (canon context.set v i (core func)) 🔀
+           | 0x0c cancel?:<cancel?>                              => (canon thread.yield cancel? (core func)) 🔀
            | 0x06 async?:<async?>                                => (canon subtask.cancel async? (core func)) 🔀
            | 0x0d                                                => (canon subtask.drop (core func)) 🔀
            | 0x0e t:<typeidx>                                    => (canon stream.new t (core func)) 🔀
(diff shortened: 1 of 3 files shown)


design/mvp/canonical-abi/definitions.py
```diff
@@ -1081,15 +1082,17 @@ def alignment(t, ptr_type):
     case CharType()                  : return 4
     case StringType()                : return ptr_size(ptr_type)
     case ErrorContextType()          : return 4
-    case ListType(t, l)              : return alignment_list(t, l, ptr_type)
+    case ListType(t, l, var)          : return alignment_list(t, l, var, ptr_type)
```
cpetig: I can't tell whether this is just a font issue, but this line looks wrongly indented.

design/mvp/canonical-abi/definitions.py
```diff
@@ -1139,15 +1145,19 @@ def elem_size(t, ptr_type):
     case CharType()                  : return 4
     case StringType()                : return 2 * ptr_size(ptr_type)
     case ErrorContextType()          : return 4
-    case ListType(t, l)              : return elem_size_list(t, l, ptr_type)
+    case ListType(t, l, var)          : return elem_size_list(t, l, var, ptr_type)
```
cpetig: another wrong indentation?

cpetig (commented):
mostly indentation doubts

design/mvp/canonical-abi/definitions.py
```diff
@@ -1301,8 +1311,15 @@ def lift_error_context(cx, i):
...
-def load_list(cx, ptr, elem_type, maybe_length):
+def load_list(cx, ptr, elem_type, maybe_length, maybe_variable=False):
   if maybe_length is not None:
+    if maybe_variable:
+      lentype = varint_type(maybe_length)
+      actual_length = load_int(cx, ptr, elem_size(lentype, cx.opts.memory.ptr_type()))
+      trap_if(actual_length > maybe_length)
+      ptr += elem_size(lentype, cx.opts.memory.ptr_type())
+      ptr = align_to(ptr, alignment(elem_type, cx.opts.memory.ptr_type()))
+      return load_list_from_valid_range(cx, ptr, actual_length, elem_type)
```
lukewagner:
```suggestion
      return load_list_from_range(cx, ptr, actual_length, elem_type)
```
I think we still need the bounds check (e.g., if we run off the end of linear memory)

design/mvp/canonical-abi/definitions.py
```diff
@@ -1301,8 +1311,15 @@ def lift_error_context(cx, i):
 MAX_LIST_BYTE_LENGTH = (1 << 28) - 1
 assert(REALLOC_I32_MAX > 2 * MAX_LIST_BYTE_LENGTH)
 
-def load_list(cx, ptr, elem_type, maybe_length):
+def load_list(cx, ptr, elem_type, maybe_length, maybe_variable=False):
```
lukewagner:
```suggestion
def load_list(cx, ptr, elem_type, maybe_length, maybe_variable):
```
IIUC, there's only just the one caller, so no need for default

design/mvp/canonical-abi/definitions.py
```diff
@@ -1601,8 +1618,16 @@ def store_probably_utf16_to_latin1_or_utf16(cx, src, src_code_units):
 def lower_error_context(cx, v):
   return cx.inst.handles.add(v)
 
-def store_list(cx, v, ptr, elem_type, maybe_length):
+def store_list(cx, v, ptr, elem_type, maybe_length, maybe_variable=False):
```
lukewagner:
```suggestion
def store_list(cx, v, ptr, elem_type, maybe_length, maybe_variable):
```

design/mvp/canonical-abi/definitions.py
```diff
@@ -1601,8 +1618,16 @@ def store_probably_utf16_to_latin1_or_utf16(cx, src, src_code_units):
...
   if maybe_length is not None:
+    if maybe_variable:
+      assert(len(v) <= maybe_length)
+      lentype = varint_type(maybe_length)
+      store_int(cx, len(v), ptr, elem_size(lentype, cx.opts.memory.ptr_type()))
+      ptr += elem_size(lentype, cx.opts.memory.ptr_type())
+      ptr = align_to(ptr, alignment(elem_type, cx.opts.memory.ptr_type()))
+      store_list_into_valid_range(cx, v, ptr, elem_type)
```
lukewagner:
```suggestion
      store_list_into_range(cx, v, ptr, elem_type)
```

design/mvp/canonical-abi/definitions.py
```diff
@@ -1831,11 +1858,20 @@ def lift_flat_string(cx, vi):
   packed_length = vi.next(cx.opts.memory.ptr_type())
   return load_string_from_range(cx, ptr, packed_length)
 
-def lift_flat_list(cx, vi, elem_type, maybe_length):
+def lift_flat_list(cx, vi, elem_type, maybe_length, maybe_variable=False):
```
lukewagner:
```suggestion
def lift_flat_list(cx, vi, elem_type, maybe_length, maybe_variable):
```

design/mvp/canonical-abi/definitions.py
```diff
@@ -1831,11 +1858,20 @@ def lift_flat_string(cx, vi):
...
   if maybe_length is not None:
+    actual_len = maybe_length
+    if maybe_variable:
+      lentype = varint_type(maybe_length)
+      actual_len = lift_flat(cx, vi, lentype)
+      trap_if(actual_len > maybe_length)
```
lukewagner:
```suggestion
    if maybe_variable:
      lentype = varint_type(maybe_length)
      actual_len = lift_flat(cx, vi, lentype)
      trap_if(actual_len > maybe_length)
    else:
      actual_len = maybe_length
```

design/mvp/canonical-abi/definitions.py
```diff
@@ -1730,15 +1755,17 @@ def flatten_type(t, opts):
...
     case VariantType(cases)               : return flatten_variant(cases, opts)
     case FlagsType(labels)                : return ['i32']
     case OwnType() | BorrowType()         : return ['i32']
     case StreamType() | FutureType()      : return ['i32']
 
-def flatten_list(elem_type, maybe_length, opts):
+def flatten_list(elem_type, maybe_length, maybe_variable, opts):
   if maybe_length is not None:
+    if maybe_variable:
+      return flatten_type(varint_type(maybe_length), opts) + flatten_type(elem_type, opts) * maybe_length
```
lukewagner: This is pre-existing, but a question: for flattening bounded- or fixed-length lists: should we have some low (e.g., 4) per-value maximum flattening length, beyond which the list gets passed via pointer?  I'm mostly thinking of the case where there is some list that has a large bound, and it gets used as a function parameter, and it "blows" the `MAX_FLAT` budget, causing all the parameters to go into the heap when instead you probably wanted the bounded-/fixed-length list to go in the heap.

design/mvp/Binary.md
```diff
@@ -282,8 +283,9 @@ Notes:
   to match the preceding `sort`.
 * (The `0x00` immediate of `case` may be reinterpreted in the future as the
   `none` case of an optional immediate.)
-* 🔧 for fixed-sized lists the length of the list must be larger than 0 to pass
-  validation.
+* 🔧 for fixed-sized lists (`0x67`) the length of the list must be larger than
```
lukewagner: Pre-existing, but it looks like the grammar already covers this twice: once by using `<u32>` (unsigned) and once with the `(if maxlen > 0)`.  We could also remove the `(if maxlen > 0)`.  But should we specify a maximum for `maxlen`?

lukewagner (commented):
Thanks!  This is looking generally good, a few comments:

design/mvp/Binary.md
Felix-El: For u32, zero is valid and the `maxlen > 0` checks forbids zero. What would be a legitimate upper bound... `i32::MAX`?

design/mvp/canonical-abi/definitions.py
Felix-El: From my perspective (one who wants to reduce/avoid allocations) this would make the data type harder to reason about. Fixed and bounded lists should be guaranteed to be "static".
TBH I'd rather say that some datatypes (like these special lists) should enforce* the use of a param or return area for passing - because lists are indexed dynamically unlike tuples (-> indirect addressing works only on memory locations and compilers would otherwise have to compensate for such lists split across registers - pretty terrible).

(enforce: such lists should definitely be placed into param/return area but other types, if possible could continue to be passed by args... but yeah, this obviously complicates the flattening logic)

design/mvp/Binary.md
lukewagner: There's already (just recently added) a `MAX_LIST_BYTE_LENGTH` = 2<sup>28</sup>-1.  That's just bytes, but even still, it seems like a reasonable upper bound.

design/mvp/canonical-abi/definitions.py
lukewagner: Ah, so then could we say that fixed- and bounded-length lists *never* get flattened, and are always passed as pointer (and maybe length if fixed)?

flelchuk: Yes this is what I mean (except: length if *bounded*, that could still be in a register).
I will explore how that logic extension could look (in Python) in the next commit.

design/mvp/Binary.md
Felix-El: Ok, defined that bound in `BINARY.md` but where would we check this in definitions.py ?

design/mvp/canonical-abi/definitions.py
Felix-El: Just pushed a proposal for hybrid lifting/lowering where register and memory passing can coexist.

Felix-El: # Flat-calling-convention treatment of fixed-length and bounded lists

## Background

The canonical ABI flattens component-level values into Wasm core function
parameters.  A `MAX_FLAT` cap (currently 16) limits how many flat slots a call may use before everything is
redirected through a *param-area* (a caller-allocated memory block whose
pointer is passed in r0).

The current implementation flattens fixed-length lists element-by-element
(`list<T, N>` → N flat slots) and bounded lists with a leading length register
(`list<T, ..N>` → 1 + N flat slots). This is wasteful in several ways.

A key property of fixed and bounded lists is that they can be passed without
additional allocations beyond the param area itself: because the maximum byte
count is statically known from the type, the caller can reserve space inline
rather than performing a separate heap allocation.

## Why the current strategy is problematic

### Semantic mismatch

List element access is indexed by a **runtime value**.  Even if all elements
arrive in registers, the callee must immediately spill them to memory before it
can iterate or index, because registers are not addressable.  Tuples do not
have this problem — each field is a compile-time-known register slot.

Passing list elements in registers therefore adds round-trip spill cost with no
benefit.

### Budget poisoning

When the flat slots for a list exceed `MAX_FLAT`, the ABI falls back to the
classic all-params-to-param-area strategy: **all** arguments, including simple
scalars that would fit comfortably in registers, are bundled into the param
area and only a single pointer is passed.  The list causes unrelated arguments
to pay an indirection penalty.

Example: `f(list<s32, 16>, s32)` needs 17 slots (16 elements + 1 scalar).  With
`MAX_FLAT=16` the scalar ends up in memory alongside the list — even though
there was a free register waiting for it.

## Proposal

Allow list elements to occupy the **param area** without consuming flat-register slots.

- Fixed list: **0** flat slots; list elements are stored in the param area
  (if no param area is required by other arguments, one is created and the
  elements occupy it starting at offset 0).
- Bounded list: **0** flat slots; a varint length prefix followed by the list
  elements are stored in the param area (if no param area is required by other
  arguments, one is created, with the varint prefix at offset 0 followed by the
  elements).
- After processing all typed arguments, if a param area is needed, one
  additional flat slot (param area address) is appended to the flat list.
- If the total flat slots now exceed `MAX_FLAT`, fall back to the classic
  all-params-to-param-area strategy for backward compatibility.

Benefits: scalar arguments that fit within `MAX_FLAT` stay in registers even
when a list is present.  Both fixed and bounded lists use the same in-memory
layout (`elements` for fixed, `varint | elements` for bounded) whether they
appear as top-level flat arguments or as elements of an outer fixed/bounded
list — the representation composes uniformly.
Cost: the extra param area address slot can tip the budget when many scalar arguments
accompany a list, at which point the classic fallback fires. Potentially we could excempt the param area pointer from MAX_FLAT budget, like result area pointer is.

Felix-El: Probably worth a separate discussion though I wanted to show a more natural way to deal with in-place lists.

cpetig: I shy away from the added complexity of mixed flattened and in-memory arguments. Up to now either all arguments are in registers or in memory.

Also I just realized that for guest imported functions `list<>` doesn't have to be heap allocated, as it is passed by reference, similarly guest exported results (custom cabi_post can replace malloc/free with arenas). Only the two other cases require `cabi_realloc` (which could still be different from malloc): Guest import result and guest export arguments. So this representation is only applicable to guest export (host calls into guest) argument representation.

-----

Also I don't want to derail this discussion with just another wild idea, but I have been thinking about using SIMD/vector encoding of fixed length lists into registers, e.g. passing an `[u8; 8]` in an `u64`. 

I think with SIMD and shift instructions this is efficient enough to access/process and helps reducing excessive register grabbing in the flattened case. My guess is that `[u8; 8]` to `[u8; 64]` would be a common enough structure element to make this worthwhile.

lukewagner: > Even if all elements arrive in registers, the callee must immediately spill them to memory before it
can iterate or index, because registers are not addressable.

This seems like a good reason to always put the elements of bounded-/fixed-length lists in memory.  But, agreeing with @cpetig's point, I'd suggest:
* when lifting: work just like regular lists (flatten to `i32` pointer and, if bounded, `i32` length)
* when lowering: force *all* params/results to fall back to using the in-memory param/result region (perhaps by adding a `contains_inline_list` predicate styled on `contains_borrow` that is tested first in `flatten_functype`)

This still leaves a `cabi_realloc` for the lowered params of exports, but lazy-lowering ABI should zap this one in general in the future so I think it's fine to leave it for now to avoid added complexity.

cpetig: After a longer call with Felix I propose the following summary of what we could agree on (by some reading this could be what we have proposed above; I use an overly specific C or Rust like example to avoid my vague understanding of the term lifting and lowering):

```
  r: record { bounded: list<u32, 0, 32>, integer: u32 }
  f: func(arg: r) -> r;
```

The caller of the function (guest or host side) allocates `b: [u32; 66]` (32+34)*4 bytes somewhere (could be linear stack, cabi_realloc, lazy lowered, etc). Then because the arguments can be flattened (with this new scheme) it passes (addr=`&b[0]`, len, integer, retarea=`&b[32]`) to the function.

The function implementation can handle the list argument like a normal list, but must not free the memory. Because multi-value is not implemented yet, the result can't be flattened and thus is stored in the return value pointer provided by the caller (b[32] = length, b[33..64]= data, b[65]=integer). Multi-value would make the result area shrink to 64*4 bytes and return (addr, len, integer).

No allocation/de-allocation within the function call, the argument pointer and result pointer can be re-used across calls.

@Felix-El  proposed to also apply this to fixed length lists and I agree that being able to iterate over arrays is at odds with passing them in individual register/value-stack locations. (Consider my implementation as change-averse because it was more identical to the existing tuple layout, but I see the value of this separation into pointer (flat) and data (memory)).

Felix-El: Yes, this is different from what I proposed in the [lenghty comment](#discussion_r3409589269) and might be closer to what @lukewagner meant. I'll try to rephrase in my own words:

# Representation

| Type                | Flat         | Memory                     |
| ------------------- | ------------ | -------------------------- |
| fixed-length list   | `[ptr]`      | `[data0, ..., dataN-1]`      |
| bounded-length list | `[ptr, len]` | `[len, data0, ..., dataN-1]` |

# Semantics of Passing Bounded/Fixed Length Lists

- unlike dynamic lists, the memory is *caller-owned*
- as an optimization, the caller may provide a single allocation
  for all bounded/fixed length lists and even the return area
  (alignment must be ensured though)
- (*multi-value result only*) for *flattened* bounded/fixed length lists in results, the caller has to provide memory (for all lists in sequence, alignment must be ensured) and the callee will write into that memory
  - this will only occur when there is no return area, so the "return area pointer" core parameter can be reused for this (somewhat similar) matter
  - bounded/fixed length lists *inside dynamic lists* are *not flattened* so their memory demand is part of the list's allocation
  - each *flattened* list's returned `ptr` lies within the caller-provided memory

cpetig: (comment about ambiguous terms was already integrated above)

Felix-El: @lukewagner,

> * when lowering: force _all_ params/results to fall back to using the in-memory param/result region (perhaps by adding a `contains_inline_list` predicate styled on `contains_borrow` that is tested first in `flatten_functype`)

This part clearly sounds like "let's just enforce param/return area use if *any* fixed/bounded list exists outside of *unbounded* lists".

However, kindly elaborate on

> * when lifting: work just like regular lists (flatten to `i32` pointer and, if bounded, `i32` length)

which I just can't follow. Are you suggesting that the host would get this inlined from the caller (import side) but then passes as flat `(ptr, len)` to the callee (export side)? And the reverse for the result path?

lukewagner: @cpetig Several parts of that match what I'd expect, but a few questions:

> The caller of the function (guest or host side) allocates b: [u32; 66] (32+34)*4 bytes somewhere

Just to clarify: while the caller *could* allocate the param and result buffers in a single contiguous allocation, because the core function signature of the import is `(addr, len, integer, retarea)`, the caller could also allocate disjoint regions for `addr` and `retarea`, yes?  I.e., there's not a *requirement* of contiguity.

> The function implementation can handle the list argument like a normal list, but must not free the memory.

(I believe we're talking about the core function signature of guest exports here.) So a normal list will call `cabi_realloc` for the list elements of each `list` parameter, but I'm guessing that's not what you mean here (b/c of the "must not free memory" clause).  Now what I suggested in the second bullet of my last comment was that all parameters go into the heap as a single contiguous `tuple` (just like when `> MAX_FLAT_PARAMS`), and so there is a single `cabi_realloc` call (containing `r` with the bounded list inline).  But it sounds like you mean something different; given the current eager ABI, I think a `cabi_realloc` has to happen *somewhere* for bounded-/fixed-length list parameters of core wasm exports, so what are you thinking?

> Because multi-value is not implemented yet, the result can't be flattened and thus is stored in the return value pointer provided by the caller (b[32] = length, b[33..64]= data, b[65]=integer). 

If we're still taking about the core function signature of a guest export, I don't know where this "return value pointer provided by the caller" comes from... is it allocated at the end of the `cabi_realloc` (for params) mentioned above?  That seems like kindof a bigger departure from the rest of the current ABI.  The way it works today if you return a `list` (which never fits in `MAX_FLAT_RESULTS=1`) is you have to return an `i32` pointer to a `record {ptr: u32, len:u32}` in memory.  Instead,  for bounded-/fixed-length lists I would imagine that we'd say that the return value would be an `i32` pointer to `r` in memory (which contains the bounded list and `len` inline, avoiding an indirection).

@Felix-El Sorry, what I wrote about "lifting" was imprecise and only considered the "guest lifting params to import call" case (which matches what @cpetig wrote above, iiuc).

cpetig: @lukewagner You caught me misremembering details about the argument/result representation (I keep pruning the number of calling convention variants in my mind over time, likely because I added another one)

Guest calling guest imported function of above type:
Caller reserves **34** u32 somewhere (return area, e.g. on linear stack), passes (addr, len, integer, retarea) calling into the host. Host stores result in retarea and returns.

Host calling guest exported function of above type:
Host finds **len** u32 in linear memory from somewhere (e.g. a previous remembered `cabi_realloc`) places the active members of the argument in there and calls the guest with (addr, len, integer). Guest returns **34** u32 in arbitrary location (e.g. static return area), host reads from there and then calls `cabi_post_X` (preferably before calling again into the function).

The host can reuse the argument storage buffer many times.

If the number of flat elements grow beyond 16 guest calling host will simplify into `void f(R const*input, R*retarea)` and host calling guest into `R const* f(R const*input)`. (all R point to u32[34])

Lazy lifting will likely replace the pointers with a handle.

PS: Yes, I meant that the two buffers wouldn't need to be contiguous, but this is moot as I was mixing up import and export allocation schemes.

cpetig: PPS: I tend to propose to change the representation of **flattened** fixed length lists to align with this new convention (passing pointer) - at the expense of becoming different from tuples (which are deeply flattened)

cpetig: P3S: This should be identical to what currently happens, but with the new proposal in flattened form these arrays would pass a pointer to some arbitrary (!) place where a tuple would pass the values themselves.

cpetig:
Associated with https://github.com/orgs/bytecodealliance/projects/19/

Felix-El:
Here is another shot at an ABI tweak for efficient passing of fixed/bounded lists. It is already implemented in the commit I just pushed. @cpetig 's ideas with refinement by discussion, my write-up; more versatile / less invasive compared to my previous proposal.

---

# Flat ABI for Fixed and Bounded Length Lists

## Motivation

The current flat ABI for fixed length lists expands all N element slots
into the core function's parameter/result list.
For large N this causes parameter overflow (exceeding
`MAX_FLAT_PARAMS` or `MAX_FLAT_RESULTS`), forcing an early fallback to
memory passing for the entire argument list when even one or two lists
are large.

If the current approach was naively adopted for bounded length lists
(which are still theory at this time), expanded form would even waste
flat-register space on uninitialized elements for partially-populated lists.

A second, independent motivation is that fixed and bounded length lists
are, unlike tuples, indexed at runtime and registers in core WebAssembly
cannot be indexed efficiently. Therefore, the expanded-element flat representation
forces register/memory (de)composition on either guest, adding to overhead.

This proposal replaces the expanded-element flat representation with a
pointer-based one for *direct* fixed and bounded length lists in parameters —
lists which are eligible for flat representation (not nested, immediately or
transitively, in another `list`).

## Flat representation

| Type                     | Flat representation                                                                            |
| ------------------------ | ---------------------------------------------------------------------------------------------- |
| `list<T>` (unbounded)    | `[ptr, len]` — pointer to `len` contiguous elements + actual length (unchanged, for reference) |
| `list<T, N>` (fixed)     | `[ptr]` — one pointer to `N` contiguous elements                                               |
| `list<T, ..N>` (bounded) | `[ptr, len]` — pointer to `len` contiguous elements + actual length                            |

## Memory representation

| Type                     | Memory representation                                                                          |
| ------------------------ | ---------------------------------------------------------------------------------------------- |
| `list<T>` (unbounded)    | `[ptr, len]` — pointer to `len` contiguous elements + actual length (unchanged, for reference) |
| `list<T, N>` (fixed)     | `[e0, e1, ...]` — `N` contiguous elements                                                      |
| `list<T, ..N>` (bounded) | `[len*, e0, e1, ...]` — `N` contiguous elements, uninitialized beyond actual length            |

bounded actual *len*: smallest integer type that can represent values up to (including) N

## Lowering and Lifting

### Parameters

Below explanation is for bounded length lists which carry a runtime length.
The flow is similar for fixed length lists - length just would not be passed
as it is statically known.

If all core params fit within `MAX_FLAT_PARAMS`, they are passed flat, otherwise
we fall back to a in-memory tuple representation (*param area*). This is *not new*.

1. Caller
   - assume list `L` is already constructed in guest memory as `L_ptr` and `L_len` (from API use)
   - if passing flat: pass `L_ptr` and `L_len` directly, classic unbounded `list` ownership semantics
   - if passing in memory: move list elements inplace into param area

2. Host, before call forwarded
   - if passing flat:
     - allocate twin list `L2` in the exporting guest (optimization: bulk alloc, cache, etc.)
     - move elements between backing memories (import -> export)
     - pass `L2_ptr` and `L2_len = L_len`, classic unbounded `list` ownership semantics
   - if passing in memory:
     - move elements from inbound param area to outbound param area

3. Callee
   - if passing flat: take `L2_ptr` and `L2_len` from core params
   - if passing in memory: read `L2_len` from param area, let `L2_ptr` point to first inplace element in param area
   - either way: construct list from `L2_ptr`, `L2_len` for API use

### Results

Memory-passing is enforced when even a single *direct* fixed/bounded length list is in the result type -
that's a *new* second switchover reason beyond exceeding `MAX_FLAT_RESULTS`.
A return area (*ret area*) is used to forward all results as an in-memory tuple.

1. Caller
   - result area is preallocated with enough room to hold all fixed/bounded length lists inline
   - result area pointer is passed as out core param

2. Host, after call forwarded
   - copy inplace elements from returned export-side ret area into import-side ret area provided by out param

3. Callee
   - construct the list by moving elements from ret area for API use

### Fused allocation

When multiple direct lists exist among parameters and passing flat, the host might service
their backing (export) memory needs from a single allocation as an optimization.

design/mvp/canonical-abi/definitions.py
Felix-El: Started a fresh comment thread below that gathers and compresses the ideas.

lukewagner:
Thanks for the clear write-up (and sorry for the delay replying)!  That all sounds pretty good; just one suggestion: if a component *exports* a function with type `func(p: list<u8, 4>, q: list<u8, 4>)`, iiuc, the above suggests there would be 2 `realloc` calls in the callee for `p` and `q`.  I wonder if we should instead have this work like return values: everything goes into a contiguous param-buffer (so that there's always only 1 `realloc` call), as-if `MAX_FLAT_PARAMS` had been exceeded.  (A component that *imported* this function type would work the same as you outlined.)  WDYT?

Felix-El:
> if a component _exports_ a function with type `func(p: list<u8, 4>, q: list<u8, 4>)`, iiuc, the above suggests there would be 2 `realloc` calls in the callee for `p` and `q`.

I addressed that in the trailing "fused allocation" section - yes it should be done but we need not guarantee that by design. In practice, the host can alloc one big chunk for everything and hand out pointers to subobjects (list1, list2, list3, etc. correctly aligned and sized of course). The allocation will continue to be owned by the host.

> I wonder if we should instead have this work like return values: everything goes into a contiguous param-buffer (so that there's always only 1 `realloc` call), as-if `MAX_FLAT_PARAMS` had been exceeded.

Passing *all* params in a param-area, even if callee side only, feels a bit inefficient. We should try to utilize registers for non-list value passing.
Additionally it would break the symmetry between caller core func signature and callee core func signature much further than just the current return-area result vs. out pointer story.

But that's just my 2 cents...

PS: `definitions.py` modifications compared between the [previous/initial proposal](https://github.com/WebAssembly/component-model/pull/640#discussion_r3409589269) and this one, I have to admit this one felt way less intrusive to implement (thinking of differences in `definitions.py` in 91a0eae7b3157666ce561b15b388fd52cab575f9 vs 1fb578ef9626380211e8c0b14360350c1fe31175 over the predecessor.)

lukewagner:
> I addressed that in the trailing "fused allocation" section - yes it should be done but we need not guarantee that by design. 

I think the *host* can do that when given a list, b/c the host can do whatever, but if we're talking about lowering a list into the *guest*, the `realloc` calls would have to happen exactly when specified, so no fusion would be possible when lowering a fixed-/bounded-length list into guest code.

> Passing all params in a param-area, even if callee side only, feels a bit inefficient.

I suppose, but I would expect this to be amortized by the cost of at least 1 `realloc`.  FWIW, I think the way `realloc` would get removed in a lazy-lowering future is that, when parameters aren't flat (including when there are any fixed- or bounded-length lists), the callee entry function would call a (new) `task.start` built-in passing a pointer to a buffer to receive all the parameters, which is sortof analogous.

flelchuk:
> I think the *host* can do that when given a list, b/c the host can do whatever, but if we're talking about lowering a list into the *guest*, the `realloc` calls would have to happen exactly when specified, so no fusion would be possible when lowering a fixed-/bounded-length list into guest code.

Sorry if I'm repeating myself. I might be missing some crucial C-M insight...

- This proposal leaves most of lifting/lowering as-is, including dynamic lists.
- The only special new treatment is for fixed/bounded lists.
  - Memory path: all params in param area - caller passes to host, host copies*, callee receives param area. Fixed/bounded lists are inlined.
  - Register path: fixed/bounded lists enter the host as 1-2 slots each. The host allocates export-side memory *possibly as a single longlived allocation* holding elements of all lists, host copies* elements and forwards to the callee also as 1-2 slots each. *The callee uses memory from host-given address directly, no further realloc needed*
- Return is always inlined in ret area.

What obligation would make fusion impossible?

cpetig:
> * Register path: fixed/bounded lists enter the host as 1-2 slots each. The host allocates export-side memory _possibly as a single longlived allocation_ holding elements of all lists, host copies* elements and forwards to the callee also as 1-2 slots each. _The callee uses memory from host-given address directly, no further realloc needed_

My understanding is that this logic is already needed for the argument area whenever calling into the guest with more than 16 flat parameters. We simply start using this linear memory argument area earlier for spilling over the fixed/bounded list data when the rest of the arguments still fit into registers.

cpetig:
(see https://github.com/WebAssembly/WASI/issues/944 for previous status updates and reasoning)

I implemented more of this logic inside wasmtime and it needed a larger restructuring in the environ crate - because now you should allocate memory across all arguments (already default when not flattened) and pass flattened arguments pointing there. So the calculation of flattened properties needs to pass this information on. Still working on this.

PR #641, opened 2026-04-27
Emoji-gate synchronous future/stream read/write

alexcrichton:
I understand that this change is "late in the game" for component-model-async things with WASIp3 around the corner, but in light of bytecodealliance/wasmtime#13212 I'd like to try to at least propose gating the ability to use `{future,stream}.{read,write}` synchronously for the initial WASIp3 0.3.0 release. I'm not aware of any preexisting language bindings actually using these functions, for example all `wit-bindgen` languages are using the `async` version. Given that lack of use, and now-cropping-up interesting questions, I'd like to ideally gate these to be sure to have enough time to come back in the future and thoroughly consider the interleavings and repurcussions here.

Specifically this change proposes moving the non-`async` functionality of these intrinsics behind the 🚝 feature gate. This preexisting feature gate mostly gated async-related builtins but it's generalized now to basically be "more options on more builtins" as a sort of grab-bag for now.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -4356,6 +4356,7 @@ specifies:
   * [`lower($t)` above](#canonopt-validation) defines required options for `stream.write`
   * [`lift($t)` above](#canonopt-validation) defines required options for `stream.read`
   * `memory` is required to be present
+* 🚝 - `async` is allowed to be omitted, otherwise it must be present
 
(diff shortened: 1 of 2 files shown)


lukewagner:
Makes sense to me to gate this if we're not using it from any producer toolchains atm; it's just an optimization.

dicej (approved):

Issue #423, opened 2024-04-27
Add runtime instantiation to component model


pothos:
Is there anyone already working on a draft for adding runtime instantiation to the component model?

Runtime instantiation means that a component can dynamically instantiate another component given its binary representation, and then invoke them. This is needed to have something similar to `exec`, where the number of imported instances and even the module code itself is not statically predefined but evolves programmatically. Currently one could define a custom protocol between the runtime and a component to have a similar effect. Use cases are flexible reusing of modules, runtime generation or fetching of modules (network or from disk), and a small demo program you could think of is a wasm component shell or wac REPL that itself runs as wasm component.

A few links in Luke's answer here: https://github.com/WebAssembly/wasi-cli/issues/25

lukewagner:
(Oh haha, I was confused by my above comment before seeing that Dan transferred it :P)

I'm not working on a draft yet, but once 0.3-async is closed out, this is one of the last significant feature I am quite keen to add to the CM for a 1.0 MVP, because I totally agree with those use cases you mention being important.

The rough shape of the feature here is a new [canon built-in](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#canonical-built-ins) that allows "resourceifying" any component (nested or imported) to produce:
1. a new resource type representing instances of that component
2. a new function that runtime-`instantiate`s that component, returning an owned handle of type (1)
3. for each export of that component, a function that prepends to the original parameters a borrowed handle of type (1)

But lots more work to turn this idea into a real draft proposal; I'm sorry there's not more worked out yet.

PR #643, opened 2026-04-30
Add 'Component Instance Lifetime' section to Concurrency.md

lukewagner:
This PR adds a new section proposing default expectations around component instance lifetimes in the presence of `async` based on some recent discussions with @alexcrichton, @dicej and @tschneidereit about what the default instance reuse/lifetime policies should be in Wasmtime for WASI HTTP.  Ultimately, the host can create or destroy component instances whenever it wants, so this isn't really *normative*, but it still seems useful to document default expectations is to align producer toolchains and runtimes and potentially suss out other interesting use cases.

### DIFF
--- a/design/mvp/Concurrency.md
+++ b/design/mvp/Concurrency.md
@@ -1344,13 +1344,12 @@ cooperative threads (🧵), [`setInterval`] would also unfortunately delay
 returning the `EXIT` code and thus, without guest code intervention, would keep
 component instances alive until timeout limits were hit. The release of
 cooperative threads would offer a solution to this problem, but an awkward one.
-Instead, the [intention](#TODO) is to add a new built-in function (say,
-`thread.set-keep-alive`) to allow guest code to dynamically control its own
-threads' keep-alive flags, thereby allowing the JS event loop to clear its
-implicit thread's keep-alive flag once all `waitUntil` promises resolved,
-thereby allowing `setInterval` callbacks to keep running (while the host wants
-to keep the instance warm), but still indicating to the host that destruction is
-welcome at any time.
+Instead, the [intention](#TODO) is to add new built-in functions that would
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
lgtm 👍 

FWIW the companion Wasmtime PR is https://github.com/bytecodealliance/wasmtime/pull/13246

badeend (approved):
This maps pretty much exactly how I'd expect it to behave. :+1:

---

The current wording is tied specifically to threads, tasks, streams & futures. Maybe this could be generalized into a single conceptual `keep-alive` counter at the component instance level?
- Start of export func task: +1
- Task returned: -1
- Resource handed out: +1
- Resource dropped: -1
- etc.

and for additional manual control: `keep-alive.inc` & `keep-alive.dec` (similar to `backpressure.inc` & `backpressure.dec`)

lukewagner:
@badeend Ah, cool idea!  Thinking about whether I could take this all the way into the Python CABI code, the tricky bit is that we only care about stream/future ends when the host is holding the other end; if we inc/dec for component-to-component end pairs, I think that'll end up keeping component instances alive longer than expected.  At the moment, things are intentionally set up to be rather agnostic to each end of the stream of whether the other end of the stream is guest/host.  Not that this couldn't be fixed, but it would be a more invasive fix than adding a few well-placed increments and decrements.  I expect when we get to runtime instantiation (#423) these rules will end up getting reified in the semantics of `resource.drop`ing a component instance, so that would be an opportunity to formalize this a bit more.

Also, good point about `keep-alive.{inc,dec}` (vs. `set`, to avoid the clobbering problems); I was wondering about that too.

lukewagner:
@badeend ... and it's also a good point that the built-ins might not be *thread*-specific, they might be *instance*-specific.  I tweaked the wording to not presume the exact solution and just state the goal.

PR #645, opened 2026-05-04
Restrict all context.{get,set} in same component to use same elem type

lukewagner:
This PR adds validation rules that require all `context.get` and `context.set` definitions inside a component to use the same element type (currently either `i32` or, with #624, `i64`).  Conceptually, this validation rule would be attached to what it means for a `component` definition to be valid.  In the short-term, this might allow `context.set i32` to avoid a store-zero when implemented on 32-bit CPUs (to clear the high bits of the underlying `i64` storage) if there's no whole-component information to indicate the absence of `context.set i64`s in the same component.  In the longer-term, this change would allow core wasm reference types to be efficiently used as element types, since the simple aliasing approach of `i32`/`i64` wouldn't work.

(cc @adambratschikaye)

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -214,12 +214,9 @@ label'        ::= len:<u32> l:<label>                     => l    (if len = |l|)
                 | 0x01 t:<T>                              => t
 valtype       ::= i:<typeidx>                             => i
                 | pvt:<primvaltype>                       => pvt
-resourcetype  ::= 0x3f 0x7f f?:<core:funcidx>?            => (resource (rep i32) (dtor f)?)
-                | 0x3e 0x7f f:<core:funcidx>
-                            cb?:<core:funcidx>?           => (resource (rep i32) (dtor async f (callback cb)?)) 🚝
-                | 0x3f 0x7e f?:<core:funcidx>?            => (resource (rep i64) (dtor f)?) 🐘
-                | 0x3e 0x7e f:<core:funcidx>
-                            cb?:<core:funcidx>?           => (resource (rep i64) (dtor async f (callback cb)?)) 🚝🐘
+resourcetype  ::= 0x3f v:<valtype> f?:<core:funcidx>?     => (resource (rep v) (dtor f)?)
+                | 0x3e v:<valtype> f:<core:funcidx>
+                                   cb?:<core:funcidx>?    => (resource (rep v) (dtor async f (callback cb)?)) 🚝
 functype      ::= 0x40 ps:<paramlist> rs:<resultlist>     => (func ps rs)
                 | 0x43 ps:<paramlist> rs:<resultlist>     => (func async ps rs)
 paramlist     ::= lt*:vec(<labelvaltype>)                 => (param lt)*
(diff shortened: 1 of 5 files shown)


lukewagner:
Additionally, the second commit tweaks the binary format rules in `Binary.md` to use validation to gate `context.{get,set}` and `(type (resource (rep v)))` to `i32`/`i64` instead of decoding rules.  This mostly only shows up in whether WAST tests report a decoding vs. validation error, as highlighted in [wasm-tools/#2502](https://github.com/bytecodealliance/wasm-tools/pull/2502#discussion_r3138034918).

design/mvp/Binary.md
```diff
@@ -214,10 +214,8 @@ label'        ::= len:<u32> l:<label>                     => l    (if len = |l|)
                 | 0x01 t:<T>                              => t
 valtype       ::= i:<typeidx>                             => i
                 | pvt:<primvaltype>                       => pvt
-resourcetype  ::= 0x3f 0x7f f?:<funcidx>?                 => (resource (rep i32) (dtor f)?)
-                | 0x3e 0x7f f:<funcidx> cb?:<funcidx>?    => (resource (rep i32) (dtor async f (callback cb)?)) 🚝
-                | 0x3f 0x7e f?:<funcidx>?                 => (resource (rep i64) (dtor f)?) 🐘
-                | 0x3e 0x7e f:<funcidx> cb?:<funcidx>?    => (resource (rep i64) (dtor async f (callback cb)?)) 🚝🐘
+resourcetype  ::= 0x3f v:<valtype> f?:<funcidx>?          => (resource (rep v) (dtor f)?)
+                | 0x3e v:<valtype> f:<funcidx> cb?:<funcidx>? => (resource (rep v) (dtor async f (callback cb)?)) 🚝
```
alexcrichton: Could you expand the "Notes" section below with wording that `<valtype>` is only allowed to be `i32` or (🐘-gated) `i64`?

design/mvp/Binary.md
```diff
@@ -300,10 +298,8 @@ canon    ::= 0x00 0x00 f:<core:funcidx> opts:<opts> ft:<typeidx> => (canon lift
            | 0x25                                                => (canon backpressure.dec (core func)) 🔀
            | 0x09 rs:<resultlist> opts:<opts>                    => (canon task.return rs opts (core func)) 🔀
            | 0x05                                                => (canon task.cancel (core func)) 🔀
-           | 0x0a 0x7f i:<u32>                                   => (canon context.get i32 i (core func)) 🔀
-           | 0x0a 0x7e i:<u32>                                   => (canon context.get i64 i (core func)) 🔀🐘
-           | 0x0b 0x7f i:<u32>                                   => (canon context.set i32 i (core func)) 🔀
-           | 0x0b 0x7e i:<u32>                                   => (canon context.set i64 i (core func)) 🔀🐘
+           | 0x0a v:<valtype> i:<u32>                            => (canon context.get v i (core func)) 🔀
+           | 0x0b v:<valtype> i:<u32>                            => (canon context.set v i (core func)) 🔀
```
alexcrichton, line 303: Similar to above, mind expanding "Notes" with the 🐘-gating note?

alexcrichton (approved):

design/mvp/Binary.md
lukewagner: I could, and I even started to, but for all the other canon built-in validation rules, all the rules are in CanonicalABI.md (including the rules for `context.{get,set}`), and so the Notes section in Binary.md is almost empty (with one load-bearing bullet saying to see CanonicalABI.md for validation rules).

design/mvp/Explainer.md
```diff
@@ -1566,14 +1566,9 @@ See the [concurrency explainer] for background.
...
-be relaxed in the future.
-
-Mixing `i32` and `i64` results in truncating or unsigned extending the
-stored values:
-* If `context.get i32 i` is called after `context.set i64 i v`,
-  only the low 32-bits are read (returning `i32.wrap_i64 v`).
-* If `context.get i64 i` is called after `context.set i32 i v`,
-  the upper 32-bits will be zero (returning `i64.extend_i32_u v`).
+be relaxed in the future. Additionally, component-level validation requires
+that all `context.get` and `context.set` built-ins use the *same* `T`, so
```
michael-weigelt, line 1570: This PR is about `context.{set, get}`, but does the same not also apply for the other items in [this PR](https://github.com/WebAssembly/component-model/pull/624/changes#top)? 
Namely, 
- `waitable-set.{wait, poll}`
- `stream.{read. write}`
- `future.{read, write}`
- `error-context.{new, debug-message}`

All of these now have signatures which depend on the memory used, so they all must be consistent. 

For the `thread` parts I am not sure.

michael-weigelt:
>(cc @adambratschikaye)

CC @adamrk

design/mvp/Explainer.md
adambratschikaye: In all those cases, the type is determined by `memory` in the `canonopts`, so I don't think the issues here would apply. (and also since the types don't explicitly get parsed from the binary)

Issue #642, opened 2026-04-28
Interaction between synchronous stream/future read/write and various operations


alexcrichton:
This issue is borne out of discussion on https://github.com/bytecodealliance/wasmtime/issues/13212 where this first arose within Wasmtime. Specifically it's not clear at this time what should happen in various situations related to a thread being blocked on stream/future read/write and then doing something else, for example:

* If the stream/future was in a preexisting waitable-set, and then an event happens, where does the event go?
* If the read/write is cancelled, what happens?
* If the stream/future is added to a waitable set and an event happens, where does it go?

Overall this is the motivation for [gating these operations](https://github.com/WebAssembly/component-model/pull/641) in the upcoming WASIp3 release since there's no known usage of these variants, but regardless I wanted to file this to track clarifying/documenting these interactions.

PR #647, opened 2026-05-04
Prevent waitable-set.{wait,poll} from being used at the same time as sync built-ins

lukewagner:
This PR is one way to resolve the ambiguity discovered in #642.

One option is to give the sync operations "priority" over the waitable-set, and that's what I started with.  But it's a pretty complex behavior to test for and so it seems likely to be a bug in production.  It's also not clear if it's actually the desired behavior or whether it might be masking a bug that a trap would've helped catch much earlier.

So instead, this PR just traps, both if attempting to start a sync operation on a waitable that's already in a waitable-set or, conversely, if trying to add a waitable to a waitable-set when there's already a sync operation in progress.  This is also a conservative choice, and we could switch to the above later if a compelling reason arose.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1339,14 +1339,23 @@ A waitable can belong to at most one "waitable set" (defined next) which is
 referred to by the `wset` field. A `Waitable`'s `pending_event` is delivered
 (via `get_pending_event`) when core wasm code waits on its waitable set (via
 `waitable-set.wait` or, when using `callback`, by returning to the event loop).
+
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):
To me this seems reasonable yeah in the sense that in higher-level situations when you're performing a future/stream operation you've probably already got ownership of the future/stream so you should be able to have complete   control over the state of the stream before/after so having extra requirement shouldn't be too onerous.

dicej (approved):
Thanks!  Yes, the seems like the right, conservative choice for now, and we can revisit once we have a relevant use case.

PR #650, opened 2026-05-13
Clarify and fix definition of "reentrance"

lukewagner:
Currently, the rules for reentrance are defined in terms of the async callstack.  However, as explained in more detail in `Concurrency.md#asynchronous-recursion` in this PR, the async callstack is a somewhat fuzzy concept that is tricky to define for host embeddings (like JS) and also doesn't really achieve its goal.  Instead, based on the stack-switching refactoring work in #637, this PR is able to redefine reentrance in terms of flag-flipping of component instances at suspend/resume points.  In doing so, the need for the overly-conservative `call_might_be_recursive` predicate (added in #589) goes away, and an optimization of similar syntactic complexity (described at the end of `CanonicalABI.md#component-instances` in this PR) can achieve the same effect while re-enabling "donut wrapping" (which this PR adds a definition and description for in `Linking.md#higher-order-shared-nothing-linking-aka-donut-wrapping`) and removing the need for a `recursive` future feature.

The downside is that, as described in `Concurrency.md#asynchronous-recursion`, certain cases of async recursive reentrance now deadlock instead of trapping.  However, as explained in that section, there's not really much we can do about that other than say "parent components and host embeddings have to be careful", which they already have to do anyways.

The PR also clarifies the rules for what the host is allowed to do, which was pretty handwavy with the callstack-based definition and thus open to misinterpretation.  The PR also clarifies and expands the `Explainer.md#component-invariants` section to more precisely nail down what the reentrance rules actually buy component authors.

Lastly, this PR paves the way for, and was motivated by work on, some forthcoming improvements to cooperative threads when used in a synchronous context.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -123,10 +123,11 @@ class ComponentInstance:
   parent: Optional[ComponentInstance]
   handles: Table[ResourceHandle | Waitable | WaitableSet | ErrorContext]
   threads: Table[Thread]
+  may_enter: bool
   may_leave: bool
   backpressure: int
   num_waiting_to_enter: int
-  exclusive: Optional[Task]
+  exclusive_thread: Optional[Thread]
 
   def __init__(self, store, parent = None):
     assert(parent is None or parent.store is store)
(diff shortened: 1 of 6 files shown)


alexcrichton (approved):
This all sounds reasonable to me, thanks for writing this all up!

PR #655, opened 2026-05-19
Clarify and regularize text format index parsing rules

### LINKED ISSUE #648: Why is it `(dtor (func n))` instead of `(dtor (core func n))`?
The text explainer for resource types says the following:

```
resourcetype  ::= (resource (rep i32) (dtor <funcidx>)?)
```

And the binary explainer says the following:

```
resourcetype  ::= 0x3f 0x7f f?:<funcidx>?      => (resource (rep i32) (dtor f)?)
```

A signature of `i32 => []` implies that the destructor should be a core func, not a component func. If so, I would expect both grammars to use `core:funcidx`, which is what you see for e.g. realloc: `(realloc <core:funcidx>)`, `0x04 f:<core:funcidx> => (realloc f)`.

That said, I am then further confused because the `realloc` examples in the explainer all use `(realloc (func $libc "realloc"))`. I assume this is an `inlinealias`, but is there really no distinction between a `func` and `core func` when performing an inline alias? If I understand correctly, an explicit alias would look like `(alias core export $libc "realloc" (core func))`, while the implicit alias apparently is `(func $libc "realloc")`. I don't understand this inconsistency.

Regardless of `(func` vs. `(core func` in the text format, though, should the formal grammars for `resourcetype` indeed be changed to `core:funcidx`?

lukewagner:
This PR intends to resolve #648 by regularizing and more-clearly defining the text format rules for core- and component-level indices, identifiers and inline aliases.  One nice thing with the changes in this PR is that there is now an explicit `core-prefix(...)` in the grammar at all the places where `core` might (when a plain `$id` is not used) get injected.

The `test/syntax/indices.wast` tries to test all the syntactic cases touched by this PR.  The commented-out lines currently fail in `wasm-tools` and the rest pass as-is.  This helps identify what the concrete delta from the status quo is.  I didn't add any negative tests since I expect we'll need a phase-out process before rejecting anything that is accepted now.

PTAL @bvisness @alexcrichton

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -7,8 +7,8 @@ more user-focused explanation, take a look at the
 
 * [Gated features](#gated-features)
 * [Grammar](#grammar)
+  * [Index spaces](#index-spaces)
   * [Component definitions](#component-definitions)
-    * [Index spaces](#index-spaces)
   * [Instance definitions](#instance-definitions)
   * [Alias definitions](#alias-definitions)
   * [Type definitions](#type-definitions)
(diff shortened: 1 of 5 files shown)


design/mvp/Explainer.md
```diff
@@ -76,17 +76,80 @@ succinctness, with just enough detail to write examples and define a [binary
...
+
+As with Core WebAssembly, for each index space `X`, there is an `{X}idx` grammar
+rule that parses indices as either a direct `u32` or a [`core:id`] that is
+resolved to a `u32` according to the usual Core WebAssembly text format rules.
+When the text format accepts indices in one of several different index spaces,
+the sort is required to be written explicitly using the `externidx` rule, which
+embeds an expanded version of [`core:externidx`] prefixed with `core`.
+```ebnf
+idx              ::= <u32> | <core:id>
+core:{Xᵢ}sortidx ::= (Xᵢ <idx>)                                     (for Xᵢ ∈ <core:sort>)
```
bvisness, line 130: I think this should be `(core Xᵢ <idx>)`, as the `Xᵢ` will be un-prefixed (e.g. `memory` alone).

design/mvp/Explainer.md
```diff
@@ -76,17 +76,80 @@ succinctness, with just enough detail to write examples and define a [binary
...
+core:{Xᵢ}idx     ::= <idx> | <core:{Xᵢ}sortidx>                     (for Xᵢ ∈ <core:sort>)
+core:externidx   ::= <core:{X₁}sortidx> | ... | <core:{Xₙ}sortidx>  (for X₁,…,Xₙ ∈ <core:sort>)
```
bvisness, line 132: Is this too permissive? It seems to include e.g. `(core module $foo)`, but the new rule for `core:inlineexport` is:

```
core:inlineexport   ::= (export <core:name> <core:externidx>)
```

That would seemingly imply that we could have a core module exporting a core module, which is not the case. I guess the problem is that `core:sort` includes not just things in core WebAssembly, but also core modules and core instances themselves.

design/mvp/Explainer.md
```diff
@@ -285,41 +273,26 @@ an example of these, we'll also need the `alias` definitions introduced in the
...
-rules to throw out the core sorts that aren't allowed in various contexts).
-
-The `name` production reuses the `core:name` quoted-string-literal syntax of
+inlineexport   ::= (export "<exportname>" <externidx>)
+                 | (export "<exportname>" <versionsuffix> <externidx>) 🔗
+```
+Components may import and export various core definitions when they are
+compatible with the [shared-nothing] model, which currently means only `module`
+(since modules are pure immutable code) but may in the future include, e.g.,
+`data` (since data segments are pure immutale data).
```
bvisness:
```suggestion
`data` (since data segments are pure immutable data).
```

design/mvp/Explainer.md
```diff
@@ -334,22 +307,65 @@ instance, the `core export` of a core module instance and a definition of an
 `outer` component (containing the current component):
 ```ebnf
 alias            ::= (alias <aliastarget> (<sort> <id>?))
-aliastarget      ::= export <instanceidx> <name>
-                   | core export <core:instanceidx> <core:name>
-                   | outer <u32> <u32>
+aliastarget      ::= export <idx> <name>
+                   | core export <idx> <core:name>
+                   | outer <idx> <idx>
```
bvisness: This seems unfortunate to me because it's no longer clear which kind of index is required here. I'm not sure if this is how it works today, but could it possibly take e.g. `<core:instanceidx>` instead of just `<idx>`? This would allow either a bare index or `(core instance $foo)`.

Alternatively you could maybe just define a little notation for what kind of bare index is required, just for reader comprehension.

design/mvp/Explainer.md
lukewagner: When a `core:{X}idx` is used inside a `(core ...)` block (e.g., `(core instance (instantiate $M (with "" (instance (export "mem" (memory $i "mem"))))))`), there is no `core` needed, so the approach used in this PR is that `core:{X}idx` rules always have no `core` prefix, and when you need one, the grammar has an explicit `core-prefix(<core:{X}idx>)`.

design/mvp/Explainer.md
lukewagner: So this is where the validation rules come in (in this and other cases) to winnow down what the grammar allows and throw out cases that aren't currently supported by core wasm (but hypothetically could in the future); I think this is mentioned below in the explainer.

design/mvp/Explainer.md
lukewagner: With the grammar rules as defined, a `<core:instanceidx>` would mean that you could use an inline alias inside an alias which seemed like maybe too inceptive.  Ultimately, the grammar bottoms out at a `<u32>` or `<id>`, and validation rules say "it has to be an instance in the instance index space", so that seemed kinda good enough, given that the second `<idx>` also needs special validation rules?

alexcrichton (approved):

design/mvp/Explainer.md
bvisness: Fair enough; I just hope I won't be confused about what kind of index is required in any particular context.

alexcrichton:
Mind skimming over and double-checking some of the test-related changes on https://github.com/bytecodealliance/wasm-tools/pull/2562?

keithw:
Seeing this via the PR in wasm-tools; sorry for the late comment. If the component-model text format is being regularized, is this a good time to clarify the ordering requirements for symbolic identifiers?

In core Wasm, the use of an identifier can come before the identifier is bound:
```wasm
(module
  (export "" (func $f))
  (func $f)
)
```

but this seems to be malformed in a component, in the sense that wasmparser's Parser rejects it (but I don't think there is explicit discussion of this in the Explainer yet or an `assert_malformed` test case):
```wasm
(component
  (export "" (core module $m))
  (core module $m)
)
```

To me the more vexing case is something like:

```wasm
(component
  (instance (export "" (instance 1)))
  (instance $i)
)
```

... which is accepted by wasmparser's Parser, but wasmprinter prints it out as:
```wasm
(component
  (instance (;0;)
    (export "" (instance $i))
  )
  (instance $i (;1;))
)
```

which is rejected by wasmparser the next time (both before and after https://github.com/bytecodealliance/wasm-tools/pull/2562).

tl;dr If attention is going into component WAT at the moment, is this the right time to lock down what is considered well-formed? This could help get to a place where all well-formed components successfully roundtrip through wasmparser->wasmprinter->wasmparser. (We got there for core modules in https://github.com/bytecodealliance/wasm-tools/pull/2148 .)

alexcrichton:
I'd consider both of those technically bugs in `wast`, wasm-tools's text-to-binary-stage. Components are core wasm are subtly different here where core wasm can sort of have everything reordered and items like functions, can refer to items defined after them. For components this isn't the case, nothing is reordered in the text format (it all goes through as-is to the binary format) and definitions cannot refer to future definitions.  This means that text-format-resolution for core wasm modules does a pass of "register all the identifiers" followed by "now go resolve what everything points to". For components though (and this is the buggy part) it instead only registers identifiers as their visited without this initial pre-pass. That's why the module with an integer index is parsed by the text parser and prints, but can't re-parse (as printing shares a lot of its infrastructure with core wasm and components).

Basically a long way of saying that I think this is just a bug in `wast`. It should visit all identifiers to assign indices first before actually resolving names. I think that would resolve all the components above. They'd still all be invalid components, but would round-trip through parsing.

lukewagner:
@alexcrichton All the WASTs in that PR look perfect.  Thanks!!

keithw:
@alexcrichton Oop, yes, you're right -- I guess the roundtrip pipeline here is more like wast->wasmparser->wasmprinter->wast.

I think maybe there are two different issues. On the syntax, if
```wasm
(component
 (import "a" (func (type $t)))
 (type $t (func))
)
```
is meant to be well-formed, then if the rules are getting clarified now, is this a good time to make this explicit in the prose + add some `assert_invalid` test cases (and as you say, `wast` will change to reflect this)?

In terms of validity, the Explainer right now says
> Components are like Core WebAssembly modules in that their contained definitions are acyclic: definitions can only refer to preceding definitions (in the AST, text format and binary format). However, unlike modules, components can arbitrarily interleave different kinds of definitions.

which makes it a *little* surprising that while this is valid
```wasm
(module
 (import "" "" (func (type 0)))
 (type (func))
)
```

this (syntactically similar) component is well-formed but invalid:

```wasm
(component
 (import "a" (func (type 0)))
 (type (func))
)
```

The current paragraph seems a little confusing because, unlike the component text format, the core text format *does* let users "interleave different kinds of definitions" except for the restriction on imports appearing first. So maybe this is just a question of clarifying the prose a bit.

lukewagner:
@keithw That's a great point; thanks!  Added [this blurb](https://github.com/WebAssembly/component-model/pull/655/commits/0a9375ca5202efb5e00427402479058674f7edc7).

lukewagner:
Looks like the new syntax parses/validates in Wasmtime (thanks @alexcrichton!) so I think it's time to merge.

Issue #651, opened 2026-05-14
Threads & non-async exports


badeend:
Two related questions:

---

The [Returning](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Concurrency.md#returning) section mentions:

> Any of the threads contained by a task can call task.return; there is no "main thread" of a task.

Except that in the sync ABI, there _is_ a "main thread"; the one that returns the return values, right?

---

The [Blocking](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Concurrency.md#blocking) section mentions:

> each of these potentially-blocking operations will trap if the current task's function type does not declare the async effect, since only async-typed functions are allowed to block.

Does this mean that threads spawned under a non-async task can never do blocking I/O? Seems like this restriction should only apply to the "main" thread of that task.

lukewagner:
> Except that in the sync ABI, there is a "main thread"; the one that returns the return values, right?

Good point, yes; the text should be nuanced to say only when one of the async ABIs is used.

> Does this mean that threads spawned under a non-async task can never do blocking I/O? Seems like this restriction should only apply to the "main" thread of that task.

You're right that that is currently a limitation with how the `may_block()` traps are defined.  Fortunately, for this, and some other use cases that @TartanLlama found working on pthreads, I've been working on a PR to relax this limitation.  The fix ended up spawning #637 and #650 as prerequisites, but now that those are merged/posted, the fix (in the `fix-coop` branch atm and I think already mostly implemented in Wasmtime) would allow coop threads to block at any time and there is only a trap if there are no "ready" threads left to switch to in the component-instance while the sync task has not yet returned its value, which is far more lenient.

lukewagner:
For the former comment, I added [this commit](https://github.com/WebAssembly/component-model/pull/646/commits/9586b4fbb6c0124f69c46d4c9d0c0232bd34bf7b) to #646; PTAL.

badeend:
> allow coop threads to block at any time and there is only a trap if there are no "ready" threads left to switch to

Ah, okay. That could work, yeah

PR #656, opened 2026-05-22
Fix some cooperative thread limitations

lukewagner:
@TartanLlama's work implementing pthreads in wasi-libc highlighted a few limitations with cooperative threads as currently spec'd that this PR intends to address:
* If pthreads are implemented using cooperative threads in sync-typed context, they won't work (due to trapping or no-op yielding) in cases where, if they had instead been implemented via CPS transform (like Asyncify), they could have.
* pthread APIs like `pthread_join`would like to be able to switch to a thread if that thread is ready to run, but not trap if it's not ready-to-run, b/c the current state of the thread isn't known.

The first bullet suggests replacing the strict "`may_block`" traps (and the no-op special case in `thread.yield`) with a more permissive approach where: if a thread blocks (now better-defined after #637) as part of a sync-typed call, but there are other threads in the same component instance that are ready-to-run, they are resumed directly, and repeatedly, until either there are no more ready threads (in which case trap, b/c it's actually blocked) or a value is returned successfully to the caller.  This also implicitly addresses the limitation @badeend just pointed out in #651, since blocking is fine as long as there is some other thread that can make progress.  (E.g., maybe the blocking call was just doing some asynchronous logging off the critical path.)  The "in the same component instance" caveat matches the expressivity of a core wasm CPS transform and is also necessary to maintain the reentrance invariant (re)defined in #650.

This does have the unfortunate side effect of allowing more code to accidentally depend on non-portable behavior (whether various operations actually block in practice), but this is already quite possible when using the `async` ABI and it's sortof just the nature of concurrency.

The second bullet suggests first renaming (but otherwise keeping the exact same behavior):
* `thread.switch-to` => `thread.suspend-then-resume` = resume other thread (trap if it's not "suspended"), leave current thread "suspended"
* `thread.yield-to` => `thread.yield-then-resume` = resume other thread (trap if it's not "suspended"), leave current thread "ready"

and then symmetrically add two new variants:
* `thread.suspend-then-promote` = resume other thread if it's "ready", leave current thread "suspended"
* `thread.yield-then-promote` = resume other thread if it's "ready", leave current thread "ready"

Thus, if the other thread is *not* "ready", `thread.{suspend,yield}-then-promote` behave the same as `thread.{suspend,yield}`.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -64,8 +64,10 @@ specified here.
   * [`canon thread.resume-later`](#-canon-threadresume-later) 🧵
   * [`canon thread.suspend`](#-canon-threadsuspend) 🧵
   * [`canon thread.yield`](#-canon-threadyield) 🧵
-  * [`canon thread.switch-to`](#-canon-threadswitch-to) 🧵
-  * [`canon thread.yield-to`](#-canon-threadyield-to) 🧵
+  * [`canon thread.suspend-then-resume`](#-canon-threadsuspend-then-resume) 🧵
+  * [`canon thread.yield-then-resume`](#-canon-threadyield-then-resume) 🧵
+  * [`canon thread.suspend-then-promote`](#-canon-threadsuspend-then-promote) 🧵
+  * [`canon thread.yield-then-promote`](#-canon-threadyield-then-promote) 🧵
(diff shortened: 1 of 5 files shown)


alexcrichton (approved):

lukewagner:
Since cooperative threads are gated and experimental, I'll merge this for now; happy to iterate on the design in future issues/PRs.

PR #657, opened 2026-05-22
Fix Explainer.md/Binary.md definition of outer aliases

lukewagner:
This PR intends to address the first issue in [this comment](https://github.com/WebAssembly/component-model/issues/621#issuecomment-4506120636), bringing the spec in line with the implementation (and logic).

(cc @zherczeg and @alexcrichton)

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -126,12 +126,14 @@ Notes:
 * For `export` aliases, `i` is validated to refer to an instance in the
   instance index space that exports `n` with the specified `sort`.
 * For `outer` aliases, `ct` is validated to be *less or equal than* the number
-  of enclosing components and `i` is validated to be a valid
-  index in the `sort` index space of the `i`th enclosing component (counting
-  outward, starting with `0` referring to the current component).
-* For `outer` aliases, validation restricts the `sort` to one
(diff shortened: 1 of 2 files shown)


zherczeg:
Thank you for the doc update!

zherczeg:
Btw, in some cases the explainer refers to "transitive" dependencies. In this case, not just resource types, but types which transitively refer to resources (e.g. a borrow) should be disallowed as well (if I understand this correctly).

E,.g.
```
   (type $IT (instance (export "sub-res" (type (sub resource)))))
   (import "any:any/anything@1.0.0" (instance $I (type $IT)))
   (alias export $I "sub-res" (type $SUB-RES))
   (type $BORROW (borrow $SUB-RES))
   (component $C (alias outer 1 $BORROW (type)))
   (export "comp" (component $C))
```

This is the error message of `wasm-tools`:
`error: cannot alias outer type which transitively refers to resources not defined in the current component (at offset 0x56)`

alexcrichton (approved):

lukewagner:
@zherczeg Great point!  [Added](https://github.com/WebAssembly/component-model/pull/657/commits/bbe35aa4235630fe26eaeecf064619e8fbaf644c)

zherczeg:
Thank you! LGTM

PR #672, opened 2026-06-29
Add `external-id` attribute

lukewagner:
This PR extends #613, which added an `(implements <interfacename>)` attribute to imports and exports, with an `(external-id <string>)` attribute.  The `external-id` attribute is useful when a component import or export name wants to match some external identifier, but the kebab-case syntax is too restrictive.  One example is ESM-integration, where Module Specifiers can (now) be arbitrary Unicode strings (including URLs).  Another is when importing stateful stores, e.g.:
```wit
world w {
  import redis: wasi:keyvalue/store;
  import memcached: wasi:keyvalue/store;
}
```
(as enabled by #613) and the platform-defined identifier of these stores isn't a valid `<plainname>`, so you'd either have to add a whole separate name-mapping scheme (e.g. mapping `redis` and `memcached` to their host IDs) or mangle the host IDs into the `<plainname>` (say, via URL-encoding; ew).  Instead, with this PR, you can use the `@external-id` attribute in WIT:
```wit
world w {
  @external-id("$whatever:host_id")
  import redis: wasi:keyvalue/store;

  @external-id("$something:else")
  import memcached: wasi:keyvalue/store;
}
```
and the WAT `(external-id <string>)` is plumbed into the host, alongside `(implements <interfacename>)`, so that the host can reflect on it and act accordingly.

As mentioned above, `@external-id` can subsume the use cases for `url=<...>` names for ESM-integration.  Moreover, the [reasoning](https://github.com/WebAssembly/component-model/pull/613#discussion_r2996644625) in #613 for moving `implements` out of the name and into a separate attribute I think similarly suggests that the `<urlname>`, `<depname>` and `<hashname>` cases of `<importname>` aren't really the right place to solve their associated use cases.  IIUC, these 3 name cases aren't currently being produced since there's no WIT syntax to emit them.  Thus, this PR tentatively proposes removing these 3 name cases.  The *use* cases are still important, but I think when we're ready to prioritize adding them to WIT/tooling, we should consider adding them as attributes instead.  In the meantime, it lets us merge `importname` and `exportname` into `externname`, which simplifies nicely things.

The PR also tidies up some sloppiness in the text-format grammar around string literals (e.g., `"foo-bar"`) vs. the contents inside the quotes (e.g. `foo-bar`), using `<Xstr>` for the former and `<X>` for the latter.

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -87,7 +87,7 @@ sort                ::= 0x00 cs:<core:sort>                                => co
                       | 0x03                                               => type
                       | 0x04                                               => component
                       | 0x05                                               => instance
-inlineexport        ::= n:<exportname'> si:<sortidx>                       => (export n si)
+inlineexport        ::= na:<nameattributes> si:<sortidx>                   => (export na si)
 ```
 Notes:
 * Reused Core binary rules: [`core:name`], (variable-length encoded) [`core:u32`]
@@ -101,8 +101,8 @@ Notes:
   for aliases (below).
 * Validation of `core:instantiatearg` initially only allows the `instance`
   sort, but would be extended to accept other sorts as core wasm is extended.
-* Validation of `instantiate` requires each `<importname>` in `c` to match a
-  `name` in a `with` argument (compared as strings) and for the types to
+* Validation of `instantiate` requires each `<externname>` in `c` to match a
+  `name` in a `with` argument (using plain string equality) and for the types to
   match.
 * When validating `instantiate`, after each individual type-import is supplied
   via `with`, the actual type supplied is immediately substituted for all uses
@@ -200,8 +200,8 @@ defvaltype    ::= pvt:<primvaltype>                       => pvt
                 | 0x70 t:<valtype>                        => (list t)
                 | 0x67 t:<valtype> len:<u32>              => (list t len)  (if len > 0) 🔧
                 | 0x6f t*:vec(<valtype>)                  => (tuple t+)    (if |t*| > 0)
-                | 0x6e l*:vec(<label'>)                   => (flags l+)    (if 0 < |l*| <= 32)
-                | 0x6d l*:vec(<label'>)                   => (enum l+)     (if |l*| > 0)
+                | 0x6e l*:vec(<labellit>)                 => (flags l+)    (if 0 < |l*| <= 32)
+                | 0x6d l*:vec(<labellit>)                 => (enum l+)     (if |l*| > 0)
                 | 0x6b t:<valtype>                        => (option t)
                 | 0x6a t?:<valtype>? u?:<valtype>?        => (result t? (error u)?)
                 | 0x69 i:<typeidx>                        => (own i)
@@ -209,9 +209,9 @@ defvaltype    ::= pvt:<primvaltype>                       => pvt
                 | 0x66 t?:<valtype>?                      => (stream t?) 🔀
                 | 0x65 t?:<valtype>?                      => (future t?) 🔀
                 | 0x63 k:<valtype> v:<valtype>            => (map k v) (if k is in <keytype>) 🗺️
-labelvaltype  ::= l:<label'> t:<valtype>                  => l t
-case          ::= l:<label'> t?:<valtype>? 0x00           => (case l t?)
-label'        ::= len:<u32> l:<label>                     => l    (if len = |l|)
+labelvaltype  ::= l:<labellit> t:<valtype>                => l t
+case          ::= l:<labellit> t?:<valtype>? 0x00         => (case l t?)
+labellit      ::= len:<u32> l:<label>                     => "l"  (if len = |l|)
 <T>?          ::= 0x00                                    =>
                 | 0x01 t:<T>                              => t
 valtype       ::= i:<typeidx>                             => i
@@ -230,8 +230,8 @@ instancedecl  ::= 0x00 t:<core:type>                      => t
                 | 0x01 t:<type>                           => t
                 | 0x02 a:<alias>                          => a
                 | 0x04 ed:<exportdecl>                    => ed
-importdecl    ::= in:<importname'> ed:<externdesc>        => (import in ed)
-exportdecl    ::= en:<exportname'> ed:<externdesc>        => (export en ed)
+importdecl    ::= na:<nameattributes> ed:<externdesc>     => (import na ed)
+exportdecl    ::= na:<nameattributes> ed:<externdesc>     => (export na ed)
 externdesc    ::= 0x00 0x11 i:<core:typeidx>              => (core module (type i))
                 | 0x01 i:<typeidx>                        => (func (type i))
                 | 0x02 b:<valuebound>                     => (value b) 🪙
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):

ricochet (approved):
This looks good to me. I'd use it right away. 

You may want to update Linking.md as well: 
> mutually known to later stages in the deployment pipeline (specifically with
  the [`depname`] case of `importname` in the text and binary format).

To something like:
>   mutually known to later stages in the deployment pipeline. (How an import
  name is resolved to a particular registry artifact is left to the host or
  build tooling; the optional [`external-id`] attribute can carry a
  tooling-specific identifier such as a registry path or URL.)

lukewagner:
@ricochet Thanks, good catch!  [Fixed](https://github.com/WebAssembly/component-model/commit/c6d4e726d2d26ee74bb3c09b700480016892dd9f)

design/mvp/WIT.md
```diff
@@ -1107,6 +1108,14 @@ sequence of digits:
 integer ::= [0-9]+
 ```
 
+### Strings
+
+WIT string literals are double-quoted and use the same text format as the Core
+WebAssembly text format's [`string`] literal. For example `"a"`, `"☃︎"`, `"\7f"`
+and `"\u{7fff}"` are all `string` literals.
+
+[`string`]: https://webassembly.github.io/spec/core/text/values.html#text-string
```
alexcrichton: Clarification on this: core wasm string literals are for encoding byte arrays (e.g. data segments), so could/should this additionally indicate that a utf-8 check is expected?

lukewagner: Oops, you're right; I should switch these `string`s back to [`name`](https://webassembly.github.io/spec/core/text/values.html#text-name) which, as you said, is the utf-8 decoding of `string`.

lukewagner: [Fixed](https://github.com/WebAssembly/component-model/pull/672/commits/13a795c8bf65ad655667241fb9b6c2c7b05962ba)

alexcrichton:
I've got an initial implementation of this over at https://github.com/bytecodealliance/wasm-tools/pull/2555 if you wouldn't mind [skimming over the test](https://github.com/bytecodealliance/wasm-tools/pull/2555/changes#diff-92671541383a44cc6a445aa1fcfb31e020cef4ab2ed5518e2fdf94304010f6ea) I've added

lukewagner:
Awesome!  Test looks good.

design/mvp/Binary.md
```diff
@@ -412,21 +412,26 @@ Notes:
...
   `plainname`s only occur on `func` imports or exports and that the first label
   of a `[constructor]`, `[method]` or `[static]` matches the `plainname` of a
   preceding `resource` import or export, respectively, in the same scope
   (component, component type or instance type).
 * 🏷️ Validation requires that `implements`-annotated imports or exports are
-  `instance`-typed.
-* 🏷️ Validation requires that `implements`-annotated imports or exports have a
-  `<plainname>` name.
+  `instance`-typed and have a `plainname` name.
+* 🏷️/🔗 Validation requires that a `vec(<attribute>)` contains each kind of
```
ricochet, line 431:
```suggestion
* 🏷️ Validation requires that `external-id`-annotated imports or exports are
  `instance`-typed.
* 🏷️/🔗 Validation requires that a `vec(<attribute>)` contains each kind of
```

design/mvp/Explainer.md
```diff
@@ -2549,35 +2546,102 @@ exported (e.g., `(export $x "x" (func $f))` binds a new identifier `$x`).
...
+This ensures that imports and exports continue to have unique names (so that,
+e.g., guest code executing inside this component has distinct `primary`-
+and `secondary`-scoped names for all `wasi:keyvalue/store` operations) while
+still providing the semantic information conveyed by the `interfacename` (so
+that, e.g., hosts can implement both imports with a standard static host
+`wasi:keyvalue/store` implementation that is dynamically parameterized by the
+store names `"primary"` and `"secondary"`).
+
+🏷️ The `external-id` attribute addresses the naming-expressivity gap left
+between the restricted syntax of `externname` (shown next) and whatever
```
ricochet, line 2650:
```suggestion
🏷️ Like `implements`, the `external-id` attribute may only be applied to
`instance`-typed imports and exports. It addresses the naming-expressivity gap
left between the restricted syntax of `externname` (shown next) and whatever
```

design/mvp/Binary.md
ricochet: Shouldn't external-id be like implements and be only on instances?

lukewagner: Great catch; I totally missed this discrepancy between the PR and @alexcrichton's above-linked test case.  So while `implements` is inherently tied to a WIT interface (via `interfacename`) which is inherently tied to an `instance` type, `external-id` has no such tie.  I think there are also real examples where you'd naturally want to put `@external-id` on other types of imports/exports (e.g., nullary functions returning capabilities) and indeed WIT.md already includes some (toy) examples that do this too.  So perhaps we could keep the spec as-is and change the above test?  WDYT @alexcrichton?

alexcrichton: Seems reasonable to me yeah, and sorry forgot to flag this here as well. This restriction came about when I was working on the WIT side of things and I was wondering where to store external-id in the AST of a WIT package. 

Would it be ok to limit this to just functions/interfaces for now? Or should types be included? Or perhaps components/etc as well? Or, I suppose another way to ask, should there be [any validation of `external_id`](https://github.com/bytecodealliance/wasm-tools/blob/a001d3c109656f6ed9b6dec325b1a5729584621d/crates/wasmparser/src/validator/component.rs#L4728-L4731) when it's encountered?

lukewagner: Since `@external-id`s are really just about closing the kebab-case-name-expressivity gap, I think the use cases for it on `instance`s and `func`s are about as strong as for having it on `component`s, `core module`s and `type`s.  E.g., with ESM-integration, it seems valuable for a component to be able to give a `core module`- or `component`-typed import a URL `external-id` so that these dependencies can be fetched by the ESM machinery without needing an import-map.  That leaves `type`; a concrete use case there might be importing a `resource` type that corresponds to some arbitrary JS identifier?  So my inclination is that there would be no validation criteria concerning the type (only that there is at most one `external-id` in the `attribute` list).  But let me know if that's too big of a lift for now.

alexcrichton: Ok that sounds reasonable to me yeah, I'll prep the requisite plumbing in WIT parsing to ensure that `@external-id` can basically show up anywhere.

Although, question on that, in WIT should `@external-id` be allowed within an `interface`? Or only within a `world` (e.g. on an `import` or `export`)

lukewagner: I think it'd be useful on items inside an `interface` too.  One use case might be declarative configuration imports: the `@external-id` of an `interface` could name the configuration store and the interface's fields could name the (non-kebab-cased) configuration keys (of type `func() -> value`).

lukewagner: Great, so I think that resolves this to keep things as-is, so marking this as resolved.

design/mvp/Explainer.md
lukewagner: Based on the other conversation, I think this is also resolved without change.

design/mvp/WIT.md
lukewagner: (FWIW, also [updated](https://github.com/WebAssembly/component-model/pull/672/commits/225cc9031fc0d81b846a55e71dbde691b04d9c99) the grammar rule names to have a `<*lit>` suffix, instead of `<*str>`, using "literal" as the term to mean "the thing surrounded with double quotes" as opposed to the unquoted inner string.)

alexcrichton:
Ok with https://github.com/bytecodealliance/wasm-tools/pull/2558 that shoudl, I believe, implement all of this and have everything round-tripped in all the places

lukewagner:
That's great!  I guess then I'll wait a bit to see if anyone else has comments and then merge, noting that the feature isn't yet released, so we can continue to iterate in subsequent issues/PRs.

PR #673, opened 2026-06-30
Clarify type import/export rules in new factored-out subsection

lukewagner:
Currently `Explainer.md` mentions the validation rule that imported/exported function types may only transitively refer to `resource` types that have themselves been imported/exported (thereby giving them an externally-visible name), but there is no mention of the symmetric requirement on the compound value types `record`, `variant`, `enum` and `flags` (which also need externally-visible names for the sake of bindings generators).  This PR intends to plug this spec hole (which has been pointed out a few times recently) to match the existing implementation, as tested by the new WAST test in the PR.  But @alexcrichton please LMK if anything doesn't match and then happy to iterate.


alexcrichton (approved):
Text/test look good to me, thanks!

PR #674, opened 2026-07-01
Adjust some documentation for threading intrinsics

alexcrichton:
* List a binary encoding for refactorings in #656 
* Update the feature for `thread.yield` to be 🔀 instead of 🧵


lukewagner:
Great catches, thanks!

PR #675, opened 2026-07-01
Clarify 'alias outer' rules, add test

lukewagner:
This PR is meant to address the question in [wasm-tools/#2545](https://github.com/bytecodealliance/wasm-tools/issues/2545).  Outer aliases were a bit hand-wavily defined before, so the PR tries to clarify things by ruling out more invalid cases in the grammar (which has been a source of confusion).  Also a WAST test with better coverage is added (which passes on current Wasmtime).

### DIFF
--- a/design/mvp/Binary.md
+++ b/design/mvp/Binary.md
@@ -116,24 +116,23 @@ Notes:
 
 (See [Alias Definitions](Explainer.md#alias-definitions) in the explainer.)
 ```ebnf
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):

alexcrichton:
FWIW the test doesn't currently pass in wasm-tools here, but that's for unrelated tooling issues where it looks like wasmprinter, when printing an invalid component, makes it invalid in a different way. Otherwise lgtm 👍

Issue #678, opened 2026-07-17
What context.{get,set} slots are used when an async task returns a string?


alexcrichton:
I've been thinking about https://github.com/bytecodealliance/wasmtime/issues/13890 and how best to fix it recently. I've come across behavior though where I'm not entirely sure what to do, hence the question here. The basic problem of this Wasmtime issue is that when a thread exits there remains lingering references to the thread's ID and data structure which causes problems when they're later accessed. My original goal was to continue to "promptly deallocate" thread structures when a thread exits (as opposed to keeping them alive until subtasks have returned or something like that), but this led to a bit of a deeper question.

Specifically there's currently the scenario where a host async function returns in a "delayed" fashion. For example it's invoked, not ready yet, and then a subtask is created within the guest. Eventually when this host subtask actually generates its result it lowers the result into guest memory and then signals an event saying that the subtask is ready. The process of lowering, however, can invoke guest functions like `realloc` and need `context.{get,set}` intrinsics to work correctly. At this time I don't actually know what the right values are to use here and I think Wasmtime is relatively buggy.

A rough example of what I'm thinking of is this test:

<details>

```wasm
(component
  (core module $libc
    (import "" "context.set" (func $context.get (result i32)))
    (memory (export "m") 1)
    (func (export "realloc") (param i32 i32 i32 i32) (result i32)
      (i32.store
        (i32.const 100)
        (call $context.get))
      i32.const 0
    )
  )
  (component $A
    (core func $context.get (canon context.get i32 0))
    (core instance $libc (instantiate $libc
      (with "" (instance
        (export "context.set" (func $context.get))
      ))
    ))

    (core module $m
      (import "libc" "m" (memory 1))
      (import "" "task.return" (func $task.return (param i32 i32)))

      (global $go (mut i32) (i32.const 0))

      (func (export "run") (result i32)
        i32.const 1 ;; CALLBACK_CODE_YIELD
      )

      (func (export "run-cb") (param i32 i32 i32) (result i32)
        (if (result i32) (global.get $go)
          (then
            (call $task.return (i32.const 0) (i32.const 0))
            i32.const 0 ;; CALLBACK_CODE_EXIT
          )
          (else (i32.const 1)) ;; CALLBACK_CODE_YIELD
        )
      )

      (func (export "go") (global.set $go (i32.const 1)))
    )

    (core func $task.return (canon task.return (result string) (memory $libc "m")))
    (core instance $m (instantiate $m
[...]

PR #680, opened 2026-07-22
Clarify behavior of context.{get,set} during CABI realloc

lukewagner:
This PR intends to address #678 which observes that `realloc` is underspecified w.r.t `context.{get,set}`.  The root (spec) problem is that the core `realloc` function is being called "directly" during lowering without going through the usual coop-threads machinery that defines the current thread.  To fix this, this PR puts calls to `realloc` behind new, slightly-more-convenient `LiftLowerContext.{allocate,reallocate}` methods that call `realloc` as-if it was an export (via `canon_lift`) which ends up creating a fresh "thread" for the `realloc` cal.  However, as described in the optimization paragraph in this PR, b/c it's a fully-sync call and b/c of the `may_leave` guards, an engine can unobservably use whatever thread it wants as long as it sets/restores the 2 thread-local-storage cells and, if necessary, the current instance.

There is one extra choice in this PR that could go either way and that is to add a `trap_if(not may_leave)` to `thread.index`, so that `realloc`'s thread index is unobservable, which I thought might simplify optimization. It's also just a weird omission in the original coop threads PR to allow `thread.index` but not any of the threading builtins that actually do something with an index that `thread.index` was added for.  Happy to discuss this though.


alexcrichton (commented):
Ok I've implemented this in Wasmtime and wasi-libc (linked above) and this looks good to me. The wrapper in wasi-libc wasn't too too hard to get right and none of the tests are currently failing (e.g. `free` never tries to get the current thread id). I'm not 100% confident this'll continue to slide through uneventfully once more interesting ABI situations are exercised in wit-bindgen for example, but I'm also pretty confident that should this situation arise it'd be best handled in wasi-libc. 

Basically, lgtm 👍

alexcrichton (approved):

alexcrichton:
One minor test-related note is that in https://github.com/bytecodealliance/wasmtime/pull/13951 I found that some of `post-return.wast` will need updating to expect traps for `thread.index` now

lukewagner:
Thanks @alexcrichton!  Updated `test/values/post-return.wast` in this PR and confirmed all tests now pass.

PR #688, opened 2026-08-09
Specify maximum static size for component value types

### LINKED ISSUE #682: Should types have a spec-mandated maximum size?
When reviewing the implementation of fixed-length-lists throughout Wasmtime/wasm-tools one of the primary things I've seen is the hazard of overflows that are possible when calculating the size of a fixed-length-list. This is the first component model type where the size of the value is variable depending on the definition (statically at least, as opposed to lists which are runtime-variable-sized). It's quite easy to just multiple the element size and list size together and ignore overflow, and this has led to a number of bugs or potential bugs in wasm-tools and Wasmtime. Personally I see this as a potential security hazard because it's quite easy for the host to hit these overflow cases.

I'd like to propose the possible idea of capping the size of all component model values in the specification itself. For example disallowing any fixed-length-list above 16M or something large-ish like that. This would apply to all preexisting types and additionally handle something like `(tuple $big-list $big-list)`. Currently in `wasm-tool` there's already a limit to "type complexity" of sorts where everything is counted in the recursive tree and the sum needs to be below a fixed amount. Capping the maximum size of the list, however, is probably a better heuristic.

Effectively what I'd like to do is to put the complicated overflow/ABI logic in one place, the component validator, and then not have to worry anywhere else about this. Right now Wasmtime has a few potential would-be-CVEs if fixed-length-lists were stable, and I feel that without having checks on the size of values it's too easy for runtimes to run into overflow situations.

ANAMASGARD:
Fixes #682 

## Summary

Adds a spec-defined maximum static value size of `2^28 - 1` bytes for Component Model value types.

Validation now:
- checks both `i32` and `i64` pointer types;
- uses overflow-safe size calculations;
- validates nested/despecialized types recursively;
- rejects oversized types during validation instead of at runtime.

Includes Python and WAST regression tests for fixed-length lists, compound types, nested maps, pointer-width differences, and integer-overflow cases.

The WAST test remains in `nyi.txt` until wasm-tools/Wasmtime implement the corresponding validator rule.


lukewagner:
Thanks for working on this!  To resolve #682, I had been thinking that perhaps we could reuse the `elem_size` function as already defined, and simply add validation notes mentioning it in all the relevant places: Explainer.md, Binary.md, and [`canonopt` validation](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#canonopt-validation).  That should then reduce the size of the PR.

ANAMASGARD:
@lukewagner   Thanks for the feedback 

Validation now reuses the existing elem_size definition instead of adding a separate checked_* algorithm. The full rule is in Element Size in CanonicalABI.md (every despecialized AST node, both i32 and i64), with short pointers in Explainer.md, Binary.md, and canonopt validation. WAST tests cover the edge cases; nyi.txt until wasm-tools/Wasmtime implement the validator.

@alexcrichton  please review this PR and give your feedback , thanks

alexcrichton:
This seems fine by me modulo precise spec text yeah, it's what I was imagining on #682 at a high level

design/mvp/CanonicalABI.md
```diff
@@ -3594,6 +3605,9 @@ performed for a component. These are defined as:
 * `lift(T)`
   * requires `realloc` if `T` contains a `list` or `string`
 
+Value types used by `lift`/`lower` are already rejected at `defvaltype`
+definition if they exceed the [Element Size](#element-size) bound, so lift and
+lower may assume static layouts fit.
```
lukewagner:
```suggestion
definition if they exceed the [Element Size](#element-size) bound.
```

design/mvp/Explainer.md
```diff
@@ -1040,6 +1040,15 @@ is a central part of validation and, e.g., occurs when validating that the
 `with` arguments of an [`instantiate`](#instance-definitions) expression are
 type-compatible with the `import`s of the component being instantiated.
 
+##### Maximum static value type size
```
lukewagner: Since this is a top-level requirement on `defvaltype`, could you put the following paragraph, without the heading, after the big quoted block that defines `defvaltype` (right before the "#### Fundamental value types" section heading).

design/mvp/Explainer.md
```diff
@@ -1040,6 +1040,15 @@ is a central part of validation and, e.g., occurs when validating that the
...
+
+Each `defvaltype` is rejected unless every node of its resolved structural AST
+(after despecialization) has `elem_size` at most `2^28 − 1` for both `i32` and
+`i64` pointer types. Checking only the root is not enough: wrappers such as
+`map`, `option`, `stream`, and `future` can have a small header while a nested
+node exceeds the bound. This is a static validation error. See
+[Element Size](CanonicalABI.md#element-size).
```
```suggestion
To prevent integer overflow in obscure corner cases, as an extra validation requirement, `defvaltype`s may not be equal or greater than `2<sup>28</sup>` bytes when serialized into linear memory according to the `i64` ABI definition of [Element Size](CanonicalABI.md#element-size).
```

design/mvp/CanonicalABI.md
```diff
@@ -2361,6 +2361,17 @@ def elem_size_flags(labels):
...
+Validation of each `defvaltype` requires that every node `u` in the resolved
+structural AST — after `despecialize`, including implicit nodes such as `map`'s
+key/value record — satisfies `elem_size(u, ptr_type) ≤ 2^28−1` for both
+`ptr_type ∈ {i32, i64}`. This is a static validation error, not a runtime trap.
+The bound is the same number as `MAX_LIST_BYTE_LENGTH` / `MAX_STRING_BYTE_LENGTH`.
+`elem_size` is unbounded-integer math; fixed-width implementations must reject
+overflow rather than wrap. Checking only the root type is not enough:
+`map<u8, list<u8, 2^28−1>>` has a small list header but an oversized pair record.
+Incremental or memoized checking is allowed only if every resolved node of that
+definition is still covered.
```
lukewagner:
```suggestion
To prevent integer overflow in obscure corner cases, component validation rules require that for every value type `t` defined by a component, `elem_size(t, 'i64')` is less than 2<sup>28</sup> (the same upper bound as `MAX_LIST_BYTE_LENGTH`).
```
Also, can you move this sentence to the first introductory paragraph of this section (before the quote block)?

design/mvp/Binary.md
```diff
@@ -287,6 +287,10 @@ Notes:
   `none` case of an optional immediate.)
 * 🔧 for fixed-sized lists the length of the list must be larger than 0 to pass
   validation.
+* Validation of each `defvaltype` requires every resolved AST node (after
+  despecialization) to satisfy the `elem_size` bound in
+  [Element Size](CanonicalABI.md#element-size). This is a static validation
+  error, not a runtime trap.
```
```suggestion
* Validation requires that, for every `defvaltype` `t`, `elem_size(t, 'i64')` is less than 2<sup>28</sup>, as defined by the [Canonical ABI](CanonicalABI.md#element-size).
```

design/mvp/Explainer.md
lukewagner: Also, could you put this paragraph, without the heading, after the big quoted block that defines `defvaltype` (right before the "#### Fundamental value types" section heading).

test/validation/max-value-size.wast
```diff
@@ -0,0 +1,92 @@
+;; Validation rejects defvaltypes whose resolved structural AST exceeds
+;; 2^28 - 1 bytes. See CanonicalABI.md#element-size.
+
+;; valid boundaries
```
lukewagner: Could you fold all these valid types into a single `(component ...)` and also add valid-up-to-the-boundary cases of compound types, analogous to all the invalid cases you added below?

lukewagner (changes requested):
Overall looks good to me; thanks!  A few requested changes and then I expect it'll be good to go.

ANAMASGARD:
@lukewagner & @alexcrichton Thank you so much for the review .
I have  applied the requested wording and placements : Element Size intro + Binary/canonopt notes, Explainer paragraph moved next to defvaltype. WAST now has one valid component (including just-under-bound compounds); assert_invalid covers types whose elem_size(t, i64) is not less than 2^28.

design/mvp/Binary.md
```diff
@@ -287,6 +287,7 @@ Notes:
   `none` case of an optional immediate.)
 * 🔧 for fixed-sized lists the length of the list must be larger than 0 to pass
   validation.
+* Validation requires that, for every `defvaltype` `t`, `elem_size(t, 'i64')` is less than 2<sup>28</sup>, as defined by the [Canonical ABI](CanonicalABI.md#element-size).
```
lukewagner: (can you line-wrap this please)

lukewagner (approved):
Looks good to me, thanks.

design/mvp/Binary.md
ANAMASGARD: Done

ANAMASGARD:
@lukewagner I have wrapped the Binary.md validation bullet as requested.

Issue #571, opened 2025-10-21
Upstream Wasmtime's async test suite


yoshuawuyts:
This issue is a sibling of https://github.com/WebAssembly/wasi-testsuite/issues/181. Wasmtime currently has its own set of component model async tests written in Rust. With an eye on portability, we should upstream these tests from Wasmtime into the component model repo. I've ported a first test in https://github.com/WebAssembly/component-model/pull/569, but we should make it a goal to eventually port all tests. If done right, runtimes like Jco would eventually be able to drop their dependence on Wasmtime's test suite for the component model, and instead be able to directly test against the component model repo. Thanks! 

## Tasks
_These are all the component model async test programs in Wasmtime, grouped by scenario. There are a current total of 65 tests. Each test refers to a specific test case, which itself may be composed of one or more test programs interacting with one another. When translating that to WAST, it should result in a single `.wast` file which captures both the scenario and test programs._

### backpressure

- [ ] **[async_backpressure_caller](https://github.com/bytecodealliance/wasmtime/blob/bd7b59dad0923097760eb73103582252fcd2f408/crates/misc/component-async-tests/tests/scenario/backpressure.rs#L12-L19)**: [`async_backpressure_caller`](https://github.com/bytecodealliance/wasmtime/blob/bd7b59dad0923097760eb73103582252fcd2f408/crates/test-programs/src/bin/async_backpressure_caller.rs), [`async_backpressure_callee`](https://github.com/bytecodealliance/wasmtime/blob/bd7b59dad0923097760eb73103582252fcd2f408/crates/test-programs/src/bin/async_backpressure_callee.rs)

### borrowing

- [ ] **[async_borrowing_callee](https://github.com/bytecodealliance/wasmtime/blob/bd7b59dad0923097760eb73103582252fcd2f408/crates/misc/component-async-tests/tests/scenario/borrowing.rs#L58-L64)**: [`async_borrowing_callee`](https://github.com/bytecodealliance/wasmtime/blob/bd7b59dad0923097760eb73103582252fcd2f408/crates/test-programs/src/bin/async_borrowing_callee.rs)
[...]

PR #689, opened 2026-08-09
Port Wasmtime's async_error_context test to WAST

ANAMASGARD:
## Summary
- Port Wasmtime's `async_error_context` scenario to `test/async/error-context.wast` as part of [#571](https://github.com/WebAssembly/component-model/issues/571)
- Single-component test exercises `error-context.new`, `error-context.debug-message`, and `error-context.drop` without asserting debug message contents (nondeterministic per spec)
- Document `-W component-model-error-context` in `test/README.md` for running error-context tests

## Test plan
- [x] `wasmtime wast -W component-model-error-context ... test/async/error-context.wast` passes locally (wasmtime 47.0.3)

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -6,12 +6,12 @@ This directory contains Component Model reference tests, grouped by functionalit
 
 A single `.wast` test can be run with full backtrace on trap via:
 ```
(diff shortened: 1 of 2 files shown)


lukewagner:
Thanks for the consideration, but until we get closer to enabling `error-context` and releasing it in a 0.3 dot-release (my impression is: as part of Lazy Lowering, so that we have an opt-in `canonopt` that lets us put `error-context` into the error payload of every `result` so as to enable idiomatic language error bindings), I'd like to avoid landing tests, since this will make work for other implementations trying to run the test suite that we may break later.

lukewagner:
I'll tentatively close, but we can reopen if there's more discussion.

PR #696, opened 2026-08-12
Fix CABI definition of coop thread promotability and add tests

lukewagner:
Currently, there is some divergence between Wasmtime and the spec for which coop threads can be switched-to during a synchronous call (when one thread blocks but others can be run).  Digging into it, I realized that the spec is buggy (at ensuring the stated [Component Invariant #3](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#component-invariants)), so this PR fixes the spec to match what Wasmtime is already correctly doing.  However, testing revealed other cases where I expect it's Wasmtime that we want to change to match the (fixed) spec.

(The WASTs added here assume #687 and its WAST tests pass (regarding `thread.*-then-promote`) and also probably dovetail with the WASTs merged in #691 discussed in [wasmtime/#14117](https://github.com/bytecodealliance/wasmtime/issues/14117), so cc @alexcrichton and @dicej)


dicej (approved):

test/async/trap-if-block-and-sync.wast
```diff
@@ -322,10 +309,9 @@
...
-(component instance $i $Tester)
-(assert_trap (invoke "trap-if-sync-future-cancel-read") "cannot block a synchronous task before returning")
-(component instance $i $Tester)
-(assert_trap (invoke "trap-if-sync-future-cancel-write") "cannot block a synchronous task before returning")
+
+;; TODO: also test sync calls to `{stream,future}.cancel-{read,write}`.  As of
+;; this writing, such calls can only ever block if the host has the other end,
+;; meaning we can't currently test that case in pure WAST; we'd need custom host
+;; code.  In the future, additional intrinsics (e.g. `stream.read-complete`) may
+;; enable us to test those cases in pure WAST.
```
alexcrichton, line 317: For this Wasmtime has a psuedo-host-world of sorts which is intended to be simple-to-define here -- https://github.com/bytecodealliance/wasmtime/blob/bc2f967927f9f0e839095a752bf554c4a09fdd13/crates/wast/src/spectest.rs#L91. I think it'd be reasonable to have a `world spectest { ... }` or similar with some canonical WIT that has comments for how the functions are expected to be implemented, and that could exercise some host-related functionality too without too much of an increase in complexity on the host side.

alexcrichton (approved):

lukewagner:
Tentatively closing this PR assuming we go with #705 instead.

PR #697, opened 2026-08-13
Fill in missing case of implicit backpressure when donut-wrapping

lukewagner:
Auditing the validity of [Component Invariant #3](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#component-invariants) surfaced that this invariant can be broken in [donut-wrapping](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Linking.md#higher-order-shared-nothing-linking-aka-donut-wrapping) scenarios.  (IIUC, donut wrapping is not-yet-implemented (and only recently re-enabled at the spec-level in #650), so this doesn't affect existing implementations.) The scenario is a parent-to-child-to-parent call where the outer parent call is expecting LIFO sync execution but the inner parent call is async-typed (which can block and break LIFO) and the solution is simply to exert backpressure at the attempted inner parent call.  The PR has more details.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -126,6 +126,7 @@ class ComponentInstance:
   threads: Table[Thread]
   may_enter: bool
   may_leave: bool
+  sync_depth: int
   backpressure: int
   num_waiting_to_enter: int
   exclusive_thread: Optional[Thread]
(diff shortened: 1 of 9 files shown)


lukewagner:
Tentatively closing this PR assuming we go with #705 instead.

Issue #694, opened 2026-08-12
WIT: allow type declarations at package scope


yordis:
A type can only be declared inside an `interface` or `world` today, and those are also the units of import and export, so sharing a type vocabulary means making an interface to hold it, and that interface reaches the artifact:

```wit
package t:demo@0.1.0;

interface types {
  record point { x: u32, y: u32 }
}

interface api {
  use types.{point};
  move-to: func(p: point);
}

world w {
  export api;
}
```

```console
$ wasm-tools component embed --dummy demo.wit --world w -o dummy.wasm
$ wasm-tools component new dummy.wasm -o c.wasm
$ wasm-tools component wit c.wasm
world root {
  import t:demo/types@0.1.0;

  export t:demo/api@0.1.0;
}
```

Nothing calls `t:demo/types`. It is imported because it is the naming container for a record. [#402](https://github.com/WebAssembly/component-model/pull/402) would make that more common: once a `use`d interface has to already be imported, every consuming world writes `import types;` for something nobody calls, and that PR notes `wasi:http/imports` would need exactly that addition.

[bytecodealliance/wasm-tools#1195](https://github.com/bytecodealliance/wasm-tools/issues/1195) has been open since 2023 out of wasmCloud composition, where the composed world came out carrying `import wasmcloud:messaging/types@0.1.0` with the reporter's inline note "this shouldn't be here since it is only types". The fix proposed there was a composer-level workaround and it never landed. [#629](https://github.com/WebAssembly/component-model/issues/629)'s motivating example is the same `interface types { ... }` shape, and [#295](https://github.com/WebAssembly/component-model/issues/295) is a newcomer stuck on what to even call the container, noting that a bare function can go directly on a world but "that doesn't work for modules that export enums, records, or resources".

My suggestion is to allow `typedef-item` at package scope, which is one production in WIT.md:

```ebnf
package-items ::= toplevel-use-item | interface-item | world-item | gate typedef-item
```

reusing `gate` and `typedef-item` unchanged, so package-scope types are feature-gateable the way interface items are. Every form is a parse error there today (`wasm-tools 1.255.0`: `expected world, interface or use, found keyword type`). I would leave `resource` out of the first cut, since alexcrichton's note on [wasm-tools#1195](https://github.com/bytecodealliance/wasm-tools/issues/1195) is that resource types still have to be imported regardless because they carry identity, which makes `resource` the one form where the container is doing real work.

The same package without the container:

```wit
package t:demo@0.1.0;

[...]

Issue #695, opened 2026-08-12
WIT: declare as a package-scope type, apply with `@annotate`


yordis:
WIT has three `@` attribute productions today — `gate` (`@since`/`@unstable`/`@deprecated`), `@implements`, and `@external-id`

Each with its own placement rules, and two of them landed this year.

Several more are queued in other issues: user-defined gate strings (#388), `@non-exhaustive` (#454), `@get`/`@set` (#235), and `shared`/`async`/`reentrant` as function attributes. So: should the next one be general?

**An annotation is a package-scope type, and applying it is a WAVE value of that type.**

Declaring:

```wit
package tooling:rust@1.0.0;

@annotation
type derive = list<string>;

@annotation
type rename = string;
```

Applying:

```wit
@annotate(tooling:rust/derive, ["Hash", "Eq"])
record object-id {
  @annotate(tooling:rust/rename, "objectID")
  id: string,
}
```

**Why a type instead of a bespoke parameter list?** The obvious alternative is to give `@annotation` a signature like a function's — `@annotation derive(names: list<string>)` — but that's a second, weaker type system that you then have to grow. As soon as an annotation needs "exactly one of these six", a parameter list can only offer six optional fields and a comment asking you to set one. That's a `variant`, and a `variant` is a type. Make the payload an ordinary WIT type and annotations inherit the whole type system, including whatever gets added to it later.

**Why package-qualified?** Bare `@word` stays the platform's. Users never take a bare name, so everything in the queue above can still land as a built-in later without colliding with anything anyone wrote. Collisions between user annotations become impossible, and versioning falls out of ordinary package deps rather than a registry.

This leans on two things in flight: #639 for WAVE literals, and #694 for a place to put package-scope types. String-only payloads would work with today's grammar, since `@external-id` already takes a string literal.

**What would help most** is use cases from people shipping WIT: which annotations you'd declare, where you'd attach them, and whether anything downstream needs to read them at runtime rather than at build time.

---

<details>
<summary><b>Why now: the built-in attribute queue</b></summary>

This picks up #58, which ends with an invitation to bring something concrete.

[...]

PR #699, opened 2026-08-15
Allow type declarations at package scope

yordis:
This PR allows `type`, `record`, `variant`, `enum` and `flags` declarations at package scope in WIT, as proposed in #694. Today, sharing a type vocabulary across interfaces requires inventing an interface to hold it, and that container reaches the artifact as an instance import that nothing calls.

`resource` is deliberately excluded. Resources carry identity rather than structure, so an interface is what gives a resource its identity, and package scope has nothing to offer there. That is why the grammar names a `package-typedef-item` rather than reusing `typedef-item`, which already includes `resource-item`.

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -1182,8 +1182,7 @@ the enclosing `interface` is what gives that import an identity for a consumer
 to name. The types above are structural, so a referencing `interface` can depend
 on one directly, without an `interface` having to be invented to hold it and
 without an instance having to be supplied to satisfy the dependency (see
-[Package Format][package-format]). The same reasoning is why outer aliases are
-also limited to structural types.
+[Package Format][package-format]).
 
 ### Feature Gates
 
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -1134,9 +1134,56 @@ wit-file ::= (package-decl ';')? (package-items | nested-package-definition)*
 nested-package-definition ::= package-decl '{' package-items* '}'
 
 package-items ::= toplevel-use-item | interface-item | world-item
+                | gate package-typedef-item
+
+package-typedef-item ::= variant-items
+                       | record-item
+                       | flags-items
+                       | enum-items
+                       | type-item
 ```
 
-Essentially, these top level items are [worlds], [interfaces], [use statements][use] and other package definitions.
+Essentially, these top level items are [worlds], [interfaces], [use statements][use], [package-scope types](#package-scope-types) and other package definitions.
+
+### Package-scope Types
+
+A `variant`, `record`, `flags`, `enum` or `type` may be declared directly at
+package scope, without an enclosing `interface` or `world`. A package-scope type
+is in scope for every `interface` and `world` in the package, with no `use`
+required, since there is no container to path through:
+
+```wit
+package local:demo;
+
+record point { x: u32, y: u32 }
+
+enum direction { north, south, east, west }
+
+interface api {
+  move-to: func(p: point);
+  heading: func() -> direction;
+}
+
+world w {
+  export api;
+}
+```
+
+Package-scope types share the one name space that all top-level definitions are
+exported into, so a package-scope `point` conflicts with an `interface point` or
+a `world point` in the same package. That shared name space is also what lets a
+`namespace:package/name` path address a package-scope type from another package,
+with no new syntax: `use local:demo/point;`.
+
+`resource` is deliberately excluded from `package-typedef-item`, which is why it
+names its alternatives rather than reusing `typedef-item`. Resource types carry
+identity rather than structure, so sharing one requires an abstract import, and
+the enclosing `interface` is what gives that import an identity for a consumer
+to name. The types above are structural, so a referencing `interface` can depend
+on one directly, without an `interface` having to be invented to hold it and
+without an instance having to be supplied to satisfy the dependency (see
+[Package Format][package-format]). The same reasoning is why outer aliases are
+also limited to structural types.
 
 ### Feature Gates
 
@@ -1413,6 +1460,12 @@ use-path ::= id
 Here `use-path` is an [interface name]. The bare form `id`
 refers to interfaces defined within the current package, and the full form
 refers to interfaces in package dependencies.
+
+A `use-path` may also name a [package-scope type](#package-scope-types), which
+brings the type into scope under its own name, or under the `as` name when one
+is given. This is the only way to name a package-scope type from another
+package, since the `use-path '.' '{' ... '}'` form of [`use`](#item-use)
+projects names out of an `interface` and a type has no such names to project.
 
 The `as` syntax can be optionally used to specify a name that should be assigned
 to the interface. Otherwise the name is inferred from `use-path`.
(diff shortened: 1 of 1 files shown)


design/mvp/WIT.md
```diff
@@ -2093,6 +2138,39 @@ This example illustrates the basic structure of interfaces:
...
+(component
+  (type (export "point") (record (field "x" u32) (field "y" u32)))
+  (type (export "api") (component
+    (export "local:demo/api" (instance
+      (type $point (record (field "x" u32) (field "y" u32)))
+      (export "move-to" (func (param "p" $point)))
+    ))
+  ))
+)
+```
```
alexcrichton, line 2175: For myself this is the main point of question for me of how these top-level types would be encoded in WIT. This component as-is is not valid (doesn't pass `wasm-tools validate`), and I assume that you don't want to change the validation rules for components, so could this be updated with an encoding that's valid?

lukewagner: Great question!  Here's one idea I think might work:

For WIT `interface`s like the above, we treat these dependencies on package-level types just like `use` of an `interface` that contains the analogous `type` definition, we just strip out the wrapping `instance` type:
```wat
(component $C
  (type $Point (export "point") (record (field "x" u32) (field "y" u32)))
  (type (export "api") (component
    (import "local:demo/point" (type $Point' (eq $Point)))
    (export "local:demo/api" (instance
      (export "move-to" (func (param "p" $Point')))
    ))
  ))
)
```
and then for the analogous WIT `world` (that exports `move-to`), we move the `(import "local:demo/point" (type ...))` inside the inner `component` type so that it works like a `type` definition inside an imported `interface`, just not wrapped in an `instance` type:
```wat
(component $C
  (type $Point (export "point") (record (field "x" u32) (field "y" u32)))
  (type (export "api") (component
    (export "local:demo/api" (component
      (import "local:demo/point" (type $Point' (eq $Point)))
      (export "move-to" (func (param "p" $Point')))
    ))
  ))
)
```

yordis: Thanks, this is much better than what I had. Pushed in 15d772e.

One deviation from your snippet worth flagging. You wrote `(eq $Point)` against the outer package-scope definition, which validates, but it reaches outside the wrapping component-type and so gives up the "each top-level definition is self-contained and valid (according to Component Model validation rules) independent of each other definition" property stated a little further down in the same section. The `use` encoding keeps that property by restating the imported type inside the wrapper, so I did the same here:

```wat
(component
  (type (export "point") (record (field "x" u32) (field "y" u32)))
  (type (export "api") (component
    (type $point (record (field "x" u32) (field "y" u32)))
    (import "local:demo/point" (type $point' (eq $point)))
    (export "local:demo/api" (instance
      (export "move-to" (func (param "p" $point')))
    ))
  ))
)
```

Since value types are structural these are the same type, but let me know if you meant the outer alias.

For the `world` case I put your second snippet next to the existing paragraph about a world's instance-type copies being duplicated for self-containment, since it is the same rationale.

@alexcrichton two notes on decoding, in case it changes your read on cost. Both the previous encoding and this one fail `wasm-tools component wit` today, but for different reasons. The top-level `(export "point" (type ...))` gives `component export 'point' was not a function or instance`, which is inherent to the feature either way. This encoding additionally gives `import 'local:demo/point' is not an instance`, since the decoder assumes an import with an interface name is an instance. So it looks like two changes rather than one, though the second one is what buys the unambiguous decode: the import appears if and only if the source referenced the package-scope type, and it carries the defining package and version, which a bare type export in the instance did not.

Still open, and neither of us has covered it: a package-scope type that references another package-scope type.

```wit
record point { x: u32, y: u32 }
type point-list = list<point>;
```

Package-scope types encode as bare type exports with no wrapping component-type, so there is nowhere to hang the import. Either they get a wrapper as well, or they reference each other directly and give up self-containment for that case. Do you have a preference? I did not want to invent a rule for it in this PR.

design/mvp/WIT.md
```diff
@@ -1134,9 +1134,56 @@ wit-file ::= (package-decl ';')? (package-items | nested-package-definition)*
...
+
+`resource` is deliberately excluded from `package-typedef-item`, which is why it
+names its alternatives rather than reusing `typedef-item`. Resource types carry
+identity rather than structure, so sharing one requires an abstract import, and
+the enclosing `interface` is what gives that import an identity for a consumer
+to name. The types above are structural, so a referencing `interface` can depend
+on one directly, without an `interface` having to be invented to hold it and
+without an instance having to be supplied to satisfy the dependency (see
+[Package Format][package-format]). The same reasoning is why outer aliases are
+also limited to structural types.
```
```suggestion
[Package Format][package-format]).
```
(in WIT.md the reader is not expected to know what an outer alias is.)

lukewagner (approved):
This all looks reasonable to me.  I'll plan to leave this open until we get some further implementation feedback.

alexcrichton:
One point I'm personally a bit murky on: how is this expected to handle merging across packages? The componentization process routinely merges `world`s together and various WIT packages, and that works out well today because conventionally everyone picks a unique name for their package and has their own version tracks. With this sort of global namespace, however, that breaks this merging property because now every package must agree on that `record point` has an `x` and a `y` field, and not a `z` field (or similar). 

I could put this a different way which is that I don't feel I fully understand the rationale for adding this as a feature. I feel like this breaks encapsulation of type definitions within WIT packagse and makes something that should be local a global decision. Increasing the risk of not being able to merge world together gives rise to the problem of bindings work in isolation and don't work together which is a major composability hazard.

Additionally this feels a bit weird to allow some types but not others. For example resources aren't allowed, and while I understand the rationale for not including them it means that this is a sort of "wart" where pieces don't compose together.

yordis:
> Additionally this feels a bit weird to allow some types but not others. 

Right now, I only need the types for the sake of https://github.com/WebAssembly/component-model/issues/695 and since the spec can be additive, I am personally Ok focusing on types right now.

Being said, better wait for @lukewagner since he has more context outside the Annotations as well.

lukewagner:
> With this sort of global namespace, ...

There might be some confusing wording that says "global" in the PR to be clarified, but what I believe is being proposed here (and shown in the WIT-to-WAT encodings) is using the existing package-scoped namespace (wherein any two `interface` and `world` names already have to be unique; this just adding a 3rd kind of "thing").  So I think it should fit right in to how we currently merge packages.

As I mentioned in the [design issue](https://github.com/WebAssembly/component-model/issues/694#issuecomment-5283833705), I think this is a logical feature in isolation but, more importantly, if we want to do "annotations" in the style of #695 (big if!  but maybe let's discuss that there) this feature is a natural basis.

alexcrichton:
If I've got two packages:

```wit
package a:b1 {
  record r { a: u32 }

  world w1 {
      export f1: func() -> r;
  }
}

package a:b2 {
  record r { a: f32 }

  world w2 {
      export f2: func() -> r;
  }
}
```

and I were to merge those two `world`s together, what would the resulting component type be? I suppose when I say "global" I mean for the world-merging process here which isn't quite the same as the package-merging process. Merging packages is easy enough since everything is scoped under a package, but merging worlds is where semantic knowledge comes in (e.g. interfaces are injected, union'd, dedup'd by semver, etc).

yordis:
Merging those two worlds works, and the two rs stay distinct. A package-scope type is keyed by its owning package, not by a global name, so nothing gets unified.

Running your example through `merge_worlds` and encoding the result gives:

```wat
(component
  (type (;0;) (record (field "a" u32)))
  (export (;1;) "r" (type 0))
  (type (;2;)
    (component
      (type (;0;)
        (component
          (type (;0;) (record (field "a" u32)))
          (import "a:b1/r" (type (;1;) (eq 0)))
          (type (;2;) (func (result 1)))
          (export (;0;) "f1" (func (type 2)))
          (type (;3;) (record (field "a" f32)))
          (import "a:b2/r" (type (;4;) (eq 3)))
          (type (;5;) (func (result 4)))
          (export (;1;) "f2" (func (type 5)))
        )
      )
      (export (;0;) "a:b1/w1" (component (type 0)))
    )
  )
  (export (;3;) "w1" (type 2))
)
```

alexcrichton:
I see, ok yeah. So this is sort of an implementation of "nested names" (🪺) in a sense, but with a mapping into WIT for at least one of the layers of nesting.

PR #703, opened 2026-08-18
Make strongly-unique rules transitive (#702)

### LINKED ISSUE #702: Strong uniqueness of names is not transitive
The current rules for strong uniqueness of names are quite annoying to implement, and hard to understand, because strong uniqueness is not transitive.

For reference, the current rules are:

1. If one name is `l` and the other name is `[constructor]l` (for the same
  `label` `l`), they *are* strongly-unique.
2. If one name is `l` and the other name is `[*]l.l` (for the same
  `label` `l` and any annotation `*` with a dotted `l.l` name), they *are
  not* strongly-unique.
3. Otherwise:
    * Lowercase all the `acronym`s (uppercase letters) in both names.
    * Strip any `[...]` annotation prefix from both names.
    * The names are strongly-unique if the resulting strings are unequal.

For the sake of this discussion, we will use != to mean "strongly-unique" and = to mean "not strongly-unique".

Under the current rules, `foo` = `FOO` (rule 3), and `FOO` = `[constructor]foo` (rule 3), but `foo` != `[constructor]foo` (rule 1). Similarly, `foo` = `FOO`, and `foo` = `[method]foo.foo`, but `FOO` != `[method]foo.foo`.

Image

Not only is this confusing, but it makes strong uniqueness harder to implement correctly. We would like to simply "canonicalize" each name and insert it into a hash map, but this lack of transitivity prevents us from doing so.

The fix for all this seems easy: make rules 1 and 2 case-insensitive. Is there any reason not to do so? It would make invalid some valid cases today, but is anyone actually depending on `foo` and `[method]foo.FOO` being unique? And it would be nice to define the entire strong-uniqueness algorithm in terms of canonicalized strings.

(For some history, as far as I can find it: In #197, @lukewagner suggested making name uniqueness case-insensitive _and_ potentially also hyphen-insensitive. He then implemented strong-uniqueness in #442 with the current carve-out for constructors, and it was subsequently modified in #494 and #499 to include the `[*]foo.foo` case. I'm really not sure if the case-sensitivity of rules 1 and 2 was deliberate -- although there is an explicit example to that effect (`foo-bar` vs. `[constructor]foo-BAR`), I truly cannot think of a reason to allow this because the non-uniqueness of `foo-bar` and `foo-BAR` will prevent you from exporting two such resource types in the first place.)

---

I also have some thoughts related to strong-uniqueness of (proposed) getters and setters, but I will discuss those in #235.

bvisness:
Makes the `[constructor]foo` and `[method]foo.foo` cases case-insensitive in order to make strong uniqueness a proper equivalence relation. This also enables implementations to implement strong-uniqueness checks more easily by literally canonicalizing strings in the manner described.

Resolves #702.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -2832,21 +2832,22 @@ within the same scope.
 
 To determine whether two names (defined as sequences of [Unicode Scalar
 Values]) are **strongly-unique**:
-* If one name is `l` and the other name is `[constructor]l` (for the same
-  `label` `l`), they *are* strongly-unique.
-* If one name is `l` and the other name is `[*]l.l` (for the same
-  `label` `l` and any annotation `*` with a dotted `l.l` name), they *are
-  not* strongly-unique.
-* Otherwise:
-  * Lowercase all the `acronym`s (uppercase letters) in both names.
-  * Strip any `[...]` annotation prefix from both names.
-  * The names are strongly-unique if the resulting strings are unequal.
+1. Canonicalize each name:
+  1. Lowercase all the `acronym`s (uppercase letters) in the name.
+  2. If the name is `[method]l.l` or `[static]l.l` for some `label` `l`, replace
+    the name with `l` (e.g. `[method]foo.foo` becomes `foo`).
+  3. If the name has any `[...]` annotation prefix other than `[constructor]`,
+    strip it from the name.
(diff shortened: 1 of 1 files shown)


lukewagner (approved):
Looks good, thanks!

Issue #197, opened 2023-05-17
Why `create-2d` is an invalid identifier?


ghost:
(Copy from https://github.com/bytecodealliance/wit-bindgen/issues/578)

I am declaring a function that creates a two-dimensional texture:

```
default interface client-experimental-texture {
    create-2d: func() -> string
}
```

This fails with:

```
  Caused by:
      invalid character in identifier '2'
           --> wit/client-experimental-texture.wit:2:5
            |
          2 |     create-2d: func() -> string
            |     ^', crates/wasm/build.rs:49:58
```

Kylebrown9:
From the component model's perspective, the question here is
> should `create-2d` be a legal kebab-case identifier?

The grammar as defined in `Binary.md` is as follows and only allows numbers **after** the first letter of each word.
```
label               ::= w:<word>                                           => w
                      | l:<label> '-' w:<word>                             => l-w
word                ::= w:[0x61-0x7a] x*:[0x30-0x39,0x61-0x7a]*            => char(w)char(x)*
                      | W:[0x41-0x5a] X*:[0x30-0x39,0x41-0x5a]*            => char(W)char(X)*
```

In order to make identifiers like `create-2d` legal without allowing identifiers to begin with numbers, we would need to add a different production for the first word than for subsequent words.

tschneidereit:
> In order to make identifiers like `create-2d` legal without allowing identifiers to begin with numbers, we would need to add a different production for the first word than for subsequent words.

A reason not to change anything here is that it'd not be clear how to generate unambiguous bindings for many languages. For example, if the WIT contains both `create-2d` and `create2d`, what should the generated names look like in a language like JS or Java, which camel-cases identifiers? The most natural answer for both would be `create2d` (or `Create2d` if it's a class name), and I really can't see any good alternatives :/

lukewagner:
I think it's worth considering relaxing the grammar of kebab-names to allow leading numbers in `word`s after the first as Kyle suggested.  This would also be useful if we wanted to better-support version numbers, as suggested in #134.

But Till has a good point.  Separately, we've talked about requiring kebab names to be case-insensitively unique (so that, e.g., you can't have both `create-XML` and `create-xml`, as this would conflict in a casing scheme that mapped both to `createXml`).  It might be a good idea to further tighten that requirement and require kebab names to be *hyphen-insensitively* unique (so that you can't have both `ab` and `a-b`), which may be independently valuable even outside of Till's example above).

[...]

PR #704, opened 2026-08-20
Make strongly-unique hyphen-insensitive

bvisness:
In #197 @lukewagner discussed making the strongly-unique algorithm also hyphen-insensitive. There was also some discussion of how strong this needed to be, but in my opinion it makes the most sense to remove hyphens entirely when checking for uniqueness, as this completely prevents names that differ only in case.

This is already a concern for us in Firefox, as an identifier like `[method]a-b.AB` on resource type `a-b` will produce a class named `AB` (PascalCase), and a method named `AB` (theoretically camelCase, but acronym). That's maybe workable for us as far as JS goes (because the constructor is just named `constructor`), but obviously other languages like Java and C# would choke on this case as well. I don't see a good reason for component interfaces to require case sensitivity.

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -2833,7 +2833,8 @@ within the same scope.
 To determine whether two names (defined as sequences of [Unicode Scalar
 Values]) are **strongly-unique**:
 1. Canonicalize each name:
-  1. Lowercase all the `acronym`s (uppercase letters) in the name.
+  1. Remove all hyphens and lowercase all the `acronym`s (uppercase letters) in
+    the name.
   2. If the name is `[method]l.l` or `[static]l.l` for some `label` `l`, replace
     the name with `l` (e.g. `[method]foo.foo` becomes `foo`).
   3. If the name has any `[...]` annotation prefix other than `[constructor]`,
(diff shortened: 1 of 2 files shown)


badeend (approved):
I think this makes sense and offers wider compatibility across different languages 👍

lukewagner:
Good point!  Makes sense to me.  It might be good to get a bit of implementation feedback (to see if this breaks anything) before merging though.

bvisness:
Yeah, for sure. I do think this is a direction we want to go but there's no urgency to merge this; we can work around stuff for now (and are unlikely to encounter this in practice).

alexcrichton:
Implementation-wise this is [easy enough](https://github.com/bytecodealliance/wasm-tools/pull/2626). I expect various minor tests to break with this change (e.g. [this spec test](https://github.com/WebAssembly/component-model/blob/4acb0deec245ec84972335762738adb303907612/test/validation/kebab.wast#L6-L7)), but I'd be surprised if this had any broader fallout.

@lukewagner was there any sort of specific implementation feedback you were looking to acquire? My guess is that if this actually breaks anyone we won't actually know until late in deployment when it's practically a bit too late to roll back, but I still think that the chance of that is quite small. Basically, personally, I'd be happy to go ahead and merge the wasm-tools change into Wasmtime and such think it's reasonable to merge the spec PR here.

lukewagner:
@alexcrichton Good to hear on both counts, thanks!  Then agreed to merge now.

@bvisness Do you suppose you could extend the (rather meager atm) strong-uniqueness tests at the end of `test/validation/kebab.wast` with some cases to check the new rules?

bvisness:
Yes, I can at minimum add all the new examples from the explainer, and maybe others if I think of them.

alexcrichton:
@bvisness can you double-check that the tests [added here](https://github.com/bytecodealliance/wasm-tools/pull/2626/changes#diff-76922085ee0dcd4982f91dbc4bde422c5bc17b3015e23593d446763aa8ad412f) are as you intended?

bvisness:
Yeah those look good, and I suggested an extra one as well.

bvisness:
Added. Should I just wait for a wasmtime release before merging this? I assume this is what @alexcrichton is working towards with that wasm-tools update?

alexcrichton:
Personally I think it's fine to merge this whenever, we haven't historically had a strict policy about exactly when to merge and it's mostly based on vibes. For changes like this which are technically breaking but aren't expected to break anything in practice they've just been merged whenever

lukewagner (approved):
Since this is just a restriction (not a new feature) and implementation feedback is in, I think we can merge whenever.

PR #705, opened 2026-08-20
CABI: remove the may_enter flag/trap

lukewagner:
This PR removes the `may_enter` flag+trap that shows up on the `canon lift`, `resource.drop` (destructor) and `subtask.cancel` (eager-cancellation) paths as well as from some of the cooperative thread scheduling conditions.  It also adds a bunch of WAST tests for reentrance cases that used to trap but are now valid.  This PR also includes updated versions of the new tests sitting in #696, updating them accordingly.  (Both #696 and #697 are obviated by this PR.)

As background: already #650 removed the previous attempt to define+check "asynchronous recursion" (by necessity: it's not well-defined when the host is involved, unfortunately), leaving only a weaker check that prevents "synchronous recursion".  This trap has been around since before 0.2 was released and in theory was supposed to simplify the lives of guest toolchains and bindings generators, but now that we've built a bunch, it doesn't actually appear to be load-bearing (once you already have to support all the other kinds of reentrance).  In the meantime, there are valid use cases for reentrance (even before callbacks) coming soon, so it seems best to just remove the trap if it's costing (both perf and expressivity) without buying us anything concrete.

What still remains is the "run to completion" semantics (as it's now called in Explainer.md#component-invariants in this PR) of core wasm running as part of a 0.3.0 `async` function which serializes individual turns of the `callback` event loop using automatic backpressure; this is load-bearing in our current `async` bindgen.


dicej (approved):
LGTM, thanks!

I've updated https://github.com/bytecodealliance/wasmtime/pull/14146 to match, and uncovered and fixed a few latent divergences with the spec in the process.

PR #708, opened 2026-08-24
CABI: allow dropping a readable future end with a pending write

rvolosatovs:
The Explainer specifies that a component may call `future.drop-readable` before reading a value to signal loss of interest, in which case a blocked `future.write` completes with `future-write-result.dropped`. But `SharedFutureImpl.drop` asserted that any pending buffer is a `WritableBuffer`, i.e. a pending *read*, while the only pending buffer actually possible at this point is the buffer of a pending *write*: a reader cannot drop its end while its own read is pending (`CopyEnd.drop` traps while copying) and a writer cannot drop its end before the write resolves (`WritableFutureEnd.drop` traps unless DONE). Exercising this spec-legal path failed the assert instead of notifying the writer with DROPPED.

Flip the assert to expect a `ReadableBuffer` (the source buffer of the pending write), make `ReadableBufferGuestImpl` actually subclass `ReadableBuffer` (it previously only derived `BufferGuestImpl`, unlike `WritableBufferGuestImpl`, which does mix in `WritableBuffer`), and add a test reproducing the scenario.

### DIFF
--- a/design/mvp/CanonicalABI.md
+++ b/design/mvp/CanonicalABI.md
@@ -1751,7 +1751,7 @@ class BufferGuestImpl(Buffer):
   def is_zero_length(self):
     return self.length == 0
 
-class ReadableBufferGuestImpl(BufferGuestImpl):
+class ReadableBufferGuestImpl(BufferGuestImpl, ReadableBuffer):
   def read(self, n):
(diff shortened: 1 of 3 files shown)


lukewagner:
Oops thanks, good catch!

PR #710, opened 2026-08-24
Switch two tests to assert_malformed

alexcrichton:
These specifically shouldn't parse, or even create an AST, which means that the bytes can't be converted to text, for example.


lukewagner:
Ah right, thanks!

PR #711, opened 2026-08-26
Adjust rules for canonical interface names

alexcrichton:
Tweak the string-splitting to more closely match the intent of versions found in tooling, specifically:

* Document that build metadata (`+foo` in `0.0.1+foo`) is always split out and not part of the canonical name.
* Don't split out pre-release information since pre-release versions are incompatible with all other versions.

This additionally matches preexisting merging behavior in `wit-component` as discovered in bytecodealliance/wasm-tools#2622

### DIFF
--- a/design/mvp/Explainer.md
+++ b/design/mvp/Explainer.md
@@ -2883,15 +2883,23 @@ type checking already required).
 
 Any `valid semver` (as used in WIT) can be canonicalized by splitting it into
 two parts - the `canonversion` prefix and the remaining `semversuffix`. Using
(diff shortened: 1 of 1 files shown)


alexcrichton:
cc @chenyan2002 this is in response to [this comment thread](https://github.com/bytecodealliance/wasm-tools/pull/2622#discussion_r3855792115) where I believe the intention is that the preexisting split point used in `wit-component` should be the same as the one for canonical names.

lukewagner (approved):

lann (approved):
Oof yeah I completely missed this.

Issue #586, opened 2025-12-19
Unify wit's stance on trailing commas


sno2:
These productions currently allow trailing commas:

- Include with names list: `include-names-list`
- Use lists: `use-names-list`
- Record fields: `record-fields`
- Flag fields: `flags-fields`
- Variant cases: `variant-cases`
- Enum cases: `enum-cases`
- Tuple type list: `tuple-list`

And these disallow trailing commas:

- Include with names list: `include-names-list`
- Function parameters: `named-type-list`
- Constant-length generic parameter types: `option`, `result`, `future`, `stream`, `borrow`

All of the cases that we currently disallow trailing commas can accept trailing commas without adding ambiguity. I agree with @tschneidereit's judgment and think we should allow trailing commas in all currently disallowed cases.

Split off from https://github.com/WebAssembly/component-model/pull/585

lukewagner:
Makes sense to me except that, with `option`, `future`, `stream` and `borrow`, there are no lists/commas at all so I wouldn't think we would add the ability to add a spurious dangling comma.

alexcrichton:
I'd agree with @tschneidereit that we should allow trailing commas everywhere. I'd also go so far as to explicitly allow it for `option` and single-type-parameter things too since from a parser's perspective it seems reasonable to have a shared "parse parameterized types" routine which always accepts trailing commas without having to special-case single-type-only things. One example is that `stream<T>` is the only valid thing today, but in the future it's planned to go so far as `stream<T, U, V>` where trailing commas become more reasonable at that point.

lukewagner:
Future additions to futures and streams were one of my concerns actually.  For one thing, `stream<T,>` would today be the same as `stream<T>`, but if we added more type parameters, `stream<T,>` would not have the same meaning as `stream<T, _>`, which would be surprising.  Separately, I had been worried that commas and positional parameters are too subtle for the additional stream/future type parameters and that some other token as separator might improve legibility (e.g., `stream<u8 then result reply-with result>`).  This may not end up being what we go with, but I'd rather not commit ourself to commas in the short term.

If we want a regular principle to apply here, perhaps it's this: for n-ary expressions (where the grammar allows any number of operands), allow the trailing comma; but when the arity is fixed, keep it as is.

alexcrichton:
Oh interesting! To me `stream<T,>` is never something I'd consider equivalent to `stream<T,_>`. That may be my Rust history showing though since that's how commas work everywhere in Rust AFAIK.

[...]

PR #714, opened 2026-08-28
Allow trailing comma in `use-names-list`

### LINKED ISSUE #713: confusion with `use-names-list?`
In the WIT spec, there is the following line:

https://github.com/WebAssembly/component-model/blob/d98a0c4553fba25d707810777a589bf17a068db6/design/mvp/WIT.md?plain=1#L1616

1. `use-name-list` is not defined anywhere. Is this meant to be `use-names-list` again?
2. If the comment "`use-names-list?` means at least one `use-names-list` term" is true, then `use-names-list ::= use-names-item | use-names-item ',' use-names-list?` would disallow trailing commas, and could be more succinctly written as  `use-names-item (',' use-names-item)*`

It is strange to disallow trailing commas when this pattern is used to explicitly _allow_ trailing commas in other non-terminals, e.g. `record-fields`:

https://github.com/WebAssembly/component-model/blob/d98a0c4553fba25d707810777a589bf17a068db6/design/mvp/WIT.md?plain=1#L1665-L1666

(also, https://github.com/WebAssembly/component-model/issues/586 mentions `use-names-list` explicitly as allowing commas, and it doesn't look like anyone was disputing that)

eamonburns:
Fixes #713

Allow trailing comma in use-names-list to be consistent with other constructs (e.g. `include-names-list`, `record-fields`)

```wit
use types.{
  foo,
  bar as baz,
  quux,
};
```

This was an easy-to-miss sidenote, and was essentially an exception to the grammar as written. If trailing commas are meant to be explicitly forbidden, it could more easily be written as one of the following:

```ebnf
use-names-list ::= use-names-item
                 | use-names-item ',' use-names-list

use-names-list ::= use-names-item (',' use-names-item)*
```

Related: #586

### DIFF
--- a/design/mvp/WIT.md
+++ b/design/mvp/WIT.md
@@ -1613,8 +1613,6 @@ use-names-item ::= id
                  | id 'as' id
 ```
 
-Note: Here `use-names-list?` means at least one `use-name-list` term.
-
 ## Items: type
 
 There are a number of methods of defining types in a `wit` package, and all of


eamonburns:
It doesn't look like I can add people as reviewers, but @lukewagner mentioned being willing to review.

lukewagner:
Thanks!  Checking in `wasm-tools` just now, it matches the EBNF, so I guess it is just deleting the misleading line.

PR #719, opened 2026-09-13
CABI: refactor/simplify stream/future logic; fix delivery of `DROPPED`/`CANCELLED`

lukewagner:
This PR factors out the first commit of #717, since it's pretty nice on its own and can be merged earlier since it's just behavior-preserving refactoring, except for fixing a spec bug (where `DROPPED` wasn't being delivered promptly even though the other side had dropped) that now matches wasmtime.

The PR is able to remove 125 lines (!) from `definitions.py` by merging future+stream and read+write definitions that were formerly separate.  The key simplification was removing the intermediate `shared` object that both the readable/writable ends point to.  Previously, this `shared` object seemed necessary to serve as the host/guest abstract interface, but as this PR shows, a better alternative is to instead say that the host owns and uses a readable/writable end directly (as if it was a component with its own `handles` table), and thus the two ends can just point to each other directly with no intermediate object.


lukewagner:
More testing showed another related case of spec/wasmtime divergence where I think Wasmtime is right: after a `stream.cancel-{read,write}`, the `CANCELLED` `CopyResult` takes precedence over `COMPLETED` (so you either get `CANCELLED` or `DROPPED`).  Incidentally, some of the existing WASTs already tested this behavior; updated commit now includes several more.

dicej (approved):
LGTM

https://github.com/bytecodealliance/wasmtime/pull/14342 addresses the failing tests for Wasmtime.

PR #721, opened 2026-09-14
Add some more stream/future WAST tests for partial/zero reads/writes

lukewagner:
Independently of the `DROPPED` changes in flight in the other PR, this PR adds more tests for futures, repeated partial stream reads/writes, and zero-length read/writes.  Everything passes in Wasmtime with the exception of the last (rather corner) case in `test/async/zero-length.wast` (named `zero-write-after-cancelled-write`).  In this test, a zero-length read is notified by a non-zero-length write, but then that write is cancelled (successfully), and a new zero-length write is issued which in theory should return `BLOCKED` because there is no longer a pending read (even though the original zero-length read's notification has not yet been delivered).  This is symmetric to what happens in `zero-write-after-full-read` (which passes on Wasmtime, and just replaces the zero-length buffer with a non-zero-length buffer) so my impression is that the zero-length case is a bug, but lmkwyt @dicej @alexcrichton.


dicej (approved):
I'll work on fixing the failure in Wasmtime.

PR #722, opened 2026-09-15
Require getter/setter result/param types to agree

bvisness:
This was an oversight from #701. For bindgen to make sense, the return types of getters and the param types of setters must agree. A little fussing is required to handle getters that return `result<T>`.


design/mvp/Binary.md
```diff
@@ -444,20 +444,23 @@ Notes:
...
+  particular scope, it must be preceded by a corresponding `[get]` in the same
+  scope. That is, there must already be an import or export respectively in the
+  same scope where the entire name is identical (*before* canonicalization),
+  except that the `[set]` annotation is replaced with `[get]`.
 * 📡 Validation of `[set]` names requires that the function have exactly one
   parameter, unless the name is also annotated with `[method]`, in which case
   there must be two parameters, the first of which is `self` (subject to the
-  `[method]` validation rules above).
+  `[method]` validation rules above). The type of this parameter must match the
+  corresponding getter's *property type*.
```
lukewagner, line 461: I could be misremembering, but I thought we had discussed whether we wanted this rule and decided not to because there seemed to be legitimate use cases where they might need to be slightly different?

bvisness: I went back through #235 and #701 and couldn't find any dissent to [this comment](https://github.com/WebAssembly/component-model/issues/235#issuecomment-2616769903), so basically just went ahead and implemented it. It's totally possible I missed something. I think if we don't add this "property type" notion, then we are basically implying that getters can never throw, which seems unnecessarily restrictive to me, but if there are legitimate use cases it's easy to change.

lukewagner: Ok, I guess I can't find it either; I guess I just hallucinated it.  In any case, it's the type of thing where we can always relax the predicate over time if use cases do appear.

lukewagner (approved):

PR #723, opened 2026-09-15
CABI: tighten `subtask.cancel` behavior

lukewagner:
The recent PR #716 attempted to give the host more flexibility for which-and-how-many of the subtask's threads are to be resumed during `subtask.cancel` by replacing the `candidates` set with a plain `thread.yield`.  But `thread.yield` actually isn't quite right and while testing Wasmtime I see that's both more-permissive and less-permissive in different cases.  This new PR aims to fix this in a way that matches Wasmtime's behavior (which seems pretty reasonable from what I can tell) as tested by a set of new WAST tests that all currently pass.  But @dicej and @alexcrichton lemme know if the new rules described in the PR sound right or if I'm missing something.


alexcrichton (approved):
This seems reasonable to me, although this is an aspect of Wasmtime which is "probably whatever just happened to or was easiest to implement" so I wouldn't want to over-rotate on this too too much. This is a spec corner that seems possible to tweak in the future if necessary, however, and I'm always a fan of more tests!

dicej (approved):
LGTM; thanks for the new tests!

PR #725, opened 2026-09-17
Add another task.cancel test

lukewagner:
This PR adds `test/async/cancel-not-delivered.wast` which tests that, in a cross-component setting, `task.cancel` traps if called before a cancellation request is *delivered*, even if a component caller has called `subtask.cancel` and the cancellation request is *pending*. cc @dicej and @alexcrichton since it currently fails on Wasmtime.


dicej:
Thanks; I'll fix the failure.

